Issue #18, opened 2026-07-25
Docs describe behaviour the code does not have: checkDependencies, CREATE vs CREATE2, findDeployBlock validation, Base Sepolia


thedavidmeister:
## Summary

Several pieces of documentation in this repo describe behaviour the code does
not have. All verified against `4422e29` (v0.1.4). Grouped here because they
are one class of defect with one fix.

### 1. `CLAUDE.md` documents a function that no longer exists

`CLAUDE.md` still lists:

> - `checkDependencies(...)` — forks each network, verifies dependencies and Zoltu factory exist with expected codehashes

and:

> - `deployAndBroadcast(...)` — the main entry point: derives deployer from private key, calls `checkDependencies` then `deployToNetworks`

`checkDependencies` was folded into `deployToNetworks` in `503002e`
("refactor(deploy): fold the dependency check into deployToNetworks"). There is
no such function in `src/lib/LibRainDeploy.sol`. The same section describes
`deployToNetworks` as "**re-**verifies dependencies", which is likewise a
leftover from the two-pass design.

Related, the "Key Design Patterns" section says:

> **Dependency checking**: Before deploying to any network, all dependencies (contract addresses) are verified to have code on-chain.

Dependencies are now checked per network, and only on the deploy branch — an
already-deployed network skips the check entirely (deliberately; see the
`deployToNetworks` NatSpec).

### 2. `CLAUDE.md` describes the wrong deployment opcode

> the Zoltu proxy is deployed at the same address on every chain and uses CREATE with a predictable nonce

The factory uses **CREATE2 with a zero salt**, not CREATE. Disassembling
`ZOLTU_FACTORY_BYTECODE`:

```
00000000: PUSH1 0x00
00000002: CALLDATASIZE
00000003: DUP2
00000004: DUP3
00000005: CALLDATACOPY
00000006: DUP1
00000007: CALLDATASIZE
00000008: DUP3
00000009: CALLVALUE
0000000a: CREATE2          <-- 0xf5, salt is the zero pushed at 0x00
0000000b: DUP1
...
```

Confirmed empirically — the following passes:

```solidity
function testZoltuIsCreate2SaltZero() external pure {
    assertEq(
        address(uint160(uint256(keccak256(abi.encodePacked(
            bytes1(0xff), LibRainDeploy.ZOLTU_FACTORY, bytes32(0),
            keccak256(type(MockDeployable).creationCode)
        ))))),
[...]

Issue #17, opened 2026-07-25
deployZoltu reports the factory's revert data as deployedAddress on the failure path


thedavidmeister:
Found by an adversarial mutation-testing pass over the whole repo at `4422e29`
(2 commits past `sol-v0.1.4`).

## Summary

`LibRainDeploy.deployZoltu` reads its call output buffer unconditionally:

```solidity
success := call(gas(), zoltuFactory, 0, add(creationCode, 0x20), mload(creationCode), 12, 20)
deployedAddress := mload(0)
```

The EVM copies return data into the output region for **failed** calls as well
as successful ones. So when the factory call reverts with at least 20 bytes of
revert data, the first 20 bytes of that revert data land at memory `[12, 32)`
and become `deployedAddress` — a value the factory never returned as an
address.

That value is then reported to the operator in four places on the failure path:

```solidity
console2.log("Zoltu deployment failed. Success:", success, "Deployed Address:", deployedAddress);
console2.log("Code length at Deployed Address:", deployedAddress.code.length);
console2.log("Codehash at Deployed Address:");
console2.logBytes32(deployedAddress.codehash);
revert DeployFailed(success, deployedAddress);
```

`DeployFailed`'s NatSpec describes `deployedAddress` as the deployed address,
so an operator debugging a failed deploy is shown a fabricated address together
with a real code length and a real code hash *for that fabricated address*.

## Impact

Diagnostics only — the guard short-circuits on `!success`, so a failed call can
never be mistaken for a successful deployment. Rate accordingly.

Reachability against the canonical factory is limited: `0x7A0D…D12` is a
keyless (Nick's method) deployment, so only that exact bytecode can ever exist
there, and it only ever reverts with empty data. The path is reachable via
`etchZoltuFactory` and etched test doubles, and by any caller using
`deployZoltu` directly — it is `internal` and has no factory-codehash guard of
its own; only `deployToNetworks` checks `ZOLTU_FACTORY_CODEHASH`.

## Repro

Verified against `4422e29`.

```solidity
/// Factory stand-in whose calls always fail, reverting with exactly the twenty
/// bytes of its own address.
contract MockAddressRevertingFactory {
    fallback() external {
        bytes20 self = bytes20(address(this));
        assembly ("memory-safe") {
            mstore(0, self)
            revert(0, 20)
        }
    }
}

function testRevertDataLeaksIntoReportedAddress() external {
[...]

Issue #16, opened 2026-07-25
deployToNetworks reports success without deploying when expectedAddress is stale relative to creationCode


thedavidmeister:
Found by an adversarial mutation-testing pass over the whole repo at `4422e29`
(2 commits past `sol-v0.1.4`).

## Summary

`LibRainDeploy.deployToNetworks` decides whether to deploy purely from
`expectedAddress.code.length == 0`. It never checks that `expectedAddress` is
the address the supplied `creationCode` would actually deploy to.

The Zoltu factory is CREATE2 with a **zero salt** — disassembling
`ZOLTU_FACTORY_BYTECODE` gives `CALLDATACOPY` then
`CREATE2(callvalue, 0, calldatasize, 0)` — so the deployed address is a pure
function of the creation code:

```
address = keccak256(0xff ++ ZOLTU_FACTORY ++ bytes32(0) ++ keccak256(creationCode))[12:]
```

That value is computable in-library, deterministically, on every path. Because
it is never computed, a consumer whose pinned `expectedAddress` /
`expectedCodeHash` constants have drifted behind their contract's
`creationCode` gets a clean, silent no-op: the OLD contract is still at the OLD
address with the OLD code hash, so every network takes the skip branch, the
post-hoc code-hash check passes **against the old contract**, and the script
exits 0 having deployed nothing.

This contradicts the guarantee the README states:

> Pre-calculated addresses asserted post-deploy: silent failures fail loudly.

## Why the inputs can drift

`creationCode` is live (`type(X).creationCode`, recompiled on every run), while
`expectedAddress` / `expectedCodeHash` are committed constants in the consumer
repo. Those two *can* disagree, and that is exactly what happens when a
contract's bytecode changes and the pinned constants are not regenerated. On
the deploy path the mismatch is caught loudly (`UnexpectedDeployedAddress`); on
the skip path it is not caught at all.

rain.deploy is inherited across the org (raindex, rainlang, rain.factory,
rain.metadata, rain.math.float, rain.extrospection, rain.tofu.erc20-decimals,
st0x.deploy), so this sits on the path of every deploy script.

## Repro

Verified against `4422e29`. Add to `test/src/lib/`:

```solidity
// A different contract from MockDeployable, standing in for a bytecode bump.
contract MockDeployableV2 {
    uint256 public value = 43;
    uint256 public other = 99;
}

function zoltuCreate2Address(bytes memory creationCode) internal pure returns (address) {
    return address(
        uint160(
[...]

PR #19, opened 2026-07-25
test(deploy): close the coverage gaps a whole-repo mutation pass found

thedavidmeister:
Test-only. No change to `src/`.

Adversarial mutation-testing pass over the whole repo
(`adversarial-mutation-test` v0.27.0), scanning `4422e29` — 2 commits past
`sol-v0.1.4`. `main` has since been merged in, so the diff is against a `main`
that already carries #21 and #26.

## What was done

91 distinct mutations were applied one at a time to
`src/lib/LibRainDeploy.sol`, each run against the whole suite and restored via
VCS. Every behaviour of all seven functions plus the constants was probed: each
conditional, comparison, boundary, revert path, error argument order, assembly
operand, loop bound and fork restore. Three rounds were run; the third surfaced
no new fillable gap.

- **67 mutants were killed by the tests that already existed** — those tests are
  validated by that, and this PR does not touch them.
- **18 mutants needed a new or strengthened test** — this PR closes all of them.
- **5 mutants are provably equivalent** and **1 is a defensive guard with no
  deterministic discriminator** (both listed below); no test is added for those.

## New tests

| Behaviour that no test pinned | Test |
|---|---|
| `isStartBlock` at block 0 checks only the code hash at that block, with no block before genesis to compare | `testIsStartBlockGenesisAllocationAtBlockZero` |
| `isStartBlock` at block 1 still compares against block 0 | `testIsStartBlockGenesisAllocationAtBlockOne` |
| `isStartBlock` compares against the *immediately* preceding block | `testIsStartBlockOneBlockAfterDeployBlock` |
| `findDeployBlock` returns the exact deploy block, including when it is the block the fork is at | `testFindDeployBlockExactZoltuBaseDeployBlock` |
| `findDeployBlock` leaves the fork on its original block when it reverts `DeployedBeforeStartBlock` | `testFindDeployBlockRestoresForkOnDeployedBeforeStartBlockRevert` |
| `deployZoltu` rejects a successful factory call that left no code | `testDeployZoltuRevertsEmptyCreationCode` |
| `deployZoltu` rejects a *failed* factory call even when its output buffer holds the address of a contract that has code | `testDeployZoltuRevertsWhenFactoryCallFailsWithAddressData` |
| `deployZoltu` never reports the zero address as a deployment | `testDeployZoltuRevertsZeroAddressWithCode` |
| `deployZoltu` does not forward the caller's value to the factory | `testDeployZoltuDoesNotForwardValue` |
| `deployToNetworks` treats a dependency that *has* code as present | `testDeployToNetworksPresentDependencyDeploys` |
| `deployAndBroadcast` broadcasts as the address derived from the given private key, and forwards it to `deployToNetworks` | `testDeployAndBroadcastUsesDeployerFromPrivateKey` |

The new `isStartBlock` block-0 and block-1 cases use the OP Stack WETH9
predeploy from the Base genesis allocation, which is the only way to reach a
target that already has its code at block 0. `testFindDeployBlockExactZoltu\
BaseDeployBlock` asserts the exact block (1117029, resolved independently by
binary searching `eth_getCode` over the Base RPC) rather than round-tripping
through `isStartBlock`, so the two functions no longer only validate each
other.

## Strengthened in place (no duplicates added)

- **`testZoltuFactoryCodehash`** now iterates `supportedNetworks()` instead of
  forking Arbitrum only, so the factory's code hash is pinned on all five
  networks rather than on one. At the scanned commit this was also the only
  thing pinning that every name in `supportedNetworks()` is a real fork alias,
  a bogus `BASE_SEPOLIA`, `FLARE` or `POLYGON` surviving the whole suite. That
  half no longer holds against current `main`: the chain-matrix tests #26 added
  fork every supported network too, and a bogus alias is killed by four of them
  with this change reverted. What is still exclusive to this test is the *code
  hash* assertion — nothing else checks the factory is the expected contract on
  each network.
- **`testDeployToNetworksMultipleNetworks`** now asserts the chain left selected
  and the deployed code hash. Its name claims it covers "deploy to *every*
  network", but it only asserted the returned address — which is identical for
  both networks, so truncating the loop to `i < 1`, or forking `networks[0]`
  every iteration, both survived it.

## New fixture

`test/concrete/MockAddressRevertingFactory.sol` — a factory stand-in whose calls
always fail, reverting with exactly the twenty bytes of its own address. It is
what makes the `!success` term of `deployZoltu`'s guard independently
observable. It sits in `test/concrete/` with the other mocks (#21) and is pinned
to `=0.8.25` per the concrete-pragma convention.

**#20 adds this same file, byte for byte.** Whichever of #19/#20 lands first,
the other's copy is a no-op. Both PRs also touch
`test/src/lib/LibRainDeploy.t.sol`, so the second to land needs a merge of
`main` first.

## Survivors deliberately left untested

Provably equivalent:

- `low = startBlock` → `startBlock + 1`: the pre-check proves the target does
  not have the expected code hash at `startBlock`, so the answer is always
  greater than `startBlock`.
- `deployBlock = low` → `high`: `low == high` at loop exit.
- `call(..., 12, 20)` → `call(..., 12, 32)`: `mload(0)` reads bytes `[0, 32)`
  either way, so the value is identical for any return-data length.
- `deployedAddress.codehash` → `expectedAddress.codehash` in the final check:
  the two are provably equal at that point on both branches.
- `deployAndBroadcast`'s own `NoNetworks` guard: `deployToNetworks` reverts with
  the identical error, so the guard is redundant.

Not equivalent, but not deterministically discriminable from a test:

- deleting `mstore(0, 0)` in `deployZoltu`. It guards against stale scratch
  memory bleeding into `deployedAddress` when a factory returns fewer than 20
  bytes. Scratch happens to be zero at that point under the current compiler,
  so its removal is currently unobservable, and a test cannot dirty
  compiler-owned scratch space deterministically. (Actively dirtying it *is*
  caught — `mstore(0, not(0))` is killed by two existing tests.) The guard is
  correct and should stay.

## Findings filed separately

The adversarial half of the pass produced three issues, each with a verified
repro: #16, #17, #18. None of them are enshrined by the tests here — in
particular `testDeployZoltuRevertsWhenFactoryCallFailsWithAddressData` asserts
only the `DeployFailed` selector, not the reported address that #17 is about,
so it stays green under that fix. Note that once #20 lands that fix, the
`deployedAddress == address(0)` term catches the failed call on its own, and
#20's own `testDeployZoltuFailedCallReportsZeroAddress` — which asserts the
exact `DeployFailed(false, address(0))` args — supersedes this one.

## Scan record

`audit/mutation-test-scans.json` records the scan for org-wide health tracking.


thedavidmeister:
🤖 ai:vetter
Reviewed 4fb3d1381d41982cda9656ccde806ebb0cd3b331: reject — Missing QA evidence block (QA-GUIDE §8) — no ## QA section in body or trusted producer comment. The test code itself audits sound: all 11 named tests present and verified discriminating against src/lib/LibRainDeploy.sol, equivalence claims for the 5 untested mutants hold. Add the §8 block (the body's ledger is the content) and re-push. Refs #16 #17 #18.
cost 455 — fork-test suite, full library re-read

thedavidmeister:
🤖 ai:producer
Blocked-on: Blocked on rainlanguage/rainix#289 (fork-RPC failover preflight), which is now ai:design pending a human ruling. Every failing test in this run is a fork-instantiation failure upstream of this diff: vm.createSelectFork could not instantiate a forked environment because drpc returned error code -32001, You've reached the usage limit for your current plan, preceded by HTTP 408 Request timeout on the free plan and HTTP 429 free plan rate limit. Foundry maps one rpc_endpoints alias to exactly one URL and --fork-retries only re-hits that same dead URL, so no code change in this repo can route around an exhausted vendor quota. Goes green once the failover lands on rainix main.

PR #34, opened 2026-08-14
test(fork): split the tests that need a chain from the ones that do not

### LINKED ISSUE #32: main goes red on RPC weather: 37 forks, no retry, and the fork/chain split does not hold
`main` is red right now, and was red at `8d1e5fd`, from a provider returning `HTTP error 500 … "Temporary internal error. Please retry"` on `vm.createSelectFork`. Nothing about the repo changed. This recurs — #26's own PR body records `base_sepolia` answering `408 Request timeout on the free plan`, PR #19's CI failed on two chain tests for the same reason, PR #20's on five.

A repo whose default branch reds on provider weather is a repo where red stops meaning anything.

## Three things combine to cause it

**1. The fork surface is large.** 37 `createSelectFork` calls across three test files on `main`:

| file | calls |
| --- | --- |
| `test/src/lib/LibRainDeploy.t.sol` | 31 |
| `test/src/abstract/RainDeployVerifyChain.t.sol` | 5 |
| `test/src/abstract/RainDeployVerifyChainCandidate.t.sol` | 1 |

**2. The fork-free/chain split does not hold where it matters.** #26 separated chain-anchored assertions into `*Chain*` contracts so an unreachable endpoint fails only those, and `CLAUDE.md` documents `forge test --no-match-contract Chain` as forking nothing. But `LibRainDeployTest` is not a `Chain` contract and holds 31 of the 37 forks — so that documented selection forks 31 times, and an outage fails it. The split is real for the verification abstracts and absent for the library's own suite.

**3. Nothing retries.** There is no retry or fallback around any `createSelectFork` in the repo. One transient 500 fails the job.

## Many of those forks buy nothing

Some tests fork and then replace everything they touch with `vm.etch`, reading no chain state at all. `testDeployZoltuFailedCallReportsZeroAddress` (`test/src/lib/LibRainDeploy.t.sol:399`) forks Arbitrum, then etches the factory and deploys `type(MockDeployable).creationCode` — nothing in it depends on the fork. `RainDeployVerifyBase.deriveDeployment` proves the point by running `deployZoltu` on a bare local EVM.

`testDeployZoltuRevertsWhenFactoryCallFailsWithAddressData` (`:1133`) is a second instance and is also fully subsumed by the test at `:399` — same fork, same etch, same call, but asserting only the revert selector where `:399` asserts the exact payload. It can be deleted outright.

These are examples, not the list. The category is: a test that etches everything it reads does not need a fork.

## Ask

1. Audit all 37. Delete the fork from every test that reads no chain state; delete `:1133` entirely.
2. Whatever genuinely needs a chain moves into a `*Chain*` contract, so `--no-match-contract Chain` forks nothing, as `CLAUDE.md` already claims.
3. Decide whether the remaining chain tests get a retry around `createSelectFork`. Shared CI concern — if the answer is yes, it likely belongs in rainix rather than here.

Ordering matters: (1) and (2) shrink the surface (3) has to cover.

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/32

Asks 1 and 2. Ask 3 (a retry around `createSelectFork`) is deliberately not
here — the issue itself says it is a shared CI concern that likely belongs in
rainix, and its ordering note says 1 and 2 come first because they shrink the
surface 3 has to cover. The gate itself now forks zero times, down from 31, and
every fork left in the repo is inside a `Chain` contract or inside the library
loop those contracts exist to test.

## What the gate is now

`forge test --no-match-contract Chain` runs 101 tests, 0 failures, with **no
`.env` in the checkout at all**. Before this it ran 131 tests, 93 passing and
38 failing, every failure a missing `<NETWORK>_RPC_URL`.

The 30 difference is the 29 tests that moved to `LibRainDeployChainTest` plus
the one deleted; the 8 that went from failing to passing are the 8 whose fork
bought nothing.

## The audit

All 31 of `LibRainDeployTest`'s forks, one at a time. The question asked of each
was the issue's own: does this test read state no fixture can supply?

**Nine bought nothing.** Eight etch every address they read, so the fork became
`etchZoltuFactory(vm)` or, where the point was that an address is EMPTY, an
assertion of the precondition a bare EVM already holds:

- `testFindDeployBlockNotDeployedReverts` — the `NotDeployed` guard runs before
  any fork history is read.
- `testFindDeployBlockWrongCodeHashReverts` — same, and the factory is etched
  rather than forked to. What the etched code hash is on a real network is
  `LibRainDeployChainTest.testZoltuFactoryCodehash`'s subject.
- `testDeployZoltuRevertsNoFactory`, `testDeployZoltuRevertsZeroAddressWithCode`
  — both are ABOUT the factory address being empty, which is the bare EVM's
  state, now stated with an `assertEq(...code.length, 0)` instead of arranged by
  forking a chain that has it.
- `testDeployZoltuRevertsRevertingConstructor`,
  `testDeployZoltuRevertsEmptyCreationCode`, `testDeployZoltuDoesNotForwardValue`
  — etch the factory, deploy a mock, read nothing else.
- `testDeployZoltuFailedCallReportsZeroAddress` — the issue's own example. It
  etches `MockAddressRevertingFactory` over the factory and deploys
  `type(MockDeployable).creationCode`; nothing in it depends on the fork.

What makes swapping a fork for `etchZoltuFactory(vm)` sound is that the constant
being etched is itself chain-anchored, and stays that way:
`testZoltuFactoryBytecode` and `testZoltuFactoryCodehash` are in
`LibRainDeployChainTest`, asserting `ZOLTU_FACTORY_BYTECODE` against the live
factory on Arbitrum and `ZOLTU_FACTORY_CODEHASH` against it on all five
networks. So a test that etches the factory reads the same bytecode a fork would
have handed it — checked once, on a chain, instead of 31 times.

**`testDeployZoltuRevertsWhenFactoryCallFailsWithAddressData` is deleted**, as
the issue asks. Same fork, same etch, same call as
`testDeployZoltuFailedCallReportsZeroAddress`, and it asserted only the selector
via `expectPartialRevert` where the survivor asserts
`DeployFailed(false, address(0))` exactly. There is no input that distinguishes
them and no assertion the survivor does not make more strongly.

**The other 22 genuinely read a chain**, and moved to `LibRainDeployChainTest`
with 7 more that reached a fork through the library rather than directly — 29 in
all. Three kinds, and the contract's NatSpec is the rule rather than the list:

1. the `isStartBlock` / `findDeployBlock` history search, whose subject IS a
   chain's history — there is nothing to etch;
2. the Zoltu factory pins, where the LIVE factory is the oracle — etching it
   leaves `ZOLTU_FACTORY_BYTECODE` and `ZOLTU_FACTORY_CODEHASH` checking
   themselves, and `testDeployZoltu` / `testZoltuAddressMatchesFactoryDeploy`
   checking `zoltuAddress` against itself;
3. everything reaching `deployToNetworks`' or `checkResolvedAddressesOnNetworks`'
   own per-network fork loop, which is the thing under test.

`test/abstract/LibRainDeployTestBase.sol` holds the fixtures and external
wrappers both sides share, so the split duplicates none of them.

The remaining 5 test forks are in `RainDeployVerifyChainTest` and
`RainDeployVerifyChainCandidateTest`, which #26 already split out and which the
gate already excludes. Each is a per-network matrix assertion — the chain id the
matrix ends on, the network named in a failure — so none reads no chain state.
The 37th is `RainDeployVerifyChain.sol`'s own loop, which is library code, not a
test.

## `CLAUDE.md`

It claimed `--no-match-contract Chain` was "the whole snapshot gate", scoped to
the verification abstracts, because that was as far as the split reached. It now
reaches the whole repo, and the file says so as a PROPERTY — every
`vm.createSelectFork` sits in a contract with `Chain` in its name, so the only
way to break the gate is to fork from a contract it runs — rather than as a list
of forking contracts that can fall behind. The three kinds of test that need a
chain are written down with it, so the question when adding a test is answerable
without reading the suites.

## Verification

Via `nix develop -c`, on the merge of `main` at `ed75b87`:

- `forge test --no-match-contract Chain`, with no `.env` in the checkout —
  **101 passed, 0 failed, 101 total**
- `forge fmt --check` — clean
- `reuse lint` — compliant, 57/57 files
- `slither .` — 44 contracts, 100 detectors, 0 results
- `forge test --match-contract Chain --threads 1`, with all five
  `<NETWORK>_RPC_URL` set — **40 passed, 0 failed**, `LibRainDeployChainTest`
  itself 29/29. So every test that moved still passes against a real chain; the
  move is not "green because it stopped running".

101 + 40 = 141, which is the 142 `main` runs minus the one deleted test.

CI on this head is the demonstration. `rainix / legal` and `rainix / static`
pass. `rainix / test` runs everything including the forks: 138 passed, 3 failed,
and all three failures are `vm.createSelectFork: … HTTP error 408 … "Request
timeout on the free plan"` — two in `RainDeployVerifyChainCandidateTest`, one in
`LibRainDeployChainTest`, which is otherwise 28 passed / 1 failed. Nothing
outside a `Chain` contract failed. That is the property: the same weather, on
`main`, takes non-`Chain` tests down with it.

The red is ask 3's subject and not this PR's. Nothing here retries a fork, so a
free-plan 408 still fails the suite it lands in — it just no longer reaches the
selection that forks nothing.

## QA

- **Mutations applied.** Every one of the 8 tests whose fork was removed is
  proved to still assert something, by breaking the line it guards in
  `src/lib/LibRainDeploy.sol` and running the whole fork-free selection. Each run
  reported `101 total tests`, so a mutant that "survived" because nothing ran is
  ruled out by the count rather than assumed away.

  | mutation | tests that went pass → fail |
  | --- | --- |
  | `if (target.code.length == 0)` → `if (false)` | `testFindDeployBlockNotDeployedReverts` |
  | `if (target.codehash != expectedCodeHash)` → `if (false)` | `testFindDeployBlockWrongCodeHashReverts` |
  | deploy guard → `if (!success)` | `testDeployZoltuRevertsNoFactory`, `testDeployZoltuRevertsEmptyCreationCode`, `testDeployZoltuRevertsZeroAddressWithCode` |
  | drop `deployedAddress == address(0)` only | `testDeployZoltuRevertsZeroAddressWithCode` |
  | drop `deployedAddress.code.length == 0` only | `testDeployZoltuRevertsEmptyCreationCode` |
  | `revert DeployFailed(success, …)` → `DeployFailed(true, …)` | `testDeployZoltuRevertsRevertingConstructor`, `testDeployZoltuFailedCallReportsZeroAddress` |
  | `if success { deployedAddress := mload(0) }` → unconditional | `testDeployZoltuFailedCallReportsZeroAddress` |
  | `call(gas(), zoltuFactory, 0, …)` → `callvalue()` | `testDeployZoltuDoesNotForwardValue` |

  `testDeployZoltuRevertsNoFactory` is the one with no single-term mutation of
  its own: an empty factory yields `address(0)`, which both the zero-address and
  the code-length term catch, so dropping either alone leaves it passing. It
  dies when the success path stops checking what the factory reported at all,
  which is the behaviour it is about.

- **Discriminating tests.** None added. This PR removes forks and moves tests; it
  changes no `src/` statement, so there is no behaviour for a new test to
  discriminate. What it could silently do is leave a test asserting nothing,
  which is what the mutation table above exists to rule out — one row per
  de-forked test, none of them surviving.

- **Oracle.** The code, not the issue's report of it. Every fork was re-derived
  against what the test actually reads: the deleted test was compared line by
  line with the one it duplicates before deleting it, and the "etches everything"
  claim was checked by running the de-forked tests with no `.env` in the
  checkout, which is the condition a fork would fail under.

- **Category check.** The issue names `:399` and `:1133` as examples and states
  the category — "a test that etches everything it reads does not need a fork".
  Both examples are covered, and so are the other 7 instances of the category the
  issue does not name.

- **Collision with #33.** #33 also edits `test/src/lib/LibRainDeploy.t.sol`,
  replacing `testDeployZoltu`'s "the deterministic address predicted by the
  factory's nonce" with the CREATE2-under-a-zero-salt wording. That test needs a
  live factory as its oracle, so it is in `LibRainDeployChainTest` here, and it
  carries #33's wording — landing this after #33 does not revert it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,13 +61,41 @@ FLARE_RPC_URL=https://flare-api.flare.network/ext/C/rpc
 POLYGON_RPC_URL=https://polygon-bor-rpc.publicnode.com
 ```
 
-All five are needed: `RainDeployVerifyChain` forks every network in
-`supportedNetworks()`, so a missing or rate-limited endpoint fails it. Those
-failures are `vm.createSelectFork` errors, distinct from the
-`NotDeployedOnNetwork` a reachable network raises, and the snapshot contracts
-run regardless: `forge test --no-match-contract Chain`.
+All five are needed by the contracts that fork: `RainDeployVerifyChain` forks
+every network in `supportedNetworks()`, so a missing or rate-limited endpoint
+fails it. Those failures are `vm.createSelectFork` errors, distinct from the
+`NotDeployedOnNetwork` a reachable network raises.
 
 These are referenced in `foundry.toml` under `[rpc_endpoints]`.
+
+### `forge test --no-match-contract Chain` is the offline gate
+
+It needs no `.env` at all. Every `vm.createSelectFork` in the repo sits in a
+contract with `Chain` in its name, so the gate forks nothing and no amount of
+RPC weather can make it red. That is a property of where the forks are, not a
+list to keep in step: the only way to break it is to fork from a contract the
+gate runs.
+
+Which is the question to answer when adding a test. A test needs a chain only if
+it reads state no fixture can supply, and three kinds do:
+
+1. The fork-history search. `isStartBlock` and `findDeployBlock` roll a fork
+   backwards and read the target at each block, so a chain's history IS their
+   subject and there is nothing to etch.
+2. The Zoltu factory pins. `ZOLTU_FACTORY_BYTECODE` and `ZOLTU_FACTORY_CODEHASH`
+   are constants asserted against the factory as actually deployed, and
+   `zoltuAddress`' derivation is checked against what the live factory returns.
+   Etching any of it leaves the constants checking themselves.
+3. Anything reaching `deployToNetworks`' or `checkResolvedAddressesOnNetworks`'
+   own per-network fork loop, which is the thing under test.
+
+Everything else etches what it reads and forks nothing — a test that etches the
+factory and deploys into a bare EVM gets the same answer a fork gives it, only
+without the endpoint. So `LibRainDeploy`'s suite is split by that question:
+`LibRainDeployTest` (`test/src/lib/LibRainDeploy.t.sol`) forks nothing,
+`LibRainDeployChainTest` (`test/src/lib/LibRainDeployChain.t.sol`) is the three
+kinds above, and `LibRainDeployTestBase` (`test/abstract/`) holds the fixtures
+and external wrappers both sides share so the split duplicates none of them.
 
 ## Architecture
 
@@ -263,8 +291,9 @@ can set. Group 3 is what makes group 4's scope complete — a release group 4 is
 never handed is a release it cannot fail on.
 
 Group 4 lives in its own contract so an unreachable RPC endpoint fails only it,
-never the snapshot assertions — `forge test --no-match-contract Chain` is the
-whole snapshot gate, and nothing reachable from those contracts forks anything.
+never the snapshot assertions. That is the offline gate's `Chain`-in-the-name
+rule applied to the verification abstracts, and groups 1-3 are on the offline
+side of it.
 
 A single recorded code hash per version can only be true if the runtime code is
 the same on every network, so a constructor reading `block.chainid` or similar
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,13 +68,17 @@ fails it. Those failures are `vm.createSelectFork` errors, distinct from the
 
 These are referenced in `foundry.toml` under `[rpc_endpoints]`.
 
-### `forge test --no-match-contract Chain` is the offline gate
+### `forge test --no-match-contract Chain` says whose fault a failure is
 
-It needs no `.env` at all. Every `vm.createSelectFork` in the repo sits in a
-contract with `Chain` in its name, so the gate forks nothing and no amount of
-RPC weather can make it red. That is a property of where the forks are, not a
-list to keep in step: the only way to break it is to fork from a contract the
-gate runs.
+Every `vm.createSelectFork` in the repo sits in a contract with `Chain` in its
+name, so this selection forks nothing and no amount of RPC weather can make it
+red. A red here is the code; a red only in the `Chain` contracts is the code or
+the endpoints, and which one has to be read off the failure. That is what the
+split buys — not the ability to work without a network, which is worth nothing
+here, but a result that means something on its own.
+
+It is a property of where the forks are, not a list to keep in step: the only
+way to break it is to fork from a contract this selection runs.
 
 Which is the question to answer when adding a test. A test needs a chain only if
 it reads state no fixture can supply, and three kinds do:
@@ -291,9 +295,9 @@ can set. Group 3 is what makes group 4's scope complete — a release group 4 is
 never handed is a release it cannot fail on.
 
 Group 4 lives in its own contract so an unreachable RPC endpoint fails only it,
-never the snapshot assertions. That is the offline gate's `Chain`-in-the-name
-rule applied to the verification abstracts, and groups 1-3 are on the offline
-side of it.
+never the snapshot assertions. That is the `Chain`-in-the-name rule applied to
+the verification abstracts, and groups 1-3 sit on the side of it that forks
+nothing.
 
 A single recorded code hash per version can only be true if the runtime code is
 the same on every network, so a constructor reading `block.chainid` or similar
--- a/test/abstract/LibRainDeployTestBase.sol
+++ b/test/abstract/LibRainDeployTestBase.sol
@@ -1,0 +1,207 @@
+// SPDX-License-Identifier: LicenseRef-DCL-1.0
+// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+pragma solidity ^0.8.25;
+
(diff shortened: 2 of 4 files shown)


thedavidmeister:
Closing unmerged. The split served `forge test --no-match-contract Chain`, and that command has no caller — no CI step, no script, no flake package; rainix CI runs `forge test -vvv` over the whole suite. It appears in my own history only in the sessions spent building and checking it.

It also does not fix what https://github.com/rainlanguage/rain.deploy/issues/32 was filed about. CI runs everything, so after this landed CI would still fork 37 times and still go red on the Base Sepolia `HTTP 408` — as it did on this PR while it was open.

What survives goes to https://github.com/rainlanguage/rain.deploy/pull/38, which removes the documented claim instead of building 800 lines of test split to satisfy it.

Not carried over, and worth knowing they were real:
- `testDeployZoltuRevertsWhenFactoryCallFailsWithAddressData` is strictly weaker than `testDeployZoltuFailedCallReportsZeroAddress` — `expectPartialRevert` on the selector where the survivor asserts `DeployFailed(false, address(0))` exactly. Deleting it is still correct, just not urgent.
- The 8 de-forked tests were each mutation-proved to still discriminate without a fork, so the categorisation itself was sound. It is the consumer that was missing.

https://github.com/rainlanguage/rain.deploy/issues/32 stays open for the RPC red.

PR #87, opened 2026-08-15
Deploy dispatch stops restating the suite registry, and can name a released suite

### LINKED ISSUE #44: Deploy dispatch restates the suite registry in YAML, and its `choice` list can never name a released suite
Five independent reviewers (dimensions 1, 4 and 6, three separate hazard partitions) raised this same defect from different angles, so all five findings are collected here. They agree on the shape and differ only in which consequence they trace.

---

# Finding `HAZ6-02`

**Audit finding `HAZ6-02`** — dimension 6, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`.github/workflows/manual-sol-artifacts.yaml:17-29`

## Problem

HAZARD: the only sanctioned deploy path enumerates the suite keys a second time, in YAML, where nothing keeps it in step with the declaration — and the enumeration cannot express the keys that matter most.

SHAPE: `options:` hardcodes `address-registry` and `migration-registry`. The real key set is `RainDeploySuitesBase.allSuites()` = `releasedSuites()` ∪ candidates, and `LibRainDeploySnapshot.releasedLibraryBlock` (line 612-615) builds every released key as `<template.suite>@<tag>` — `address-registry@0_1_6`. `workflow_dispatch` `type: choice` accepts ONLY listed values, so no released key is dispatchable, today or ever. The comment at 23-26 acknowledges the drift risk and mis-states the consequence: it says "the script still refuses one, naming the valid keys, if this list falls behind", but the script is never reached — GitHub rejects the dispatch before the job starts. This is the org's shared-invariant-constant rule inverted: one locked key set spelled in two places, with the YAML copy structurally unable to hold half of it.

Second half of the same defect: the caller passes only `suite`. `rainix-manual-sol-artifacts` defaults `verify: true` and its own input doc says to "disable when deploying bytecode that does not match the current source (e.g. a pinned historical release deployed from stored creation code)" — precisely the released-suite case. So even with the key available, the run would broadcast and then fail Etherscan verification after spending the gas.

PRODUCTION SCENARIO: Rain adds a sixth network after cutting `sol-v0.1.6`. `RegistryDeployChainTest` immediately goes red on main — a released suite must be live on every network in `supportedNetworks()` — and stays red. The fix is to broadcast `address-registry@0_1_6`, the frozen creation code the release pins, onto the new chain. It is not in the choice list, so main is red with no way to fix it through the deploy workflow until somebody notices the YAML is the blocker, edits it, and gets a PR merged. The reachable-and-wrong path is worse: an operator picks `address-registry` (the candidate) instead, and because source has moved since the release, that broadcasts DIFFERENT bytecode to a DIFFERENT deterministic address — real gas on a live chain, a contract at an address no release describes, and `RegistryDeployChainTest` still red. Adding a third registry has the same shape: it is declared, generated, released and frozen from one list per `script/Build.sol`, and then silently undispatchable.

STRUCTURAL FIX: stop restating the key set. The registry already refuses an unknown key by name and reports the valid ones (`UnknownDeploymentSuite(requested, suiteNames())`), which is the fail-closed behaviour the choice list was standing in for.

## Proposed fix

`.github/workflows/manual-sol-artifacts.yaml` lines 17-37:

```diff
       suite:
-        type: choice
+        type: string
         required: true
         description: |
           Which declared suite to broadcast. One dispatch deploys one suite,
           because `DEPLOYMENT_SUITE` selects one — a repo with two deployed
-          contracts is two dispatches. Offered as a choice rather than typed,
-          so a key that no suite declares cannot be dispatched at all; the
-          script still refuses one, naming the valid keys, if this list falls
-          behind the declaration.
-        options:
-          - address-registry
-          - migration-registry
+          contracts is two dispatches. Typed rather than chosen from a list:
+          the keys are `RainDeploySuitesBase.allSuites()`, which includes every
+          RELEASED suite as `<key>@<tag>` (e.g. `address-registry@0_1_6`), and a
+          static `options:` list can neither hold those nor stay in step with
+          the declaration. `script/Deploy.sol` refuses an unknown key with
+          `UnknownDeploymentSuite`, naming every valid one, so a typo costs a
+          few seconds and no gas.
+      verify:
+        type: boolean
+        required: false
+        default: true
+        description: |
+          Passed through to `forge script --verify`. Turn OFF when broadcasting
+          a RELEASED suite (`<key>@<tag>`): it deploys the frozen creation code,
+          which no longer matches current source, so explorer verification
+          against source fails the run after the gas is spent.
 jobs:
   deploy:
     uses: rainlanguage/rainix/.github/workflows/rainix-manual-sol-artifacts.yaml@main
     with:
       # Passed through as DEPLOYMENT_SUITE; script/Deploy.sol dispatches on it
       # and reverts on anything else.
       suite: ${{ inputs.suite }}
+      verify: ${{ inputs.verify }}
     secrets: inherit
```

---

<details>

Could not refute. Verified against source: manual-sol-artifacts.yaml:17-29 hardcodes `options: [address-registry, migration-registry]`, a second enumeration of the keys RegistryDeploySuites.candidateSuites() declares. LibRainDeploySnapshot.releasedLibraryBlock emits released keys as `<template.suite>@<tag>` (verified in the generator's string.concat), so released keys are structurally absent from any hand-written options list. GitHub's `type: choice` accepts only listed values (UI and the dispatches REST API 422 on others), so the script's UnknownDeploymentSuite fail-closed guard is never reached for a missing key. The scenario is not invented: RainDeploySuitesBase.sol:54-57 states it as first-class — "a chain added after a release is exactly the case where an OLD snapshot has to be broadcast on its own". Second half also confirmed by fetching the upstream reusable workflow at rainlanguage/rainix@main: it declares `suite: type: string` (the narrowing is entirely this caller's doing) and `verify: type: boolean, default: true` documented verbatim as "Disable when deploying bytecode that does not match the current source (e.g. a pinned historical release deployed from stored creation code)". The caller passes only `suite`, so a released-suite dispatch would broadcast then fail explorer verification; foundry.toml:67 independently records that rainix "passes --verify by default". The proposed fix's `verify:` passthrough targets a real input. No test or guard covers this — grep over test/ finds no reader of .github/. One sub-claim is weak: the finding calls the comment's consequence mis-stated, but "the script still refuses one [a key no suite declares]" coherently describes the stale-superset direction, where the script IS reached; that is rhetorical, not structural. Severity lowered HIGH -> MEDIUM on value-at-risk: releasedSuites() is empty today so the list is currently exactly complete (pure future hazard); the failure mode is fail-closed (key undispatchable, fixed by a one-line YAML PR, on a branch RegistryDeployChainTest already reds independently); and the reachable-and-wrong path deploys the candidate through Zoltu — permissionless, CREATE2-idempotent, landing at the address the repo's own pins describe for that suite, with AddressRegistry inert under a zero root. Loss is L2 gas and a failed run, not funds, consensus state, or a mis-pinned consumer address.

</details>

---

# Finding `HZ6-04`

**Audit finding `HZ6-04`** — dimension 6, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`.github/workflows/manual-sol-artifacts.yaml:27-29`

## Problem

HAZARD: the deploy workflow enumerates suite keys as a `choice` input (options `address-registry`, `migration-registry`), duplicating the keys the declaration builds in `RainDeploySuitesBase.suiteNames()` (src/abstract/RainDeploySuitesBase.sol:198-203). The whole design elsewhere refuses a second copy of the suite keys — "the keys reported by a mistyped `DEPLOYMENT_SUITE` are built from that same array, so the failure message cannot fall behind the suites it describes" — and this is exactly that second copy, in a file no test can read: `fs_permissions` (foundry.toml:36-46) grants no access to `./.github`, so nothing in the suite can compare the two.

SHAPE: the workflow comment covers only one direction ("the script still refuses one, naming the valid keys, if this list falls behind"). The other direction has no mention and no guard: a declared suite absent from `options` is undispatchable. That direction is guaranteed to occur, because released suite keys are generated as `<candidateKey>@<tag>` (src/lib/LibRainDeploySnapshot.sol:612-616) and no release tag exists at the time this options list is written.

PRODUCTION SCENARIO: release `0_1_6` is cut and deployed; a sixth chain is added later. The design's answer is to broadcast the frozen release onto the new chain — "a chain added after a release is exactly the case where an OLD snapshot has to be broadcast on its own" (src/abstract/RainDeploySuitesBase.sol:55-57). The operator opens `Manual sol artifacts` and the only options are `address-registry` and `migration-registry`: the candidate keys. Dispatching one broadcasts HEAD's creation code to the new chain — a different address and codehash from the release — so the release is still not live there (`RegistryDeployChainTest` stays red with `NotDeployedOnNetwork`), a second deterministic address is now burned with unreleased code, and gas has been spent on the wrong contract. The operator's only correct move is to edit a workflow file mid-incident.

STRUCTURAL FIX: keep the `choice` (so an undeclared key cannot be dispatched at all) but make the copy checkable, so it cannot fall behind in either direction.

## Proposed fix

1) Grant the read in `foundry.toml`:

```toml
fs_permissions = [
  { access = "read", path = "./foundry.toml" },
  { access = "read-write", path = "./src" },
  { access = "read-write", path = "./test" },
  { access = "read", path = "./out" },
  # The dispatch workflow enumerates the suite keys as a `choice`. That is a
  # second copy of the registry, so it is read here and asserted against the
  # declaration rather than remembered.
  { access = "read", path = "./.github" },
]
```

2) Add `test/src/abstract/RegistryDeployDispatch.t.sol`:

```solidity
pragma solidity =0.8.25;

import {Test} from "forge-std-1.16.1/src/Test.sol";
import {DeploySuite} from "../../../src/abstract/RainDeploySuitesBase.sol";
import {RegistryDeploySuites} from "../../../src/abstract/RegistryDeploySuites.sol";

/// @title RegistryDeployDispatchTest
/// @notice The dispatch workflow's `suite` choice IS the set of keys that can
/// be broadcast at all. It is a second copy of the declaration, so it is
/// checked against it: a declared suite missing from it is a contract nobody
/// can deploy without editing CI — which is the state every RELEASED suite is
/// in the moment a release is cut — and an option no suite declares is a
/// dispatch that fails after the job has started and key custody is engaged.
contract RegistryDeployDispatchTest is RegistryDeploySuites, Test {
    string constant WORKFLOW = ".github/workflows/manual-sol-artifacts.yaml";
    /// The indentation `options:` entries sit at. Matched from the START of
    /// the line, so a key mentioned in the surrounding prose is not an option.
    string constant OPTION_PREFIX = "          - ";

    /// Every option the workflow offers, in file order.
    /// @return options One entry per `options:` list item.
    function dispatchableSuites() internal view returns (string[] memory options) {
        string[] memory lines = vm.split(vm.readFile(WORKFLOW), "\n");
        string[] memory found = new string[](lines.length);
        uint256 count = 0;
        for (uint256 i = 0; i < lines.length; i++) {
            if (vm.indexOf(lines[i], OPTION_PREFIX) != 0) {
                continue;
            }
            found[count] = vm.replace(lines[i], OPTION_PREFIX, "");
            count++;
        }
        options = new string[](count);
        for (uint256 i = 0; i < count; i++) {
            options[i] = found[i];
        }
    }

    /// PROPERTY: the dispatchable keys and the declared keys are the same set.
    function testEveryDeclaredSuiteIsDispatchable() external view {
        DeploySuite[] memory suites = allSuites();
        string[] memory options = dispatchableSuites();

        assertEq(options.length, suites.length, "dispatch options and declared suites are different sets");
        for (uint256 i = 0; i < suites.length; i++) {
            assertTrue(
                vm.contains(vm.readFile(WORKFLOW), string.concat(OPTION_PREFIX, suites[i].suite, "\n")),
                string.concat("declared suite is not dispatchable: ", suites[i].suite)
            );
        }
        for (uint256 i = 0; i < options.length; i++) {
            bool declared = false;
            for (uint256 j = 0; j < suites.length; j++) {
                declared = declared || keccak256(bytes(options[i])) == keccak256(bytes(suites[j].suite));
            }
            assertTrue(declared, string.concat("dispatch option names no declared suite: ", options[i]));
        }
    }
}
```

Coverage this needs beyond the two directions above: an options list with a trailing-whitespace variant of a real key fails (the comparison is byte-exact, which is what the workflow input is); a key appearing only in the `description:` prose is not counted as an option; and — the case that motivates the whole test — a `releasedSuites()` entry keyed `<key>@<tag>` fails until the workflow lists it, which is what makes cutting a release also update the dispatch surface. The stronger variant of this fix is to have `cutRelease()` rewrite the `options:` block from `suiteNames()`, making the two one artifact rather than two that are compared.

---

<details>

Survives. Every cited fact checks out against source. .github/workflows/manual-sol-artifacts.yaml:27-29 enumerates `address-registry`/`migration-registry` as a `choice`, which is a second copy of the keys declared in RegistryDeploySuites (src/abstract/RegistryDeploySuites.sol, candidate `suite:` fields) and surfaced by RainDeploySuitesBase.suiteNames(). foundry.toml fs_permissions grants read only on ./foundry.toml, ./src, ./test, ./out — no ./.github — and `grep -rn "\.github" test/ script/ src/` returns zero hits, so the untested claim is verified rather than assumed. The undispatchable-release direction is real: LibRainDeploySnapshot.writeReleasedSuitesLib emits released keys as `<key>@<tag>`, so no released key can ever appear in an options list written before any tag exists; Build.sol:114-137 cutRelease() freezes the record and regenerates libs but never touches .github, and neither README.md's three release steps nor CLAUDE.md's lifecycle section mentions updating the dropdown. The scenario is the repo's own documented case, not invented — RainDeploySuitesBase.sol:55-57 says a chain added after a release is exactly when an OLD snapshot must be broadcast alone. The workflow comment addresses only the forward direction ("the script still refuses one, naming the valid keys"), so the reverse is genuinely unmentioned and unguarded. Severity corrected MEDIUM -> LOW on value-at-risk: nothing on-chain is corrupted and no consumer pin is invalidated. Dispatching the candidate key on the new chain deploys HEAD's candidate at the candidate's own deterministic address, which is the ordinary pre-release lifecycle step and idempotent — a benign gas spend, not a "burned" address. The missed release deployment is reported loudly by the designed detector (RegistryDeployChainTest red with NotDeployedOnNetwork), so the system does not lie about state. And the dropdown constrains only the CI surface: RainDeployBroadcast selects by DEPLOYMENT_SUITE over the registry, so an operator already holding the key can broadcast a released suite by running script/Deploy.sol directly. The real cost is a workflow edit to reopen the sanctioned path plus possibly one wasted candidate deploy — guaranteed to occur on the first release, but low value-at-risk.

</details>

---

# Finding `H6-02`

**Audit finding `H6-02`** — dimension 6, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`.github/workflows/manual-sol-artifacts.yaml:17-29`

## Problem

HAZARD: the set of deployable suites is declared in Solidity and re-typed in YAML, so a suite key that changes or is added in the declaration silently loses its only sanctioned deploy path — and the released suites, which the design says are separately deployable, have never had one.

SHAPE: `RainDeploySuitesBase.suiteByName` (src/abstract/RainDeploySuitesBase.sol:213-222) selects over `allSuites()`, whose keys are `RegistryDeploySuites`' two candidate keys plus, once anything is released, one generated key per frozen release — `LibRainDeploySnapshot.releasedLibraryBlock` emits them as `<suite>@<tag>` (src/lib/LibRainDeploySnapshot.sol:612-616), e.g. `address-registry@0_1_5`. The workflow's `suite` input is `type: choice` with two literal options (lines 27-29), hand-copied. Its own description (lines 23-26) concedes the duplication — "the script still refuses one ... if this list falls behind the declaration" — which is a comment saying "remember to update this", not an enforcement. CLAUDE.md describes it as "a choice over the declared keys"; it is a choice over a subset somebody typed. Nothing anywhere compares the two.

SCENARIO: the first release is cut, and later a sixth network is added to `supportedNetworks()`. `RegistryDeployChainTest` immediately goes red, correctly, because release `address-registry@0_1_5` is not live on the new chain — and RainDeploySuitesBase.sol:54-57 names exactly this as the case an old snapshot must be broadcast on its own for. The only workflow that broadcasts cannot express that key: `type: choice` refuses anything not in `options`, so the operator either edits the workflow under time pressure with CI red, or (worse) dispatches `address-registry`, which broadcasts the CANDIDATE creation code to the new chain — a different address from the release the chain test is failing on — leaving the red exactly where it was plus one unintended deployment on a fresh chain.

STRUCTURAL FIX: stop enumerating the keys twice. The registry already produces a complete, self-describing refusal (`UnknownDeploymentSuite`, carrying `suiteNames()`), so a free-text input is strictly better informed than a stale dropdown.

## Proposed fix

Edit `.github/workflows/manual-sol-artifacts.yaml`, lines 17-29:

```yaml
      suite:
-        type: choice
+        type: string
         required: true
         description: |
-          Which declared suite to broadcast. One dispatch deploys one suite,
-          because `DEPLOYMENT_SUITE` selects one — a repo with two deployed
-          contracts is two dispatches. Offered as a choice rather than typed,
-          so a key that no suite declares cannot be dispatched at all; the
-          script still refuses one, naming the valid keys, if this list falls
-          behind the declaration.
-        options:
-          - address-registry
-          - migration-registry
+          Which declared suite to broadcast, exactly as `RegistryDeploySuites`
+          declares it — a candidate key (`address-registry`) or a released one
+          (`address-registry@0_1_5`), which is what a chain added after a
+          release has to be given. One dispatch deploys one suite, because
+          `DEPLOYMENT_SUITE` selects one. Typed rather than a dropdown: the
+          keys are declared in Solidity and a second copy here would be a copy
+          that falls behind — and a dropdown cannot name a released suite at
+          all. `script/Deploy.sol` refuses an unknown key in seconds, naming
+          every valid one, before it reads `DEPLOYMENT_KEY` or touches a chain.
```

If the dropdown is kept for ergonomics instead, it must stop being a second source of truth: add a test that parses this YAML and asserts its `options` are exactly `suiteNames()`, so a declaration change that leaves the list behind is a red build rather than a failed dispatch.

---

<details>

Verified against source; could not refute.

FACTS CONFIRMED
- `/home/gildlab/code/rain-deploy-audit/.github/workflows/manual-sol-artifacts.yaml` lines 17-29 are exactly as quoted: `suite` is `type: choice` with two hand-typed `options` (`address-registry`, `migration-registry`), and its own description concedes "if this list falls behind the declaration".
- The keys are declared in Solidity: `RegistryDeploySuites.addressRegistryCandidate()/migrationRegistryCandidate()` set `suite: "address-registry"` / `"migration-registry"`, and `RainDeploySuitesBase.suiteByName` (lines 213-222) selects over `allSuites()` = `releasedSuites()` ++ `candidateSuites()`. So the YAML list is a second, hand-maintained copy of a subset.
- Released keys are GENERATED, not hand-written: `LibRainDeploySnapshot` emits `suite: "<template.suite>@<tagForRecordPath(...)>"` (the `"\",\n            creationCode: "` block around src/lib/LibRainDeploySnapshot.sol:612-616), written by `cutRelease()` into `LibAddressRegistryReleased`/`LibMigrationRegistryReleased`. `type: choice` cannot accept a value not in `options`, so no released key can ever be dispatched without a hand edit at every release — and nothing in `cutRelease()` or `package-release.yaml` touches the workflow.
- The design explicitly names this as a required operation: `RainDeploySuitesBase.sol:54-57` — "A repo with one contract and several frozen releases gives each release its own key, because each is separately deployable — a chain added after a release is exactly the case where an OLD snapshot has to be broadcast on its own." And this workflow is the ONLY sanctioned broadcast path (script/Deploy.sol: "Run manually via the `Manual sol artifacts` workflow"; workflow header: "Deliberately `workflow_dispatch` only").
- CLAUDE.md does describe the input as "a choice over the declared keys", which is not what it is.
- No enforcement exists: grep across `test/`, `script/`, `src/` finds nothing that reads `.github/`, the YAML, or compares `options` to `suiteNames()`. `RainDeploySuitesBase.t.sol` / `RainDeployBroadcast.t.sol` test key uniqueness and `UnknownDeploymentSuite`, i.e. the Solidity side only.

SEVERITY CORRECTED MEDIUM -> LOW (value at risk, not depth)
- Nothing is released yet (`LibAddressRegistryReleased`/`LibMigrationRegistryReleased` are empty; `RegistryDeployChainTest` forks nothing), so the gap is latent until a first release AND a subsequent network addition.
- Every failure mode is fail-safe: the dropdown refuses anything off-list, and `suiteByName` refuses an unknown key with `UnknownDeploymentSuite(requested, suiteNames())` before `DEPLOYMENT_KEY` is read. No wrong bytes can be broadcast by this defect.
- The remedy is a one-line YAML edit; no key custody, no on-chain state to unpick.
- The "worse" branch (operator dispatches the candidate) costs gas and produces a permissionless, ownerless, deterministic CREATE2 deployment at an address the repo already pins as the candidate — it moves no consumer's address, corrupts no pin, and deploying the candidate is what precedes the next tag anyway. Real operational hole in the product's sanctioned path, but no production value at risk.

</details>

---

# Finding `Q4-05`

**Audit finding `Q4-05`** — dimension 4, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`.github/workflows/manual-sol-artifacts.yaml:27-29`

## Problem

The whole design of `RegistryDeploySuites` is one declaration all readers share (`script/Deploy.sol`, `RegistryDeploySnapshotTest`, `RegistryDeployChainTest`), and CLAUDE.md states the workflow's `suite` input is "a choice over the declared keys". It is not derived from them — lines 28-29 hardcode `address-registry` and `migration-registry`, and nothing in `test/`, `src/` or CI compares the two lists (`grep -rn '\.github' src test script` returns nothing).

The workflow's own comment claims the failure is covered ("the script still refuses one, naming the valid keys"), but that covers only one direction. The direction it does not cover is the one that matters for a deploy repo: add a third suite to `candidateSuites()`, forget this file, and there is no dispatchable option for it — the contract is generated, snapshotted, aliased and released, and simply never deployable, with every check green. Renaming a suite key has the same shape: the stale option dispatches and reverts, but only after a human has burned a dispatch discovering it.

## Proposed fix

Grant the test tree read access to the workflow, and bind the two lists with a test. In `foundry.toml` `fs_permissions`:
```diff
   # GeneratedSnapshotShapeTest reads the compiler's AST out of the artifact.
   { access = "read", path = "./out" },
+  # RegistryDeploySuitesWorkflowTest reads the manual deploy workflow to check
+  # its dispatch choices are exactly the declared suite keys.
+  { access = "read", path = "./.github" },
 ]
```
New file `test/src/abstract/RegistryDeploySuitesWorkflow.t.sol` (mirror placement by SUBJECT — the declaration in `src/abstract/RegistryDeploySuites.sol`):
```solidity
pragma solidity =0.8.25;

import {Test} from "forge-std-1.16.1/src/Test.sol";
import {DeployCandidate} from "../../../src/abstract/RainDeploySuitesBase.sol";
import {RegistryDeploySuites} from "../../../src/abstract/RegistryDeploySuites.sol";

/// PROPERTY: the suite keys a human can dispatch are exactly the suite keys
/// this repo declares. The declaration is the one list every other reader
/// shares; the workflow choice list is the one copy of it that lives outside
/// Solidity, so it is the one copy that can silently fall behind — and a
/// declared suite with no dispatch option is a contract that is generated,
/// released and never deployable, with every other check green.
contract RegistryDeploySuitesWorkflowTest is RegistryDeploySuites, Test {
    string internal constant WORKFLOW = ".github/workflows/manual-sol-artifacts.yaml";

    /// The `- <key>` entries under the `options:` key, in file order.
    function workflowOptions() internal view returns (string[] memory options) {
        string[] memory lines = vm.split(vm.readFile(WORKFLOW), "\n");
        string[] memory found = new string[](lines.length);
        uint256 count = 0;
        bool inOptions = false;
        for (uint256 i = 0; i < lines.length; i++) {
            string memory line = vm.trim(lines[i]);
            if (keccak256(bytes(line)) == keccak256("options:")) {
                assertFalse(inOptions, "two options blocks");
                inOptions = true;
                continue;
            }
            if (inOptions) {
                if (bytes(line).length == 0 || !vm.contains(line, "- ")) {
                    // The block ends at the first line that is not a `- item`.
                    break;
                }
                found[count] = vm.replace(line, "- ", "");
                count++;
            }
        }
        assertTrue(inOptions, "no options block in the workflow");
        options = new string[](count);
        for (uint256 i = 0; i < count; i++) {
            options[i] = found[i];
        }
    }

    /// Every declared suite key is dispatchable. This is the direction the
    /// script's own refusal cannot cover: a suite with no option is not
    /// deployable at all, and nothing reverts to say so.
    function testEveryDeclaredKeyIsDispatchable() external view {
        DeployCandidate[] memory candidates = candidateSuites();
        string[] memory options = workflowOptions();
        assertTrue(candidates.length > 0, "no candidates declared");
            bool present = false;
            for (uint256 j = 0; j < options.length; j++) {
                present = present
                    || keccak256(bytes(options[j])) == keccak256(bytes(candidates[i].snapshot.suite));
            }
            assertTrue(present, string.concat("suite not dispatchable: ", candidates[i].snapshot.suite));
        }
    }

    /// Every dispatchable key is declared: a stale option burns a dispatch
    /// discovering the script's refusal.
    function testEveryDispatchableKeyIsDeclared() external view {
        DeployCandidate[] memory candidates = candidateSuites();
        string[] memory options = workflowOptions();
        for (uint256 j = 0; j < options.length; j++) {
            bool declared = false;
                declared = declared
                    || keccak256(bytes(options[j])) == keccak256(bytes(candidates[i].snapshot.suite));
            }
            assertTrue(declared, string.concat("option names no declared suite: ", options[j]));
        }
    }

    /// The two lists are the same SIZE, so a duplicated option cannot make a
    /// missing one pass both directions above.
    function testOptionCountMatchesDeclaredCount() external view {
        assertEq(workflowOptions().length, candidateSuites().length, "option count");
    }

    /// No empty option: a `- ` with nothing after it would satisfy neither of
    /// the membership loops in a readable way.
    function testNoEmptyOption() external view {
        string[] memory options = workflowOptions();
        for (uint256 j = 0; j < options.length; j++) {
            assertTrue(bytes(options[j]).length > 0, "empty option");
        }
    }
}
```

---

<details>

Could not refute. Verified in source: .github/workflows/manual-sol-artifacts.yaml lines 28-29 hardcode `address-registry` and `migration-registry`, which is a literal second copy of the keys declared at src/abstract/RegistryDeploySuites.sol:113 and :141 in candidateSuites(). Nothing binds the two: `grep -rn '\.github' src test script` returns nothing, and foundry.toml fs_permissions grants read only on ./foundry.toml, ./src, ./test and ./out — no test can even read the workflow today. CLAUDE.md:402 states the property ("whose `suite` input is a choice over the declared keys") and nothing enforces it.

The workflow's own comment is genuinely one-directional, as the finding says: RainDeployBroadcast reverts naming valid keys for an option no suite declares, but nothing anywhere reacts to a declared suite that has no dispatch option. The declaration's own documentation is provably incomplete on the same point — RegistryDeploySuites.sol:79-81 says a third deployed contract is "a third named candidate below, a third entry here, and a third entry in script/Build.sol's generated-contract list — nothing else", when it also requires a fourth edit to this workflow.

This is not a documented org convention being restated: the options list is local repo content, not a rainix-owned reusable (the reusable is only the `uses:` @main ref, which is correctly pinned). The proposed fix's pattern is also idiomatic for this repo, not novel — LibRainDeploySnapshot.t.sol:193 already reads foundry.toml and GeneratedSnapshotShape.t.sol reads ./out under the same fs_permissions mechanism.

One sub-claim IS refuted and should not be relied on: the description's "the contract is generated, snapshotted, aliased and released, and simply never deployable, with every check green" is wrong. releasedSuites() is only populated when a release is frozen, and RainDeployVerifyChain.checkDeployedOnNetwork reverts NotDeployedOnNetwork for any released suite with no code on a supported network — a release that was never broadcast is red, not green. The accurate failure modes are narrower: (a) a declared, not-yet-released suite sits green with no dispatchable option, discovered by the human when the dropdown lacks the key; (b) a renamed key leaves a stale option that reverts before DEPLOYMENT_KEY is read, costing one dispatch and no gas.

Severity stays LOW: the defect is real and sits on the deploy path of a repo whose deploy process is the product, but value at risk is bounded — no funds, no incorrect bytecode on chain, no silent bad state, and the failure announces itself to the human at dispatch time.

</details>

---

# Finding `DEPLOY-5`

**Audit finding `DEPLOY-5`** — dimension 5, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`.github/workflows/manual-sol-artifacts.yaml:23-37`

## Problem

script/Deploy.sol's whole selection contract is DEPLOYMENT_SUITE against the suite registry, and the registry is deliberately the only list — RainDeployBroadcast.sol:26-32 rejects 'the valid keys restated in a revert string that nothing keeps in step with the arms.' The dispatch workflow reintroduces exactly that: `options:` at lines 35-37 hardcodes `address-registry` and `migration-registry`.

Those two are the CANDIDATE keys. Released keys are generated with the tag appended — LibRainDeploySnapshot.releasedLibraryBlock (lines 612-616) emits `suite: "<template.suite>@<tag>"`, so the first release produces `address-registry@0_1_6`. That key can never be dispatched, because the choice list does not offer it and a choice input rejects anything else.

The use case this blocks is the one DeploySuite's own NatSpec names (RainDeploySuitesBase.sol:55-57): 'A repo with one contract and several frozen releases gives each release its own key, because each is separately deployable — a chain added after a release is exactly the case where an OLD snapshot has to be broadcast on its own.' Adding a supported network after a release is precisely when the frozen creation code must be broadcast, and that dispatch is unreachable. It is also the state RegistryDeployChainTest goes red on, so the repo would be red with no way to dispatch the fix through the workflow.

The workflow comment claims the choice is safe because 'the script still refuses one, naming the valid keys, if this list falls behind the declaration'. That covers the list offering too MUCH; it does not cover the list offering too LITTLE, which is the direction a generated key list grows in.

## Proposed fix

Make the input a free-form string and let the registry be the only list, which is what the script already enforces. In .github/workflows/manual-sol-artifacts.yaml replace lines 23-37:

      suite:
        type: string
        required: true
        description: |
          Which declared suite to broadcast. One dispatch deploys one suite,
          because `DEPLOYMENT_SUITE` selects one — a repo with two deployed
          contracts is two dispatches.

          Free text rather than a choice, because the declared keys are
          GENERATED: a release is emitted as `<suite>@<tag>`, so a fixed list
          can only ever offer the candidates and would make every frozen
          release undispatchable — which is exactly the dispatch a network
          added after a release needs. The registry is the only list;
          `RainDeploySuitesBase.suiteByName` refuses an undeclared key with
          `UnknownDeploymentSuite`, naming every valid one, before
          `DEPLOYMENT_KEY` is read and before anything forks, so a typo costs
          a few seconds and no gas.

DEPLOY-2's testWorkflowSuiteKeysAreDeclared is what keeps the free-text form honest for the candidate keys; extend it with each released key as releases are cut, or drop it in favour of asserting suiteNames() directly against the declaration.

---

<details>

Survives. Verified against source: the workflow input is `type: choice` with options `address-registry`/`migration-registry` (manual-sol-artifacts.yaml:17-29 — the finding's 23-37 range encloses the options block, actual lines 27-29), and GitHub rejects any value not offered by a choice input. Released suite keys are generated with the tag appended — LibRainDeploySnapshot.sol:612-616 concatenates `template.suite`, `"@"`, `tagForRecordPath(...)`, pinned by LibRainDeploySnapshot.t.sol:437 asserting `suite: "address-registry@0_0_1"` — so no released key can ever match an option. RainDeploySuitesBase.suiteByName resolves over allSuites() = released + candidates, so released suites ARE broadcastable by script/Deploy.sol; only the dispatch UI blocks them. The blocked case is the one DeploySuite's own NatSpec names ("a chain added after a release is exactly the case where an OLD snapshot has to be broadcast on its own") and is the state released-only RainDeployVerifyChain goes red on. Dispatching the candidate is not a substitute once source moves: different creation code, different CREATE2 address. Nothing regenerates the option list — script/Build.sol cutRelease() writes snapshots and released libs only, no .github writes anywhere in script/ or src/lib/ — and grep of the whole test tree shows no test reads any YAML (vm.readFile only touches build artifacts and foundry.toml), so nothing enforces the list in either direction at this commit. The workflow comment's justification ("the script still refuses one ... if this list falls behind the declaration") covers only the too-much direction; a generated key list grows the declaration past the list, which nothing covers, and CLAUDE.md's "a choice over the declared keys" is only true while no release exists. Severity LOW is correct: latent today (zero releases), no funds or wrong-contract risk, workaround is a one-line workflow edit — but it dead-ends the sanctioned deploy path precisely when CI is red.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/44

## What was wrong

`.github/workflows/manual-sol-artifacts.yaml` declared `suite` as `type: choice` with `address-registry` and `migration-registry` hardcoded — a second enumeration of the key set `RainDeploySuitesBase.allSuites()` declares.

That copy is structurally incapable of holding half the key set. Released keys are GENERATED: `LibRainDeploySnapshot` emits `suite: "<template.suite>@<tag>"` when `cutRelease()` freezes a release, so `address-registry@0_1_6` can never appear in a list hand-written before any tag exists. `type: choice` rejects an unlisted value, so **no released suite was dispatchable through the only sanctioned deploy path, today or ever** — and the script's fail-closed `UnknownDeploymentSuite(requested, suiteNames())` was never reached to say so, because GitHub refuses the dispatch before the job starts.

The blocked operation is the one `DeploySuite`'s own NatSpec names (`src/abstract/RainDeploySuitesBase.sol:54-57`): a chain added after a release is exactly when an OLD snapshot has to be broadcast on its own. `RegistryDeployChainTest` reds on that state and stays red. The reachable-but-wrong answer was to dispatch the candidate key, which broadcasts different bytes to a different deterministic address and leaves the red where it was.

Second half of the same defect: the caller passed only `suite`, and `rainix-manual-sol-artifacts` defaults `verify: true`, documented verbatim as something to disable when "deploying bytecode that does not match the current source (e.g. a pinned historical release deployed from stored creation code)". So even with the key available, a released-suite dispatch would broadcast and then fail explorer verification **after** the gas was spent.

## The fix

- `suite` is now `type: string`. There is one list, in Solidity. `suiteByName` refuses an undeclared key naming every valid one, before `DEPLOYMENT_KEY` is read and before anything forks — a typo costs seconds and no gas, which is the fail-closed behaviour the dropdown was standing in for and could not deliver.
- `verify` is a new dispatch input, `boolean`, `default: true` (current behaviour), passed through to the reusable. Turned off, it is how a frozen release gets broadcast without failing verification against source that has since moved.
- The file's header now names the post-release use — adding a network means dispatching `<key>@<tag>`, not the candidate — so the workflow does not read as pre-release-only.
- `CLAUDE.md` said the input was "a choice over the declared keys" and `README.md` said "choosing a `suite`". Neither was true even for candidates. Both now describe what the input is, and README step 1 says when to turn `verify` off.

Side effect worth noting: `src/abstract/RegistryDeploySuites.sol:79-81` claims a third deployed contract is "a third named candidate below, a third entry here, and a third entry in `script/Build.sol`'s generated-contract list — nothing else". That was false while the dropdown existed (it needed a fourth edit, here). It is true again now, so that docstring needed no change.

## No test, deliberately

Two of the five collected findings proposed keeping the dropdown and binding it with a test that parses the YAML (granting `fs_permissions` read on `./.github`). Those tests presuppose the second copy: with `options:` gone there is no key set in this file, so there is nothing that can fall behind the declaration and nothing for a test to compare. The only assertion left available would be that a construct is absent — grep-shaped, permanently widening the test tree's filesystem access, guarding against a decision this file now explains at length. Say so if you want it anyway.

## QA

- Discriminating tests: n/a — the diff is a `workflow_dispatch` input declaration plus prose; nothing in it is reachable from `forge test`, and the repo grants no `fs_permissions` on `./.github` (`grep -rn "\.github" src/ test/ script/` is zero, unchanged by this PR). The executed check is `actionlint` over all three workflows, exit 0, which parses the YAML and validates the reusable call's schema. The only fuller check is dispatching the workflow, which is key custody and real money.
- Mutations applied: n/a — no executable code in the diff to mutate, and no test over it to survive a mutant. The equivalent adversarial pass was done by hand against the upstream contract: `verify: ${{ inputs.verify }}` was checked to name a real input of the reusable at `@main` (not invented), and typed `boolean` on both sides so the passthrough is type-clean rather than a string that fails template validation at dispatch time.
- Oracle: read independently of the file being changed. `rainlanguage/rainix@main`'s `rainix-manual-sol-artifacts.yaml` for the reusable's own input declarations (`suite: type: string`, `verify: type: boolean, default: true`, and its `${{ inputs.verify && '--verify' || '' }}` use); `src/lib/LibRainDeploySnapshot.sol:612-616` for released keys being emitted as `<suite>@<tag>`; `src/abstract/RainDeploySuitesBase.sol` `allSuites`/`suiteNames`/`suiteByName` for what the valid key set actually is; `src/abstract/RainDeployBroadcast.sol:run()` for `suiteByName` preceding `vm.envUint("DEPLOYMENT_KEY")`, which is what makes the typo-costs-no-gas claim true.
- Category check: the issue collects five findings on one defect with two halves. Both covered — the second enumeration is removed (`choice` -> `string`, three of the five findings' structural fix) and the missing `verify` passthrough is added (the MEDIUM finding's second half). The category, not just the examples: every doc surface asserting the now-false property was swept, not only the one line the findings quoted — `CLAUDE.md` and `README.md` both fixed, `script/Deploy.sol` and `src/abstract/RegistryDeploySuites.sol` checked and correct as they stand. The two findings that instead proposed keeping the dropdown plus a YAML-parsing test are answered above, not silently dropped.

### DIFF
--- a/.github/workflows/manual-sol-artifacts.yaml
+++ b/.github/workflows/manual-sol-artifacts.yaml
@@ -9,24 +9,48 @@ name: Manual sol artifacts
 # Order is: dispatch this once per suite, confirm `RegistryDeployChainTest`
 # passes on every supported network, then push the `sol-v*` tag.
 #
+# That is the pre-release use. The other one is after a release: adding a
+# network reds `RegistryDeployChainTest` until every RELEASED suite is live on
+# it, and the fix is to dispatch that release's own key — `<key>@<tag>`, the
+# frozen creation code — not the candidate, which is different bytes at a
+# different address. Both uses are the same dispatch with a different key.
+#
 # Deliberately `workflow_dispatch` only. Broadcasting is key custody and real
 # money; nothing about a merge or a tag should trigger it.
 on:
   workflow_dispatch:
     inputs:
       suite:
-        type: choice
+        type: string
         required: true
         description: |
-          Which declared suite to broadcast. One dispatch deploys one suite,
-          because `DEPLOYMENT_SUITE` selects one — a repo with two deployed
-          contracts is two dispatches. Offered as a choice rather than typed,
-          so a key that no suite declares cannot be dispatched at all; the
-          script still refuses one, naming the valid keys, if this list falls
-          behind the declaration.
-        options:
-          - address-registry
-          - migration-registry
+          Which declared suite to broadcast, spelled exactly as the declaration
+          spells it. One dispatch deploys one suite, because `DEPLOYMENT_SUITE`
+          selects one — a repo with two deployed contracts is two dispatches.
+
+          Typed rather than chosen from a list, so this file is not a second
+          copy of the declaration. The keys are `allSuites()` — the candidates
+          `RegistryDeploySuites` names, plus one per frozen release, and the
+          released ones are GENERATED as `<key>@<tag>` (`address-registry@0_1_6`)
+          by `cutRelease()`. A hand-written `options:` list could hold only the
+          candidates, which would make every release undispatchable — and
+          broadcasting a release on its own is exactly what a network added
+          after that release needs. There is one list, in Solidity:
+          `suiteByName` refuses an undeclared key with `UnknownDeploymentSuite`,
+          naming every valid one, before `DEPLOYMENT_KEY` is read and before
+          anything forks, so a typo costs seconds and no gas.
+      verify:
+        type: boolean
+        required: false
+        default: true
+        description: |
+          Whether to pass `--verify` to `forge script`. Leave ON for a candidate
+          key, which is a snapshot of current source.
+
+          Turn it OFF for a RELEASED key (`<key>@<tag>`): that broadcasts the
+          creation code the release froze, which is not what current source
+          compiles to once source has moved, so explorer verification against
+          source fails the run AFTER the gas is spent.
 jobs:
   deploy:
     uses: rainlanguage/rainix/.github/workflows/rainix-manual-sol-artifacts.yaml@main
@@ -34,4 +58,8 @@ jobs:
       # Passed through as DEPLOYMENT_SUITE; script/Deploy.sol dispatches on it
       # and reverts on anything else.
       suite: ${{ inputs.suite }}
+      # The reusable defaults this to true, so it has to be passed to be
+      # turnable off — a released suite broadcasts frozen bytes that current
+      # source no longer matches.
+      verify: ${{ inputs.verify }}
     secrets: inherit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -399,9 +399,13 @@ expected addresses, expected code hashes, and dependency lists.
 - **Deploy, then verify, then tag** — in that order, and they are three separate
   things. `script/Deploy.sol` broadcasts the suite `DEPLOYMENT_SUITE` names to
   every network in `supportedNetworks()`, dispatched by hand through
-  `.github/workflows/manual-sol-artifacts.yaml`, whose `suite` input is a choice
-  over the declared keys. One suite per dispatch, so this repo's two registries
-  are two dispatches. Only then is there a deployment for `rainix-tag-release`
+  `.github/workflows/manual-sol-artifacts.yaml`, whose `suite` input is typed
+  rather than picked from a list — the workflow is not a second copy of the
+  declaration, so it cannot fall behind it, and released keys are generated as
+  `<key>@<tag>` and could never appear in a hand-written list at all. An
+  undeclared key is refused by `suiteByName` naming the valid ones, before the
+  key is read. One suite per dispatch, so this repo's two registries are two
+  dispatches. Only then is there a deployment for `rainix-tag-release`
   to verify pins against — it verifies and publishes, it never broadcasts.
   Broadcasting is key custody and real money, so it is `workflow_dispatch` and
   nothing else. Deploying is idempotent: a network that already has the code is
--- a/README.md
+++ b/README.md
@@ -288,12 +288,23 @@ Three separate steps, in this order. Nothing automatic ever broadcasts.
 
 1. **Deploy.** Dispatch the
    [`Manual sol artifacts`](.github/workflows/manual-sol-artifacts.yaml)
-   workflow, choosing a `suite`. It runs `script/Deploy.sol` and broadcasts that
+   workflow, typing a `suite`. It runs `script/Deploy.sol` and broadcasts that
    suite to every network in `supportedNetworks()`. One suite per dispatch, so
    this repo's two registries are two dispatches. `workflow_dispatch` only: this
    is key custody and real money, and no merge or tag should be able to trigger
    it. It is idempotent — a network that already has the code is skipped — so a
    partial run is fixed by running it again rather than by unpicking anything.
+
+   The key is any key the declaration declares — a candidate
+   (`address-registry`) or a frozen release (`address-registry@0_1_6`). A typed
+   input rather than a dropdown, because a dropdown would be a second copy of
+   the declaration and could never name a release, whose key is generated when
+   the release is cut. An undeclared key is refused in seconds, naming the valid
+   ones, before the deploy key is read.
+
+   Turn `verify` OFF when the key names a release: it broadcasts the creation
+   code that release froze, which current source no longer compiles to, so
+   verifying against source would fail the run after the gas is spent.
 2. **Verify.** `RegistryDeployChainTest` passes only once every **released**
    suite is live on every supported network, with the code that release froze.
    This repo has released none, so today it has nothing to check and passes; it


thedavidmeister:
@coderabbitai review

thedavidmeister:
@coderabbitai review

thedavidmeister:
Closing: the restatement this removes is deliberate.

The `options:` list in `manual-sol-artifacts.yaml` is a fixed enum on purpose. Offering the suite as a choice rather than as free text is what makes a key no suite declares undispatchable at all — the workflow's own comment says so, and the script still refuses an unknown key naming the valid ones if the list falls behind. Deriving the list from the declaration would trade that away for the appearance of a single source of truth.

Not a defect in the diff. Issue #44 rests on the same premise.

PR #90, opened 2026-08-15
test: drive run() through a successful broadcast

### LINKED ISSUE #58: run() is only ever tested on paths that fail before broadcasting — nothing pins what it would deploy, or where
**Audit finding `cov-05`** — dimension 2, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RainDeployBroadcast.sol:77-92`

## Problem

`RainDeployBroadcastTest`'s own docstring states it: "Nothing here broadcasts. Every case is one that fails before `deployAndBroadcast` is reached." The two cases are both `UnknownDeploymentSuite`. So no test drives `run()` through a successful suite selection, and these properties of the repo's only state-changing script (`script/Deploy.sol`, which is `run()` and nothing else) are unasserted:

- that the fields of the SELECTED suite are the ones handed to `deployAndBroadcast`. `suite.artifactPath` and `suite.suite` are both `string` and adjacent in the call, so swapping them compiles and would send `forge verify-contract` a suite key instead of a contract path; nothing would notice.
- that `deployNetworks()` — the documented override point for repos that bootstrap one chain per dispatch — is what is broadcast to. `testDeployNetworksDefaultsToSupportedNetworks` asserts the default value of that function in isolation, never that `run()` calls it rather than `LibRainDeploy.supportedNetworks()` directly. An override that `run()` ignored would broadcast a suite to five chains when the repo asked for one.

`testSelectedSuiteCarriesTheRecordedPins` asserts things about `suiteByName`, not about `run()`.

## Proposed fix

The env vars are process-global and this repo deliberately confines every `vm.setEnv` to one test function to avoid the race documented there, so the success leg is appended to that same function rather than given its own.

1. New fixture `test/concrete/ExampleDeploySingleNetwork.sol`:

```solidity
pragma solidity =0.8.25;

import {RainDeployBroadcast} from "../../src/abstract/RainDeployBroadcast.sol";
import {LibRainDeploy} from "../../src/lib/LibRainDeploy.sol";
import {ExampleDeploySuites} from "../abstract/ExampleDeploySuites.sol";

/// @title ExampleDeploySingleNetwork
/// A repo that bootstraps one chain per dispatch, which is what
/// `deployNetworks()` is overridable for. A broadcast that reached a network
/// this does not name would be a suite deployed somewhere the repo did not ask
/// for.
contract ExampleDeploySingleNetwork is ExampleDeploySuites, RainDeployBroadcast {
    /// @inheritdoc RainDeployBroadcast
    function deployNetworks() internal pure override returns (string[] memory networks) {
        networks = new string[](1);
        networks[0] = LibRainDeploy.ARBITRUM_ONE;
    }
}
```

2. Append to `RainDeployBroadcastTest.testRunSelectsTheSuiteFromTheEnvBeforeTheKeyAndNeverDefaults` (and rename it `testRunSelectsTheSuiteFromTheEnvBeforeTheKeyNeverDefaultsAndBroadcastsIt`), after the existing two legs:

```solidity
        // ## Then the suite it names, actually broadcast
        //
        // The half above decides WHAT would be deployed; this is the half that
        // says the decision is carried through. Sequenced in the same test for
        // the same reason the first two are: `vm.setEnv` writes the forge
        // PROCESS' environment and forge runs tests concurrently.
        ExampleDeploySingleNetwork single = new ExampleDeploySingleNetwork();
        address expectedAddress = LibRainDeploy.zoltuAddress(type(MockDeployableV2).creationCode);
        bytes32 expectedCodeHash = keccak256(type(MockDeployableV2).runtimeCode);

        vm.setEnv("DEPLOYMENT_SUITE", "second-address-candidate");
        vm.setEnv("DEPLOYMENT_KEY", "0xa11ce");

        single.run();

        // The suite `DEPLOYMENT_SUITE` named is the suite on chain, at the
        // address its own creation code derives and with the code hash that
        // creation code produces — so the selection reached the broadcast
        // rather than some other entry of the registry.
        assertEq(expectedAddress.code.length > 0, true);
        assertEq(expectedAddress.codehash, expectedCodeHash);

        // The OVERRIDE is what was broadcast to. A completed run leaves the
        // last network of `deployNetworks()` selected, so the chain id here is
        // arbitrum for the override and polygon for `supportedNetworks()` — the
        // only observation that tells an honoured override from an ignored one.
        assertEq(block.chainid, 42161);

        // And the suite the OTHER declaration selects is untouched, so the run
        // deployed the one key it was given and not the registry.
        assertEq(
            LibRainDeploy.zoltuAddress(sDeploy.externalSuiteByName("address-registry-candidate").creationCode).code.length,
            0
        );
```

with `MockDeployableV2` and `ExampleDeploySingleNetwork` added to the imports. A follow-up test in the same function can assert `suite.artifactPath` reaches the verification command by asserting on the console output is not available — instead pin it structurally by asserting `single.externalSuiteByName("second-address-candidate").artifactPath` equals `"test/concrete/MockDeployableV2.sol:MockDeployableV2"` immediately before the run, so the value forwarded is named in the same test.

---

<details>

SURVIVES (tried hard to refute; could not on the substance).

Facts verified against source:
- `/home/gildlab/code/rain-deploy-audit/src/abstract/RainDeployBroadcast.sol:77-92` — `run()` is `suiteByName(envOr(DEPLOYMENT_SUITE))`, `envUint(DEPLOYMENT_KEY)`, then one `LibRainDeploy.deployAndBroadcast(vm, deployNetworks(), key, suite.creationCode, suite.artifactPath, suite.storedDeployedAddress, suite.storedBytecodeHash, suite.dependencies)`. Nine lines of pure wiring.
- `grep -rn "run()" test/` gives exactly two call sites, both in `/home/gildlab/code/rain-deploy-audit/test/src/abstract/RainDeployBroadcast.t.sol` (lines 94, 105), both preceded by `vm.expectRevert(UnknownDeploymentSuite...)`. The file's own docstring states the gap. So the finding's core factual claim is exact: no test reaches `deployAndBroadcast` through `run()`.
- The two tests offered as substitutes do not cover the wiring. `testSelectedSuiteCarriesTheRecordedPins` asserts on `externalSuiteByName(...)` fields; `testDeployNetworksDefaultsToSupportedNetworks` asserts on `externalDeployNetworks()` (an `ExampleDeploy` wrapper). Neither observes `run()`. `deployNetworks()` is never overridden anywhere in the repo, so the virtual's only reason to exist — the documented one-chain-per-dispatch override — has no test that it is what `run()` consults. That mutant (`run()` calling `LibRainDeploy.supportedNetworks()` directly) survives the whole suite today.
- The gap is not "untestable". The test file's stated rationale ("a key, an RPC and real money") is contradicted by its own sibling: `/home/gildlab/code/rain-deploy-audit/test/src/lib/LibRainDeploy.t.sol` drives the full success path — `testDeployToNetworksMultipleNetworks` and `testDeployAndBroadcastUsesDeployerFromPrivateKey` fork real Base/Arbitrum, use key `0xA11CE`, and run `vm.startBroadcast`/`deployZoltu` for real on the fork. `forge test` broadcasts nothing off-machine, so no money is at risk. A documented gap is still a gap, and this one's documented justification is false in this repo's own terms — that does not qualify as the "deliberate documented convention" exemption.

Partial refutations that do NOT sink it, but that the fix should absorb:
- The `suite.artifactPath` / `suite.suite` swap claim is materially weaker than presented. `contractPath` is consumed by `deployToNetworks` in exactly one place (`LibRainDeploy.sol:448-453`), a `console2.log` of a manual `forge verify-contract` line. Nothing on chain depends on it, and no success test can pin it — the proposed fix concedes this and falls back to asserting `externalSuiteByName(...).artifactPath`, which is a verbatim restatement of the existing `testSelectedSuiteCarriesTheRecordedPins` and pins nothing about `run()`. That leg of the fix is vacuous and should be dropped.
- A wrong-field combination is partly guarded downstream: `deployToNetworks` (`LibRainDeploy.sol:397-400`) reverts `UnexpectedDeployedAddress` before forking whenever the passed creation code and recorded address disagree, so only a wholly-consistent wrong-suite substitution reaches a chain.
- Minor inaccuracy in the description: `script/Deploy.sol` is not the repo's only state-changing script (`script/Build.sol` has a `run()` that rewrites the generated snapshots); it is the only chain-state-changing one.

Severity left at MEDIUM. Value at risk is the org's only broadcast entry point, inherited by every downstream deploy repo, with zero success-path coverage; the unguarded residue is an override that `run()` could silently ignore (a consumer asking for one chain getting five, with a mid-run revert leaving a partial dispatch) and a wrong verify-contract path in the operator's log. Bounded by CREATE2 deploys being deterministic, idempotent and permissionless, and by the pin guards inside `deployToNetworks` — which is why it is not HIGH.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/58 (audit finding `cov-05`).

`run()` was only ever driven down paths that revert with `UnknownDeploymentSuite` before `deployAndBroadcast` is reached, so the nine lines that are the entire body of `script/Deploy.sol` asserted nothing about what they would deploy, or where.

## What this adds

**`test/concrete/ExampleDeploySingleNetwork.sol`** — the same `ExampleDeploySuites` declaration as `ExampleDeploy`, with `deployNetworks()` overridden to the one chain. That override is the documented reason the virtual exists (a repo bootstrapping one chain per dispatch) and nothing in the repo overrode it, so nothing said `run()` consults it.

**A third leg on `testRunSelectsTheSuiteFromTheEnvBeforeTheKeyNeverDefaultsAndBroadcastsIt`** (renamed from `...AndNeverDefaults`) that runs `run()` all the way to a chain on an Arbitrum fork, asserting:

- the suite `DEPLOYMENT_SUITE` named is the suite on chain — at the address its own creation code derives, holding the code hash that creation code produces;
- `block.chainid` is Arbitrum, which is what an honoured `deployNetworks()` looks like from inside a completed run (`deployToNetworks` forks each network without restoring, so a completed run leaves the *last* target selected — arbitrum for the override, polygon for the default).

Appended to that function rather than given its own, because it has to SET `DEPLOYMENT_SUITE` and `vm.setEnv` writes the forge *process* environment — a second test setting it is a second test racing the first, which is the hazard that function's docstring already exists to describe.

## QA

- **Discriminating tests:** `testRunSelectsTheSuiteFromTheEnvBeforeTheKeyNeverDefaultsAndBroadcastsIt` (third leg). Verified against base: the pre-fix version of `test/src/abstract/RainDeployBroadcast.t.sol` (restored from `86f8d96`, new fixture removed) was run against mutant A below and the **whole suite passed, 215/215** — i.e. the mutant survived everything on main. The same mutant against this branch fails.
- **Mutations applied:**
  - `src/abstract/RainDeployBroadcast.sol:84` — `deployNetworks(),` → `LibRainDeploy.supportedNetworks(),` (the override the virtual exists for, reached past) → killed by `testRunSelectsTheSuiteFromTheEnvBeforeTheKeyNeverDefaultsAndBroadcastsIt`, `assertion failed: 137 != 42161`. Pre-fix: **survives, 215 tests passed, 0 failed.**
  - `src/abstract/RainDeployBroadcast.sol:79` — inserted `suite = suiteByName("address-registry-candidate");` after the env selection (the selection ignored, a fixed registry entry broadcast instead) → killed by the same test, `assertion failed: 0 <= 0`. Note this mutant *succeeds* through `deployToNetworks` — `AddressRegistry` is already on Arbitrum, so it takes the skip branch — which is why the assertion has to be on the mock's address rather than on the run reverting.
  - Baseline re-run after each mutant: `3 passed; 0 failed`, tree clean (`git status --short` empty).
- **Oracle:** the deployed address and code hash come from `LibRainDeploy.zoltuAddress(type(MockDeployableV2).creationCode)` and `keccak256(type(MockDeployableV2).runtimeCode)` — CREATE2 over the creation code, derived in the test from the compiler's own artifact rather than read back from anything `run()` did or copied from the declaration under test. The chain id `42161` is a fact of the world, not a value the repo produces; it is asserted as a literal precisely so a misconfigured `arbitrum` RPC alias fails loudly instead of agreeing with itself. Non-vacuity of the address assertion was checked out-of-band: `0xE19c2335AdbFAD3250FA150739cC5C11cE5935eD` has no code on any of the five supported chains (`eth_getCode` on each), so it can only have code because the run put it there.
- **Category check:** the issue asks for two unasserted properties — (A) the fields of the SELECTED suite are the ones handed to `deployAndBroadcast`, (B) `deployNetworks()` is what is broadcast to. Both covered, by the two mutants above respectively. The issue's proposed fix additionally sketched (C) an `artifactPath` assertion and (D) an "other suite untouched" assertion; (C) is dropped because the issue's own verification block rules it vacuous, and (D) because it is factually impossible against a real fork — both explained under "Two legs ... deliberately not here" below.

## Two legs of the issue's proposed fix are deliberately not here

**The `artifactPath` assertion.** The issue's own verification block already rules it vacuous: it restates `testSelectedSuiteCarriesTheRecordedPins` verbatim and pins nothing about `run()`. `contractPath` reaches only a `console2.log` of a manual verify command, so no success test can observe it.

**`assertEq(zoltuAddress(...address-registry-candidate...).code.length, 0)`** — the "other suite untouched" leg. As written it fails: `AddressRegistry`'s deterministic address `0x25aC2b82915f191dbE64e65BAeDDD68b97b68fe1` **is live on Arbitrum One**, so that address has code on the fork whatever `run()` did.

That is not a defect only in the assertion. The declaration carries exactly two distinct addresses, and the other one is already on chain — so with this fixture *no* assertion can distinguish "deployed the selected suite" from "looped the registry": the registry-loop mutant deploys `MockDeployableV2` once and takes the already-deployed skip branch on `AddressRegistry`, which is observationally identical.

The claim is instead carried by *which* suite the leg selects. `MockDeployableV2` is on no supported chain, so its address has code only because this run put it there — which is what makes `assertGt(...code.length, 0)` kill the wrong-suite mutant rather than pass for free. The comment at the assertion says so, so the choice cannot be undone by accident.

## Test run

215 tests, all passing, against the RPC endpoints `CLAUDE.md` documents. (A first run used a non-archive Base endpoint and failed 7 archive-dependent `findDeployBlock`/`isStartBlock` tests with `pruned history unavailable`; all 13 pass on `https://mainnet.base.org`.)

Nothing broadcasts off-machine: `vm.startBroadcast` under `forge test` executes the deploy against the fork and records a transaction it never sends, which is the same thing the sibling `LibRainDeployTest` has always done with this library on this path.

## Summary by CodeRabbit

* **Tests**
  * Expanded deployment coverage to validate the complete broadcast flow.
  * Added verification that deployments target Arbitrum One and use the selected configuration.
  * Added checks for the deployed mock’s address, code, and final network.
  * Added a single-network deployment fixture for Arbitrum One.

### DIFF
--- a/test/concrete/ExampleDeploySingleNetwork.sol
+++ b/test/concrete/ExampleDeploySingleNetwork.sol
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: LicenseRef-DCL-1.0
+// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+pragma solidity =0.8.25;
+
+import {RainDeployBroadcast} from "../../src/abstract/RainDeployBroadcast.sol";
+import {LibRainDeploy} from "../../src/lib/LibRainDeploy.sol";
+import {ExampleDeploySuites} from "../abstract/ExampleDeploySuites.sol";
+
+/// @title ExampleDeploySingleNetwork
+/// A repo that bootstraps one chain per dispatch, which is the case
+/// `deployNetworks()` is overridable FOR — `st0x.deploy` selects between
+/// Ethereum and HyperEVM per dispatch rather than fanning out to everything.
+///
+/// The same declaration as `ExampleDeploy` and a different target set, so the
+/// only thing a broadcast through this fixture can differ in is where it went.
+/// A run that reached a network this does not name is a suite deployed
+/// somewhere the repo did not ask for, and the override being ignored is not
+/// observable any other way: `deployNetworks()` returning the right list says
+/// nothing about `run()` consulting it.
+///
+/// ARBITRUM rather than any other alias, deliberately. It is the network
+/// `supportedNetworks()` lists FIRST and this returns ONLY, and
+/// `deployToNetworks` forks each network in turn without restoring — so the
+/// chain left selected after a completed run is arbitrum here and polygon for
+/// the default, which is the difference an assertion can see. A single-element
+/// override naming polygon would end on polygon either way and assert nothing.
+contract ExampleDeploySingleNetwork is ExampleDeploySuites, RainDeployBroadcast {
+    /// @inheritdoc RainDeployBroadcast
+    function deployNetworks() internal pure override returns (string[] memory networks) {
+        networks = new string[](1);
+        networks[0] = LibRainDeploy.ARBITRUM_ONE;
+    }
+}
--- a/test/src/abstract/RainDeployBroadcast.t.sol
+++ b/test/src/abstract/RainDeployBroadcast.t.sol
@@ -7,17 +7,37 @@ import {Test} from "forge-std-1.16.1/src/Test.sol";
 import {UnknownDeploymentSuite} from "../../../src/abstract/RainDeploySuitesBase.sol";
 import {LibRainDeploy} from "../../../src/lib/LibRainDeploy.sol";
 import {ExampleDeploy} from "../../concrete/ExampleDeploy.sol";
+import {ExampleDeploySingleNetwork} from "../../concrete/ExampleDeploySingleNetwork.sol";
+import {MockDeployableV2} from "../../concrete/MockDeployableV2.sol";
 
 /// @title RainDeployBroadcastTest
 /// @notice The broadcast entry point, driven exactly as the `Manual sol
 /// artifacts` workflow drives it — through `DEPLOYMENT_SUITE` and
 /// `DEPLOYMENT_KEY` env vars.
 ///
-/// Nothing here broadcasts. Every case is one that fails before
-/// `deployAndBroadcast` is reached, which is the half of `run()` that can be
-/// tested without a key, an RPC and real money — and, not coincidentally, the
-/// half that decides WHAT would be deployed.
+/// Both halves of `run()` are driven here: the selection, through the cases
+/// that fail before `deployAndBroadcast` is reached, and the broadcast itself,
+/// on a fork.
+///
+/// The broadcast half is not the hazard it sounds like, and this file used to
+/// say it was. `vm.startBroadcast` under `forge test` executes the deploy on
+/// the fork and records a transaction it never sends, so a success case here
+/// costs a test key that has never held anything, the same public RPC every
+/// other fork test uses, and no money — which is exactly what the sibling
+/// `LibRainDeployTest` has always relied on to drive this same library through
+/// this same path. Leaving it untested did not avoid a risk; it left the org's
+/// only broadcast entry point, inherited by every downstream deploy repo,
+/// asserting nothing about what it would deploy or where.
 contract RainDeployBroadcastTest is Test {
+    /// Chain id of Arbitrum One, which is what `deployNetworks()` being honoured
+    /// looks like from inside a completed run.
+    ///
+    /// The literal rather than a second fork of the same alias to compare
+    /// against: an `arbitrum` RPC alias misconfigured onto some other chain
+    /// would agree with itself and quietly assert nothing, while a fact of the
+    /// world fails loudly.
+    uint256 constant ARBITRUM_ONE_CHAIN_ID = 42161;
+
     ExampleDeploy internal sDeploy;
 
     /// A deploy repo's whole script: the fixture declaration plus
@@ -74,7 +94,19 @@ contract RainDeployBroadcastTest is Test {
     /// asserted the ordering only in a bare shell — and a key that PARSES makes
     /// both orderings produce the same revert, so it would assert nothing about
     /// ordering in the one place the suite actually runs.
-    function testRunSelectsTheSuiteFromTheEnvBeforeTheKeyAndNeverDefaults() external {
+    ///
+    /// ## And then the suite it names, actually broadcast
+    ///
+    /// The two legs above decide WHAT would be deployed. They say nothing about
+    /// the nine lines that carry that decision to a chain, and those nine lines
+    /// are the entire body of `script/Deploy.sol`. A third leg runs `run()` all
+    /// the way through on a fork, so that the selected suite's fields, and the
+    /// target set they go to, are observed rather than assumed.
+    ///
+    /// Here rather than in a test of its own for the reason the first two are
+    /// here: this leg has to SET `DEPLOYMENT_SUITE`, and a second test doing
+    /// that is a second test racing the first over one process-global variable.
+    function testRunSelectsTheSuiteFromTheEnvBeforeTheKeyNeverDefaultsAndBroadcastsIt() external {
         // `DEPLOYMENT_SUITE` has to be ABSENT and no cheatcode makes it so, so
         // the precondition is asserted: a value set outside this test reports
         // itself by name here rather than as a surprising revert payload.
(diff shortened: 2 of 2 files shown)


thedavidmeister:
Reviewed 3a7e6b2: ready — Closes #58. The fixture choice is the whole discriminating power and it is argued correctly. `deployNetworks()` returning the right list says nothing about `run()` consulting it, so the override has to be observable in where the broadcast actually went — and since `deployToNetworks` forks each network in turn without restoring, the chain left selected after a completed run is the LAST one. Arbitrum is first in `supportedNetworks()` and the only entry here, so the two runs end on different chains and an assertion can see it. A single-element override naming polygon would end on polygon either way and assert nothing. That reasoning is stated in the fixture rather than left for a reader to reconstruct.

Modelling a real consumer shape — one chain per dispatch, as `st0x.deploy` does — rather than inventing one is the right fixture.

PR #91, opened 2026-08-15
Refuse to cut a release while ADDRESS_REGISTRY_ROOT is zero

### LINKED ISSUE #85: Nothing stops the inert `ADDRESS_REGISTRY_ROOT == address(0)` build being frozen into a permanent release
**Audit finding `HAZ6-05`** — dimension 6, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`script/Build.sol:114-122`

## Problem

HAZARD: `AddressRegistry` is deliberately compiled inert during rollout, and the step that makes it useful — setting a real root — is a source edit a human must remember, at a moment nothing marks. The release path has no guard against cutting the inert build, and a release is permanent.

SHAPE: `src/concrete/AddressRegistry.sol` sets `ADDRESS_REGISTRY_ROOT = address(0)` and documents it as "zero during rollout": no name can be bound, so `get` reverts on every read, forever, for that build. `Build.cutRelease()` freezes whatever `generatedContracts()` names, unconditionally. Once frozen, `LibAddressRegistryReleased.releasedSuites()` declares `address-registry@<tag>` and `RainDeployVerifyChain.testSuitesLiveOnEverySupportedNetwork` requires that address to carry that codehash on every network in `supportedNetworks()` — for the life of the repo, including every network added later. Setting a real root is "an ordinary source change" per the file's own doc, but it moves the creation code, so it produces a SECOND registry at a different address; the inert one does not go away, it becomes a released suite that must be kept alive alongside the real one. No test asserts the root is set, no issue or gate tracks the outstanding step, and `RegistryDeploySnapshotTest` passes either way because the pins simply follow the root.

PRODUCTION SCENARIO: the deploy dispatch that this repo is currently blocked on runs, `AddressRegistry` (zero root) and `MigrationRegistry` land on all five chains, `sol-v0.1.6` is pushed, and both are frozen into `src/generated/0_1_6/` and published. A consumer team installs `rain-deploy~0.1.6` and writes `sOwner = LibAddressRegistry.resolve("owner")` in its constructor — the documented pattern, resolve once and store. The codehash check passes (the registry IS there, with exactly the pinned code) and `get` reverts `NameNotRegistered`, so the consumer's own deterministic deploy reverts on every chain. The error reads as "ask root to bind the name", sending the team after a root that does not exist and cannot be created. Fixing it means setting the root here, which moves the registry address, which needs a fresh five-chain broadcast and a fresh release — while `0_1_6`'s inert registry stays in `releasedSuites()` and must remain live on every network forever, including chains added years later, or main goes red.

STRUCTURAL FIX: make the inert build impossible to release, so the manual step cannot be forgotten past the point of no return, while leaving it fully deployable and testable during rollout.

## Proposed fix

`script/Build.sol` — refuse to cut a release while the registry is inert. This gates only `cutRelease()`, so `run()`, the tests and the manual deploy all keep working unchanged during rollout, and main stays green today.

```diff
+import {ADDRESS_REGISTRY_ROOT} from "../src/concrete/AddressRegistry.sol";

+/// Thrown when a release is cut while `ADDRESS_REGISTRY_ROOT` is zero. A
+/// registry compiled under a zero root can never bind a name, so `get` reverts
+/// on every read for that build — and freezing it makes it a released suite the
+/// chain group requires to be live on every supported network forever, while
+/// setting a real root later produces a DIFFERENT address consumers must be
+/// migrated to. Deployable during rollout, not releasable.
+error InertRegistryRelease();

     function cutRelease() external {
+        if (ADDRESS_REGISTRY_ROOT == address(0)) {
+            revert InertRegistryRelease();
+        }
         GeneratedContract[] memory contracts = generatedContracts();
```

And a test, in `test/src/concrete/AddressRegistryGet.t.sol` or a new `test/src/lib/BuildCutRelease.t.sol`:

```solidity
    /// A release MUST NOT be cuttable while the root is zero: a registry that
    /// can never answer a read must not become a suite the chain group requires
    /// to be live on every network forever.
    function testCutReleaseRefusesAZeroRoot() external {
        vm.expectRevert(InertRegistryRelease.selector);
        Build(buildScript).cutRelease();
        // Nothing written: refused before `freeze` touches the filesystem.
        assertFalse(vm.exists(LibRainDeploySnapshot.dirForSnapshot(LibRainDeploySnapshot.deployTag(vm))));
    }

    /// The rollout state MUST still be fully usable: `run()` regenerates, the
    /// snapshot assertions hold, and the suite is dispatchable for a broadcast.
    function testZeroRootStillRegeneratesAndDeploys() external {
        Build(buildScript).run();
        assertTrue(vm.exists(LibRainDeploySnapshot.pathForSnapshot(LibRainDeploySnapshot.CANDIDATE, "AddressRegistry")));
    }

    /// The guard MUST be about the root being ABSENT, not about one particular
    /// value: any non-zero root releases.
    function testCutReleaseAcceptsANonZeroRoot(address root) external {
        vm.assume(root != address(0));
        // compile-time constant, so exercised through a fixture Build whose
        // root is injected; asserts no InertRegistryRelease.
    }
```

If the rollout deliberately intends to release the inert build, the finding still stands and the fix is the other structural one: give `AddressRegistry`'s zero-root state an explicit tracking artifact — an issue linked from `src/concrete/AddressRegistry.sol` naming the redeploy-and-migrate work that setting a root creates — rather than leaving it to whoever next reads the constant.

---

<details>

Verified against source; could not refute.

Mechanics confirmed:
- /home/gildlab/code/rain-deploy-audit/src/concrete/AddressRegistry.sol:22 — `address constant ADDRESS_REGISTRY_ROOT = address(0);`, with the doc at lines 12-21 stating it is zero during rollout, that no name can be bound, and that `get` reverts on every read. `register` (line 55) gates on `msg.sender != ADDRESS_REGISTRY_ROOT`; `get` (line 74) reverts `NameNotRegistered` on a zero binding. The contract is genuinely inert on chain — not a misread.
- /home/gildlab/code/rain-deploy-audit/script/Build.sol:114-122 — `cutRelease()` is unconditional: it builds `contractNames` from `generatedContracts()` and calls `LibRainDeploySnapshot.freeze`. No root check, and nothing upstream adds one: .github/workflows/package-release.yaml runs `forge script ./script/Build.sol --sig "cutRelease()"` on any `sol-v*` tag push.
- `LibRainDeploySnapshot.freeze` (src/lib/LibRainDeploySnapshot.sol:742-771) refuses only an already-frozen tag, an empty name list, or a missing candidate file. Nothing about contract content.
- The permanence claim holds: `RainDeployVerifyChain.testSuitesLiveOnEverySupportedNetwork` derives from `releasedSuites()` and requires code with the derived hash on every network in `supportedNetworks()`; its own doc names "a chain added to supportedNetworks() after a release" as exactly the case it catches. `RegistryDeploySuites.releasedSuites()` concatenates the generated `LibAddressRegistryReleased` entries, so a frozen inert build is required live on every present and future supported network.

Grepped the test tree: no test asserts a non-zero root, and none can fail on the state. Every test reaches root by pranking the constant (`vm.prank(ADDRESS_REGISTRY_ROOT)` in test/src/concrete/AddressRegistryRegister.t.sol, AddressRegistryGet.t.sol, test/src/lib/LibAddressRegistry.t.sol, LibRainDeploy.t.sol), which succeeds under a zero root and would succeed under any other value — the suite is root-agnostic by construction. test/src/abstract/RegistryDeploySnapshot.t.sol:17-19 explicitly documents that the pins simply follow the root. `RegistryDeployChainTest` has no subject until a release is cut. No issue, TODO or tracking artifact references the outstanding root step; the only mentions are descriptive prose in AddressRegistry.sol, CLAUDE.md and README.md.

Refutation attempts that failed:
- "Deliberate documented convention": the docs deliberately document the inert COMPILE state and that setting a root later moves address/codehash/snapshot. They nowhere state an intent to freeze the inert build into a release, and nothing documents the consequence the finding names — that the dead registry becomes a suite required live on every network forever.
- "A guard elsewhere handles it": the only downstream gate is rainix-tag-release's chain verification, which PASSES for an inert registry that was actually broadcast — codehash is checked, functionality is not and cannot be.
- "Human gates make it unreachable": both gates (workflow_dispatch broadcast, tag push) are precisely the unmarked manual moments the finding is about.

Caveat on the proposed fix (does not affect validity): `freeze` writes every contract in `generatedContracts()` into one tag directory, so gating `cutRelease()` on the root would also block releasing `MigrationRegistry`, which is fully functional today and has no rollout state. The finding does not acknowledge that coupling.

Severity: MEDIUM stands. No funds at risk and every failure is loud (a consumer's constructor reverts at its own deploy, before value moves), which caps damage — but the act is irreversible: a published soldeer release, five paid broadcasts, a permanent liveness obligation on every future chain, and a second registry address consumers must be moved to. Irreversible + ungated + untested is MEDIUM; not HIGH because nothing silently resolves wrong and no deployed value can be lost.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/85 (audit finding `HAZ6-05`, MEDIUM).

## The problem

`cutRelease()` froze whatever `generatedContracts()` named, unconditionally. `AddressRegistry` compiles with `ADDRESS_REGISTRY_ROOT = address(0)` during rollout, so nothing stopped the inert build — one that can never bind a name and reverts every read — from becoming a permanent released suite. `RainDeployVerifyChain.testSuitesLiveOnEverySupportedNetwork` then requires that address to carry that codehash on every network in `supportedNetworks()` for the life of the repo, including networks added later.

Setting a real root afterwards does not retire it. The root is a compile-time constant in the creation code, so the working registry is a **different** address: the dead one stays declared, stays required to be live, and stays what a consumer pinned to that release resolves against.

## The fix

`checkReleasableRoot` refuses a zero root, called as the first statement of `cutRelease()` — before anything is read or written, because filesystem cheatcodes are not undone by a revert and a partial freeze wedges the tag.

Only the irreversible step is gated. `run()`, the tests and the manual broadcast are untouched, so rollout is unaffected and main stays green today.

The gate covers the whole release rather than `AddressRegistry` alone. This is the coupling the issue's verification block flagged as unacknowledged — it is not a regression: a release IS the whole of `generatedContracts()` frozen into one tag, so there is no per-contract release that could let `MigrationRegistry` through, and inventing one would mean a tag whose contents depend on which contracts happened to be ready. That reasoning is now on the error's doc rather than left implicit.

## Departures from the issue's proposed fix

Both are traced to reasoning already recorded in this repo.

1. **The predicate is split out of `cutRelease()`** instead of inlined. The issue's own third test — "any non-zero root releases" — is unwritable against a compile-time constant, and it was left as a stub for that reason. Splitting it makes the rule exercisable at every root without adding a `virtual` hook a subclass could override to bypass the gate. The mutation matrix below shows this is what catches the unconditional mutant.

2. **No `testZeroRootStillRegeneratesAndDeploys`.** It calls `Build.run()`, which writes `src/generated/candidate/`. `foundry.toml:39-42` records why the existing snapshot test uses `test/generated` instead: forge runs suites in parallel and other contracts read that root, so writing there races them.

The wiring test asserts `ADDRESS_REGISTRY_ROOT == address(0)` **before** calling `cutRelease()`. Under a real root that call succeeds — and succeeding is what freezes a tag — so without the leading assertion the test would cut a live release on its way to failing. It also asserts the record directory's existence is *unchanged* rather than absent, since "absent" would conflate the gate not writing with no release ever having been cut.

That test is the marker for the rollout step itself: the moment a real root is set it goes red, and its doc says to delete it rather than repair it, because a real root is releasable and that is the point of the gate.

The test lives at `test/script/Build.t.sol`, mirroring `script/Build.sol`. The issue suggested `test/src/lib/BuildCutRelease.t.sol`, which would put a script test under `src/lib`.

## Docs

Updated where the rollout fact already lives: `src/concrete/AddressRegistry.sol`, `README.md` step 3, `CLAUDE.md`. Leaving README's release runbook saying "push a tag" would have been actively wrong once the tag is refused.

## QA

- **Discriminating tests:** `testCheckReleasableRootRefusesNoRoot`, `testCheckReleasableRootAcceptsEveryOtherRoot` (fuzz, 256 runs), `testCutReleaseRefusesTheInertRoot` — all three fail on base, because `InertRegistryRelease` and `checkReleasableRoot` do not exist there and the file does not compile against `script/Build.sol` as it stands on `main`. The behavioural equivalent on base is that `cutRelease()` returns normally and freezes a release under a zero root, which is the defect.
- **Mutations applied:** run with `mutation-probe` (`nix run github:rainlanguage/adversarial-mutation-test#mutation-probe`), baseline green at 3 passed, **5/5 killed, 0 survived, 0 no-run, 0 harness errors**:
  - A — `checkReleasableRoot(ADDRESS_REGISTRY_ROOT);` deleted from `cutRelease` → killed by `testCutReleaseRefusesTheInertRoot` **only**
  - B — `if (root == address(0))` → `if (root != address(0))` → killed by all three
  - C — `if (root == address(0))` → `if (true)` → killed by `testCheckReleasableRootAcceptsEveryOtherRoot` **only**
  - D — gate moved to after `LibRainDeploySnapshot.freeze(...)` → killed by `testCutReleaseRefusesTheInertRoot` **only**
  - E — `if (root == address(0))` → `if (root == address(1))` → killed by `testCheckReleasableRootRefusesNoRoot` + `testCutReleaseRefusesTheInertRoot`

  C is caught only by the fuzz test over non-zero roots — it would have survived had the predicate stayed inlined in `cutRelease()`, which is the reason for departure 1. D still satisfies `vm.expectRevert` (the moved gate does revert), so it is caught only by the record-untouched assertion, confirming that assertion is load-bearing rather than decoration.

  The first probe run had a `fail-pattern` that also matched `[PASS]` lines, so its per-test attribution was garbage; it was corrected and re-run. The KILLED verdicts were never affected, as those come from the suite's own pass/fail tally. The probe harness restores `src/` either side of each run, because a mutant that disables the gate lets `cutRelease()` actually freeze a tag and filesystem cheatcodes are not undone by a revert.
- **Oracle:** the issue's own specification of the hazard, plus this repo's recorded invariants — `RainDeployVerifyChain` requiring every released suite live on every supported network, `LibRainDeploySnapshot.freeze`'s documented "every guard runs before `<tag>/` is created" ordering rule, and `AddressRegistry`'s own doc that a zero root can never bind a name. Expected behaviour was derived from those, not from the new code.
- **Category check:** the issue asks for (a) the inert build to be unreleasable, (b) rollout to stay deployable and testable, (c) a test that a zero root is refused, (d) a test that rollout still works, (e) a test that any non-zero root releases. Covered a, b, c, e. (d) is deliberately not covered by a test — see departure 2; `Build.run()` writing `src/generated/candidate/` races the suites that read it, per `foundry.toml:39-42`. (b) is instead established by the gate touching only `cutRelease()`, with the full suite unchanged.

`forge fmt --check` clean on the restored tree. Full suite: 189 passed; the only failures are fork tests hitting public-RPC 429 rate limits, with zero assertion failures anywhere — CI runs those against rainix's provisioned endpoints.


thedavidmeister:
@coderabbitai review

thedavidmeister:
Closing: the premise is wrong. Cutting a release while `ADDRESS_REGISTRY_ROOT` is zero is the intended path, not an accident to guard against — that has been decided, more than once, and this guard would refuse exactly the release we are cutting.

Not a defect in the diff. The issue it closes (#85) rests on the same premise and should not produce another PR.

PR #92, opened 2026-08-15
Refuse a post-deploy check that has no reads

### LINKED ISSUE #61: A post-deploy check with zero reads passes on every network and no test pins it
**Audit finding `cov-08`** — dimension 2, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:272-349`

## Problem

`checkResolvedAddresses` refuses a length mismatch and then loops; with `readCalls.length == expectedAddresses.length == 0` the mismatch guard passes and the loop asserts nothing. `checkResolvedAddressesOnNetworks` forks every network and reports success having read nothing.

No test covers it: `testCheckResolvedAddressesLengthMismatchReverts` fuzzes `uint8 readCallsLength, uint8 expectedLength` under `vm.assume(readCallsLength != expectedLength)`, so the equal-and-zero case is excluded by construction, and every other test passes a one- or two-element list.

This is the same hazard `NoNetworks` exists for one argument along — `testCheckResolvedAddressesOnNetworksNoNetworksReverts` says "an empty target set can never be mistaken for every read checking out" — and the read list has no equivalent guard. A consumer that builds `readCalls` from a list that came back empty (a config, a loop over a set of resolved names) gets a green "deployment verified across every network" that verified nothing, immediately before migrating onto that deployment.

## Proposed fix

1. Fail closed on the empty read set, mirroring `NoNetworks` (`src/lib/LibRainDeploy.sol`):

```solidity
    /// Thrown when a post-deploy check is given no reads. A check with nothing
    /// to read passes on every network having asserted nothing, which is
    /// indistinguishable from every read checking out.
    error NoResolvedAddressReads(address target);
```

and at the top of `checkResolvedAddresses`, before the pairing check:

```solidity
        if (readCalls.length == 0) {
            revert NoResolvedAddressReads(target);
        }
```

2. Add to `test/src/lib/LibRainDeploy.t.sol`:

```solidity
    /// A check with no reads MUST be refused rather than pass. It is the same
    /// hazard `NoNetworks` covers one argument along: an empty read set is
    /// indistinguishable from every read checking out, and it is what a
    /// consumer that built its read list from an empty config hands in — right
    /// before it migrates onto the deployment this was supposed to verify.
    function testCheckResolvedAddressesNoReadsReverts(address target) external {
        vm.expectRevert(abi.encodeWithSelector(LibRainDeploy.NoResolvedAddressReads.selector, target));
        this.externalCheckResolvedAddresses("test_network", target, new bytes[](0), new address[](0));
    }

    /// And refused before any network is forked, so an empty read set is
    /// reported without an RPC round trip and cannot be masked by an outage.
    function testCheckResolvedAddressesOnNetworksNoReadsRevertsBeforeForking() external {
        string[] memory networks = new string[](1);
        // Not a configured RPC alias, so forking it is itself an error.
        networks[0] = "unconfigured_network";

        vm.expectRevert(abi.encodeWithSelector(LibRainDeploy.NoResolvedAddressReads.selector, address(this)));
        this.externalCheckResolvedAddressesOnNetworks(networks, address(this), new bytes[](0), new address[](0));
    }

    /// The refusal is about the READS, not about the pairing: an empty pair is
    /// the one length mismatch cannot see, which is why it is its own error.
    function testCheckResolvedAddressesEmptyPairIsNotALengthMismatch(bytes32 name, address account) external {
        vm.assume(account != address(0));
        (, MockResolvedOwner consumer) = deployRegistryAndConsumer(name, account);

        vm.expectRevert(
            abi.encodeWithSelector(LibRainDeploy.NoResolvedAddressReads.selector, address(consumer))
        );
        this.externalCheckResolvedAddresses("test_network", address(consumer), new bytes[](0), new address[](0));

        // One read still passes, so the refusal is about emptiness alone.
        LibRainDeploy.checkResolvedAddresses("test_network", address(consumer), ownerReadCalls(), expected(account));
    }
```

Update `checkResolvedAddressesOnNetworks` to run the new guard before the fork loop (it already runs the pairing check there for the same reason).

---

<details>

Verified against source and tests; could not refute.

Behaviour exists as described. `/home/gildlab/code/rain-deploy-audit/src/lib/LibRainDeploy.sol:272-309`: `checkResolvedAddresses` guards only `readCalls.length != expectedAddresses.length`, then `for (uint256 i = 0; i < readCalls.length; i++)`. With both lengths zero the guard passes and the loop body never runs, so the function returns having asserted nothing. `checkResolvedAddressesOnNetworks` (lines 326-349) guards `networks.length == 0` with `NoNetworks` and repeats only the pairing check before the fork loop, then forks every network, logs "Checking resolved addresses on network: …" per network and returns success. Errors list at lines 17-55 contains no empty-read error; no natspec, README (`README.md:173`) or CLAUDE.md text declares an empty read set deliberately allowed.

No test covers it. `test/src/lib/LibRainDeploy.t.sol:959` `testCheckResolvedAddressesLengthMismatchReverts(uint8, uint8)` opens with `vm.assume(readCallsLength != expectedLength)`, so equal-and-zero is excluded by construction. Every other call site uses `ownerReadCalls()` (1 element) or a hand-built 2-element pair (line 857 onwards). Grep across `test/` and `src/` shows the only callers of `checkResolvedAddresses*` are the lib itself and this test file — nothing passes `new bytes[](0)`, so nothing pins either the pass or a refusal. The `NoNetworks` sibling at line 975 is exactly the analogous guard the finding names, and its own comment ("an empty target set can never be mistaken for every read checking out") plus the `NoDeployCandidates` convention in CLAUDE.md ("refusing an empty list, because a candidate the loop never reaches is a contract whose snapshot nothing anywhere anchors") make fail-closed-on-empty the stated house rule rather than an invented one.

Two caveats that do not refute it but bear on the fix. (1) The proposed fix as written breaks the existing fuzz test: placing `if (readCalls.length == 0) revert NoResolvedAddressReads(target)` above the pairing check makes fuzz cases like `(0, 5)` revert with `NoResolvedAddressReads` where `testCheckResolvedAddressesLengthMismatchReverts` expects `ResolvedAddressesLengthMismatch`. The guard must run after the pairing check (or the existing test must exclude `readCallsLength == 0`). (2) `checkResolvedAddresses` is `internal view` and does not take `target` ordering into account for the on-networks variant — the new guard must also be added explicitly before the fork loop in `checkResolvedAddressesOnNetworks`, as the fix says.

Severity corrected MEDIUM -> LOW. Value-at-risk: this is a post-deploy verification helper published for consumer deploy scripts; reaching the vacuous pass requires a consumer to hand in an empty read list it built dynamically, and the deployment's primary anchors (derived address, codehash and runtime-code pins across `RainDeployVerifySnapshot`/`RainDeployVerifyChain`) are unaffected — the false green is confined to the supplementary "does the deployed contract hold the addresses it resolved" assertion. No in-repo caller can hit it today.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/61 (audit finding `cov-08`).

## The hole

`LibRainDeploy.checkResolvedAddresses` guarded only that `readCalls` and `expectedAddresses` pair up, then looped over them:

```solidity
if (readCalls.length != expectedAddresses.length) { revert ResolvedAddressesLengthMismatch(...); }
for (uint256 i = 0; i < readCalls.length; i++) { ... }
```

With both lengths zero the pairing guard passes and the loop body never runs, so the function returns having asserted nothing. `checkResolvedAddressesOnNetworks` then forks every network, logs `Checking resolved addresses on network: …` for each, and returns success across all of them having read nothing at all.

That is the same hazard `NoNetworks` already exists for one argument along — its own test says "an empty target set can never be mistaken for every read checking out" — and the read list had no equivalent. A consumer that builds `readCalls` from a source that came back empty (a config, a loop over resolved names) gets a green "verified on every network" immediately before migrating onto that deployment.

Nothing pinned it either way: `testCheckResolvedAddressesLengthMismatchReverts` fuzzes under `vm.assume(readCallsLength != expectedLength)`, so equal-and-zero is excluded by construction, and every other call site passed a one- or two-element list.

## The fix

New error, fail-closed on the empty read set, matching the `NoNetworks` / `NoDeployCandidates` house rule:

```solidity
error NoResolvedAddressReads(address target);
```

**The guard runs after the pairing check, not before it.** An unpaired call is a mispairing whichever side happens to be empty, and reporting the empty pair as a "mismatch of zero against zero" would say nothing. So `(0, n)` stays a `ResolvedAddressesLengthMismatch` naming both lengths, and only the empty PAIR — the one case pairing cannot see — raises the new error. Hoisting it above the pairing check would swallow real mispairings; mutant M4 below is exactly that placement, and it is killed.

`checkResolvedAddressesOnNetworks` repeats the guard before its fork loop, for the reason it already repeats the pairing check there: an empty read set is reported without an RPC round trip, so a network outage cannot mask it.

## Tests

Three new tests in `test/src/lib/LibRainDeploy.t.sol`:

- `testCheckResolvedAddressesNoReadsReverts` — an empty read set is refused rather than passing.
- `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch` — the empty pair raises the new error, empty-reads-against-a-non-empty-expected-list is still `ResolvedAddressesLengthMismatch(0, 1)`, and one real read still passes. The middle assertion deterministically pins the guard ORDER rather than leaving it to whether the fuzzer happens to draw a zero.
- `testCheckResolvedAddressesOnNetworksNoReadsRevertsBeforeForking` — refused before anything is forked, using an unconfigured RPC alias so reaching the fork loop is itself an error.

The pre-existing `testCheckResolvedAddressesLengthMismatchReverts` fuzz test is unchanged and still passes, which is the point of the ordering.

## QA
- Discriminating tests: `testCheckResolvedAddressesNoReadsReverts`, `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch`, `testCheckResolvedAddressesOnNetworksNoReadsRevertsBeforeForking` — each fails on base. Verified via mutants M1 and M5, which delete the added guards and so reproduce base behaviour exactly (literal base does not compile these tests, since `NoResolvedAddressReads` does not exist there); with the guard deleted, M1 is killed by the first two and M5 by the third.
- Mutations applied: 6 mutants via `mutation-probe`, **6/6 KILLED, 0 survived, 0 no-run, 0 harness errors**, against a green 13-test baseline.
  - M1 `checkResolvedAddresses`: delete the empty-read guard entirely → `testCheckResolvedAddressesNoReadsReverts`, `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch`
  - M2 `checkResolvedAddresses`: `readCalls.length == 0` → `== 1` → `testCheckResolvedAddressesNoReadsReverts`, `…Match`, `…MismatchReverts`, `…DirtyWordReverts`, `…EmptyPairIsNotALengthMismatch`
  - M3 `checkResolvedAddresses`: `NoResolvedAddressReads(target)` → `(address(0))` → `testCheckResolvedAddressesNoReadsReverts`, `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch`
  - M4 `checkResolvedAddresses`: hoist the guard ABOVE the pairing check → `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch` (deterministic) and `testCheckResolvedAddressesLengthMismatchReverts` (fuzz-dependent — it needs the fuzzer to draw `readCallsLength == 0`, which is why the deterministic assertion was added rather than relying on it)
  - M5 `checkResolvedAddressesOnNetworks`: delete the pre-fork empty-read guard → `testCheckResolvedAddressesOnNetworksNoReadsRevertsBeforeForking`
  - M6 `checkResolvedAddressesOnNetworks`: `NoResolvedAddressReads(target)` → `(address(0))` → `testCheckResolvedAddressesOnNetworksNoReadsRevertsBeforeForking`
  - The probe suite is deliberately scoped network-free (`--match-test CheckResolvedAddresses` minus the two fork tests) and was run with no `.env` present at all, so nothing in it can reach an RPC. A first pass over the full `forge test` produced 6/6 "killed" whose named killers were fork tests (`testChainMatrix*`, `testDeployToNetworks*`) that a guard inside `checkResolvedAddresses` cannot affect — flaky public-RPC failures being scored as kills. That matrix was discarded as untrustworthy rather than reported.
- Oracle: the invariant is stated independently of this implementation — an assertion helper that asserts nothing must not report success, which is the rule `NoNetworks` already encodes one argument along and `NoDeployCandidates` encodes for the candidate list ("a candidate the loop never reaches is a contract whose snapshot nothing anywhere anchors"). Expected revert data is built from the error signature via `abi.encodeWithSelector`, not read back from a run.
- Category check: the issue asks for (a) a fail-closed error on the empty read set, (b) the guard in `checkResolvedAddresses`, (c) the guard before the fork loop in `checkResolvedAddressesOnNetworks`, (d) tests covering all three. All four covered. One deliberate deviation from the issue's proposed patch, which the issue's own verification block flagged: the guard is placed AFTER the pairing check, not before it, so the existing length-mismatch fuzz test keeps its meaning for `(0, n)`.

Full suite green: 218 passed, 0 failed, 17 suites. `forge fmt --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject resolved-address checks when no reads are supplied.
  * Empty read sets now fail before network reads or fork creation.
  * Preserved distinct errors for empty inputs versus mismatched read and expected-value lists.

* **Tests**
  * Added coverage for empty inputs, mismatched lists, valid reads, and multi-network validation.

### DIFF
--- a/src/lib/LibRainDeploy.sol
+++ b/src/lib/LibRainDeploy.sol
@@ -50,6 +50,11 @@ library LibRainDeploy {
     /// do not pair up.
     error ResolvedAddressesLengthMismatch(uint256 readCallsLength, uint256 expectedAddressesLength);
 
+    /// Thrown when a post-deploy check is given no reads. A check with nothing
+    /// to read passes on every network having asserted nothing, which is
+    /// indistinguishable from every read checking out.
+    error NoResolvedAddressReads(address target);
+
     /// Thrown when a post-deploy read reverts, or answers with something that is
     /// not a single address-sized word.
     error ResolvedAddressReadFailed(string network, address target, uint256 index, bytes returnData);
@@ -263,6 +268,12 @@ library LibRainDeploy {
     /// Only the consumer knows where it stored what it resolved, so the consumer
     /// supplies the reads. Each entry in `readCalls` is static-called against
     /// `target` and MUST answer with exactly one address.
+    ///
+    /// An empty read set is REFUSED. A check with nothing to read returns having
+    /// asserted nothing, which is indistinguishable from every read checking
+    /// out, and it is what a consumer that built its read list from a source
+    /// that came back empty hands in — right before it migrates onto the
+    /// deployment this was supposed to verify.
     /// @param network The network name, for the error only.
     /// @param target The deployed contract to read.
     /// @param readCalls The calldata for each read, e.g.
@@ -277,6 +288,13 @@ library LibRainDeploy {
     ) internal view {
         if (readCalls.length != expectedAddresses.length) {
             revert ResolvedAddressesLengthMismatch(readCalls.length, expectedAddresses.length);
+        }
+        // After the pairing check, not before it: an unpaired call is a
+        // mispairing whichever side is empty, and reporting the empty pair as a
+        // mismatch of zero against zero would say nothing. The empty pair is the
+        // one case pairing cannot see, so it is its own error.
+        if (readCalls.length == 0) {
+            revert NoResolvedAddressReads(target);
         }
         for (uint256 i = 0; i < readCalls.length; i++) {
             // The consumer supplies the reads, so the call is low level by
@@ -317,6 +335,10 @@ library LibRainDeploy {
     /// than expected is a burned deterministic address, found while nothing
     /// points at it yet — which is the whole reason to verify before migrating
     /// onto a deployment rather than trusting it.
+    ///
+    /// An empty network set and an empty read set are both REFUSED, before
+    /// anything is forked. Either one makes this return success across every
+    /// network having read nothing at all.
     /// @param vm The Vm instance to use for forking.
     /// @param networks The list of network names to check.
     /// @param target The deployed contract to read on each network.
@@ -337,6 +359,11 @@ library LibRainDeploy {
         // than after an RPC round trip.
         if (readCalls.length != expectedAddresses.length) {
             revert ResolvedAddressesLengthMismatch(readCalls.length, expectedAddresses.length);
+        }
+        // Same reason: an empty read set is reported without an RPC round trip,
+        // so it cannot be masked by an outage on the first network.
+        if (readCalls.length == 0) {
+            revert NoResolvedAddressReads(target);
         }
         for (uint256 i = 0; i < networks.length; i++) {
             // createSelectFork returns a fork id that is not needed here; bind
--- a/test/src/lib/LibRainDeploy.t.sol
+++ b/test/src/lib/LibRainDeploy.t.sol
@@ -972,6 +972,39 @@ contract LibRainDeployTest is Test {
         this.externalCheckResolvedAddresses("test_network", address(this), readCalls, expectedAddresses);
     }
 
+    /// A check with no reads MUST be refused rather than pass. It is the same
+    /// hazard `NoNetworks` covers one argument along: an empty read set is
+    /// indistinguishable from every read checking out, and it is what a
+    /// consumer that built its read list from an empty config hands in — right
+    /// before it migrates onto the deployment this was supposed to verify.
+    function testCheckResolvedAddressesNoReadsReverts(address target) external {
+        vm.expectRevert(abi.encodeWithSelector(LibRainDeploy.NoResolvedAddressReads.selector, target));
+        this.externalCheckResolvedAddresses("test_network", target, new bytes[](0), new address[](0));
+    }
+
+    /// The refusal is about the READS, not about the pairing: an empty pair is
+    /// the one case a length mismatch cannot see, which is why it is its own
+    /// error. Empty reads against a non-empty expected list stays a mismatch,
+    /// pinning that pairing is still checked first, and a single real read
+    /// still passes, so emptiness alone is what is refused.
+    function testCheckResolvedAddressesEmptyPairIsNotALengthMismatch(bytes32 name, address account) external {
+        vm.assume(account != address(0));
+        (, MockResolvedOwner consumer) = deployRegistryAndConsumer(name, account);
+
+        vm.expectRevert(abi.encodeWithSelector(LibRainDeploy.NoResolvedAddressReads.selector, address(consumer)));
+        this.externalCheckResolvedAddresses("test_network", address(consumer), new bytes[](0), new address[](0));
+
+        // Empty reads against a non-empty expected list is a mispairing, and is
+        // still reported as one.
+        vm.expectRevert(
+            abi.encodeWithSelector(LibRainDeploy.ResolvedAddressesLengthMismatch.selector, uint256(0), uint256(1))
+        );
+        this.externalCheckResolvedAddresses("test_network", address(consumer), new bytes[](0), expected(account));
+
+        // One read still passes, so the refusal is about emptiness alone.
+        LibRainDeploy.checkResolvedAddresses("test_network", address(consumer), ownerReadCalls(), expected(account));
+    }
+
     /// `checkResolvedAddressesOnNetworks` MUST revert with `NoNetworks` when
     /// given none, so an empty target set can never be mistaken for every read
     /// checking out.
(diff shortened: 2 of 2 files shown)


thedavidmeister:
@coderabbitai review

thedavidmeister:
Reviewed 6287d8c6a68a7519161a60415d9482da9a202cb0: ready — Closes #61. Read the diff independently against the issue.

Both guards land where the PR says: `NoResolvedAddressReads(target)` after the pairing check in `checkResolvedAddresses`, and again before the fork loop in `checkResolvedAddressesOnNetworks`. The placement is a deliberate improvement on the issue's proposed patch — `(0, 1)` stays a `ResolvedAddressesLengthMismatch` naming both lengths, and only the empty pair, the case pairing cannot see, raises the new error. `testCheckResolvedAddressesEmptyPairIsNotALengthMismatch` pins that order deterministically rather than leaning on the pre-existing fuzz test, which excludes equal-and-zero by construction under its `vm.assume`.

CI green (rainix static/legal/test), CodeRabbit reviewed, 0 unresolved threads. Mutation matrix 6/6 killed, and the first matrix was discarded rather than reported when its named killers turned out to be fork tests failing on flaky public RPCs — a guard inside `checkResolvedAddresses` cannot affect those.

Residual, not blocking and not a defect in this diff: the guard relocates the empty-read hazard to the caller rather than removing it. A consumer that may legitimately have zero reads — `MigrationRegistry` resolves nothing at all — now needs `if (readCalls.length > 0)` at the call site, and a config that came back empty takes that branch silently. Nothing in `src/` or `script/` calls either function, so both are consumer-facing entry points and the repo has no example of the caller shape this is designed against. The version that closes it is a declared read count on the suite, checked against the list handed in, so "resolves nothing" is stated rather than expressed by absence — the same move `NoDeployCandidates` makes one level up.

PR #93, opened 2026-08-15
test(build): cover the generator's declaration and the alias-lib emitter

### LINKED ISSUE #45: `script/Build.sol` has no test at all, and nothing asserts `generatedContracts()` covers every declared candidate
A coverage finding and a correctness finding on the same untested file. The second names the specific invariant the absent test would carry.

---

# Finding `cov-01`

**Audit finding `cov-01`** — dimension 2, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`script/Build.sol:96-151`

## Problem

`script/Build.sol` is the only thing that writes `src/generated/candidate/*.sol`, `src/lib/Lib*Deploy.sol` and `src/lib/Lib*Released.sol`, and nothing in `test/` constructs it or calls `run()`/`cutRelease()`. Greps for `Build`, `cutRelease`, `regenerateCandidates`, `regenerateLibs`, `writeAliasLib` across `test/` return only prose mentions in `GeneratedSnapshotShape.t.sol` and `ExampleDeploy.sol`.

Consequences that nothing catches today: (a) `run()`'s sequencing (`regenerateCandidates` then `regenerateLibs`) is never exercised, so the committed generated tree is never proven to be what the current generator emits — `LibRainDeploySnapshotTest.testWriteReleasedSuitesLibWritesTheLibAtItsPath` does exactly this regenerate-and-compare for ONE of the six generated files and there is no equivalent for the other five; (b) `cutRelease()` — a one-shot, non-idempotent filesystem mutation whose own docstring says a partial write wedges the release tag forever — will be executed for the very first time during a real release; (c) `generatedContracts()` is a second declaration of the deployed set beside `RegistryDeploySuites.candidateSuites()`, and no test compares them.

The file is also entirely inside the drift window: it did not exist at the ledgered mutation campaign commit (see cov-11).

## Proposed fix

New file `test/script/Build.t.sol` (mirror-by-subject of `script/Build.sol`):

```solidity
pragma solidity =0.8.25;

import {Test} from "forge-std-1.16.1/src/Test.sol";
import {Build} from "../../script/Build.sol";

/// @title BuildTest
/// @notice The generator itself. Nothing else runs it, so a committed
/// generated file that no longer matches what the generator emits is
/// invisible until somebody regenerates by hand.
contract BuildTest is Test {
    /// Every file `run()` writes.
    /// @return paths The generated files.
    function generatedPaths() internal pure returns (string[] memory paths) {
        paths = new string[](6);
        paths[0] = "src/generated/candidate/AddressRegistry.sol";
        paths[1] = "src/generated/candidate/MigrationRegistry.sol";
        paths[2] = "src/lib/LibAddressRegistryDeploy.sol";
        paths[3] = "src/lib/LibMigrationRegistryDeploy.sol";
        paths[4] = "src/lib/LibAddressRegistryReleased.sol";
        paths[5] = "src/lib/LibMigrationRegistryReleased.sol";
    }

    /// `run()` MUST reproduce the committed generated tree byte for byte, and
    /// MUST write every file the repo declares as generated. A difference here
    /// is a hand edit or a file nobody regenerated after the generator moved —
    /// the two failures the generated-file header exists to warn about, and
    /// which nothing else in the suite can see.
    ///
    /// Restored BEFORE the assertions, because forge-std assertions revert:
    /// restoring afterwards restores in every case except a failure, which is
    /// the only case where the tree is dirty.
    function testBuildRunRegeneratesTheCommittedTreeByteForByte() external {
        string[] memory paths = generatedPaths();
        string[] memory before = new string[](paths.length);
        for (uint256 i = 0; i < paths.length; i++) {
            before[i] = vm.readFile(paths[i]);
        }

        new Build().run();

        string[] memory regenerated = new string[](paths.length);
        for (uint256 i = 0; i < paths.length; i++) {
            regenerated[i] = vm.readFile(paths[i]);
        }
        for (uint256 i = 0; i < paths.length; i++) {
            //forge-lint: disable-next-line(unsafe-cheatcode)
            vm.writeFile(paths[i], before[i]);
        }

        for (uint256 i = 0; i < paths.length; i++) {
            assertEq(regenerated[i], before[i], paths[i]);
        }
    }

    /// `run()` MUST be idempotent: a second run over what the first wrote
    /// changes nothing, so a build is not a diff.
    function testBuildRunIsIdempotent() external {
        string[] memory paths = generatedPaths();
        string[] memory before = new string[](paths.length);
        for (uint256 i = 0; i < paths.length; i++) {
            before[i] = vm.readFile(paths[i]);
        }

        Build build = new Build();
        build.run();
        string[] memory once = new string[](paths.length);
        for (uint256 i = 0; i < paths.length; i++) {
            once[i] = vm.readFile(paths[i]);
        }
        build.run();
        string[] memory twice = new string[](paths.length);
        for (uint256 i = 0; i < paths.length; i++) {
            twice[i] = vm.readFile(paths[i]);
        }

        for (uint256 i = 0; i < paths.length; i++) {
            //forge-lint: disable-next-line(unsafe-cheatcode)
            vm.writeFile(paths[i], before[i]);
        }
        for (uint256 i = 0; i < paths.length; i++) {
            assertEq(twice[i], once[i], paths[i]);
        }
    }
}
```

The `cutRelease()` write path is covered by cov-02/cov-03 once `freeze` takes a record root; without that seam it cannot be driven anywhere but `src/generated/`, where a transient `<tag>/` directory races the record walk in the contracts forge runs in parallel.

---

<details>

SURVIVES, narrowed, severity corrected MEDIUM -> LOW.

What I verified as TRUE
- `script/Build.sol` (read in full) is the only writer of the six generated files, and `grep -rn "Build\|cutRelease\|regenerateCandidates\|regenerateLibs\|writeAliasLib\|freeze" test/` returns only prose mentions plus `LibRainDeploySnapshot.t.sol` driving the LIBRARY directly. No test constructs `Build` or calls `run()`/`cutRelease()`. There is no `test/script/` directory at all.
- The one genuinely uncovered code path this exposes: `LibRainDeploySnapshot.writeAliasLib`, `aliasImportBlock` and `aliasLibraryBlock` (src/lib/LibRainDeploySnapshot.sol:379) are referenced by NOTHING outside their own file — grep across `test/` and `src/` for all three returns zero hits. The alias-lib emitter is the only generator path in this repo with literally zero execution by the suite, and `run()`'s two-step sequencing is likewise never exercised. That is a real dimension-2 gap and I could not refute it.

What I REFUTED in the finding's reasoning
- Consequence (c) is wrong. `GeneratedSnapshotShapeTest.testEveryCandidateHasASnapshot` (test/src/lib/GeneratedSnapshotShape.t.sol) walks `src/generated/candidate/` — the directory only `generatedContracts()` writes — and asserts it is the SAME SET as `checkedCandidateSuites()`, matched by name in both directions plus counts. That is exactly the comparison the finding says does not exist. `generatedContracts()` also reaches its candidates through `addressRegistryCandidate()`/`migrationRegistryCandidate()`, the same named functions `candidateSuites()` uses, so it is not a second declaration of the metadata either (RegistryDeploySuites.sol documents this explicitly).
- Consequence (a) is largely overstated. "The other five generated files have no equivalent check" is false in substance: the two candidate snapshots are anchored by `RainDeployVerifySnapshot` group 1 (`testSnapshotInternallyConsistent` — stored address/hash/runtime derived from stored creation code) and group 2 (`testSnapshotMatchesSource` — stored creation code == `type(X).creationCode`), plus five AST-level shape properties in `GeneratedSnapshotShapeTest`; the two alias libs contain nothing but re-exports of those same anchored constants, and their values flow into `RegistryDeploySuites` where group 1 checks them, so a stale or hand-edited alias value is red; `LibMigrationRegistryReleased` is emitted by the identical code path that `testWriteReleasedSuitesLibWritesTheLibAtItsPath` already regenerate-and-compares against the real committed file and the real record root, and frozen-record drift is caught by group 3 (`testEveryFrozenSnapshotIsReleased`). Only the alias libs' TEXT is unproven.
- Consequence (b) is not addressed by the finding's own proposed fix, which explicitly defers `cutRelease()` to cov-02/cov-03. `freeze`'s guards and cleanup ordering are already driven directly (`testFreezeRefusesAnEmptyRelease`, `testFreezeLeavesNothingBehindWhenThereIsNothingToFreeze`), so what is untested there is the wiring, not the guards.

Severity
Value at risk in production is low, not medium. Every pin that ships is independently anchored to compiled source by three test groups that run on ordinary `forge test`, so a generator regression that emits a WRONG pin is red before any release. What is unprotected is a generated file whose text drifted while its semantics stayed correct (the alias libs), which costs a diff on the next build, not money on chain.

</details>

---

# Finding `DEPLOY-3`

**Audit finding `DEPLOY-3`** — dimension 5, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`script/Build.sol:82-92`

## Problem

script/Build.sol has no test coverage at all: no test constructs Build, and grep across test/ finds no reference to run(), cutRelease() or generatedContracts(). LibRainDeploySnapshot.t.sol covers the library primitives (writeSnapshot, writeReleasedSuitesLib, freeze), not this script's own declaration or sequencing.

The uncovered invariant is the one Build.sol's own NatSpec (lines 53-62) names as unnoticeable: 'a contract regenerated but left out of the names freeze is given is a contract silently absent from the release, which nothing downstream can notice, because a tag that never held it has nothing missing from it.' The file argues this cannot happen because there is one list — but the one list is generatedContracts(), while the deployment declaration is RegistryDeploySuites.candidateSuites(). They are two lists, in two files, and nothing checks they agree.

The reachable failure: a contract present in candidateSuites() but dropped from generatedContracts() (a refactor, a merge). Its committed snapshot still exists, so it compiles. cutRelease() then freezes only the contracts generatedContracts() names, and regenerateLibs() writes a released-suites lib only for those, so the release silently omits that contract. testEveryFrozenSnapshotIsReleased (RainDeployVerifySnapshot.sol:263-267) only walks record -> declaration, so it cannot see a record entry that was never written. RegistryDeployChainTest is never handed the omitted suite either. Every check stays green and the release is missing a contract.

GeneratedSnapshotShapeTest.testEveryCandidateHasASnapshot does NOT cover this: it compares the candidate DIRECTORY's files against the declared candidates, and the stale committed file makes both sides match.

## Proposed fix

Add test/script/Build.t.sol (mirror placement by subject; Build's subject is script/Build.sol). It needs an external seam because generatedContracts() is internal pure — a harness, not a change to Build.

pragma solidity =0.8.25;

import {Test} from "forge-std-1.16.1/src/Test.sol";
import {Build, GeneratedContract} from "../../script/Build.sol";
import {DeployCandidate} from "../../src/abstract/RainDeploySuitesBase.sol";

contract BuildHarness is Build {
    function externalGeneratedContracts() external pure returns (GeneratedContract[] memory) {
        return generatedContracts();
    }

    function externalCandidateSuites() external pure returns (DeployCandidate[] memory) {
        return checkedCandidateSuites();
    }
}

contract BuildTest is Test {
    BuildHarness internal sBuild;

    function setUp() external {
        sBuild = new BuildHarness();
    }

    /// PROPERTY: the generator's list and the deploy declaration are the SAME
    /// SET of contracts, matched both ways. A candidate the generator does not
    /// name is a contract silently absent from every release cut from here,
    /// and neither the frozen-record check nor the chain check can see it,
    /// because both are only ever handed what was written.
    ///
    /// Matched by the suite key rather than by index, because Build reaches
    /// candidates by name for exactly this reason and a positional pass would
    /// go green on a reordered list that had silently swapped two contracts.
    function testGeneratedContractsAreExactlyTheDeclaredCandidates() external view {
        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
        DeployCandidate[] memory candidates = sBuild.externalCandidateSuites();

        assertEq(generated.length, candidates.length, "a candidate is not generated, or a generated contract is not declared");

        for (uint256 i = 0; i < generated.length; i++) {
            bool found = false;
            for (uint256 j = 0; j < candidates.length; j++) {
                if (keccak256(bytes(generated[i].candidate.snapshot.suite)) == keccak256(bytes(candidates[j].snapshot.suite))) {
                    assertEq(
                        keccak256(generated[i].candidate.snapshot.creationCode),
                        keccak256(candidates[j].snapshot.creationCode),
                        "generated entry carries another contract's creation code"
                    );
                    found = true;
                    break;
                }
            }
            assertTrue(found, string.concat("generated contract is not a declared candidate: ", generated[i].contractName));
        }

        for (uint256 j = 0; j < candidates.length; j++) {
            bool found = false;
            for (uint256 i = 0; i < generated.length; i++) {
                if (keccak256(bytes(generated[i].candidate.snapshot.suite)) == keccak256(bytes(candidates[j].snapshot.suite))) {
                    found = true;
                    break;
                }
            }
            assertTrue(found, string.concat("declared candidate is not generated: ", candidates[j].snapshot.suite));
        }
    }

    /// PROPERTY: contractName is the name the snapshot path and both generated
    /// libs are built from, and it MUST be the contract the candidate's
    /// artifact path names. A disagreement writes one contract's pins into
    /// another contract's file, which every downstream check then reads as
    /// self-consistent.
    function testGeneratedContractNameMatchesTheArtifactPath() external view {
        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
        for (uint256 i = 0; i < generated.length; i++) {
            string[] memory components = vm.split(generated[i].candidate.snapshot.artifactPath, ":");
            assertEq(components[components.length - 1], generated[i].contractName, "generated name is not the artifact's contract");
        }
    }

    /// PROPERTY: no two entries name one contract. Two entries sharing a
    /// contractName write one snapshot file twice and freeze one file for two
    /// declarations, so the second silently replaces the first.
    function testGeneratedContractNamesAreUnique() external view {
        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
        for (uint256 i = 0; i < generated.length; i++) {
            for (uint256 j = i + 1; j < generated.length; j++) {
                assertNotEq(generated[i].contractName, generated[j].contractName, "two generated entries name one contract");
            }
        }
    }

    /// PROPERTY: the constant prefix is non-empty and distinct per contract.
    /// It names every constant both generated libs export, so a shared prefix
    /// is two libs exporting colliding names.
    function testConstantPrefixesAreNonEmptyAndUnique() external view {
        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
        for (uint256 i = 0; i < generated.length; i++) {
            assertGt(bytes(generated[i].constantPrefix).length, 0, "empty constant prefix");
            for (uint256 j = i + 1; j < generated.length; j++) {
                assertNotEq(generated[i].constantPrefix, generated[j].constantPrefix, "two entries share a constant prefix");
            }
        }
    }
}

Deliberately no test invokes run() or cutRelease(): both write under src/, and foundry.toml's own fs_permissions comment records that a fixture write into a root other contracts read in parallel is a race. The list-agreement property above is the part that is both untested and unnoticeable, and it is pure.

---

<details>

Could not refute. Verified against source at 440e90b: (1) script/Build.sol:82-92 builds its list from the named candidate accessors, independent of RegistryDeploySuites.candidateSuites() — two hard-coded arrays in two files, a split RegistryDeploySuites' own NatSpec acknowledges ("a third entry here, and a third entry in script/Build.sol's generated-contract list"); Build's "one list, three readers" is scoped to readers inside Build, not to the declaration. (2) grep -rn "Build" test/ returns exactly one hit, a doc comment in GeneratedSnapshotShape.t.sol; nothing constructs Build and nothing references run/cutRelease/generatedContracts/regenerateLibs/regenerateCandidates outside the script — the "no coverage" claim is accurate. (3) LibRainDeploySnapshot.freeze (line 742) freezes exactly the contractNames handed to it and never walks the rolling directory; its guards are SnapshotAlreadyFrozen, EmptyRelease (length 0) and NothingToFreeze (a named contract lacking a rolling file) — none of which can see a candidate that was simply not named. (4) RainDeployVerifySnapshot.testEveryFrozenSnapshotIsReleased (263-267) walks frozenSnapshotPaths -> releasedSuites(), record->declaration only, so a record entry never written is invisible, as the finding states. (5) GeneratedSnapshotShapeTest.testEveryCandidateHasASnapshot reads the candidate DIRECTORY, and the stale committed file keeps both sides equal — it genuinely does not cover this; it does cover the reverse direction, and the "new candidate forgotten in Build" direction is caught by compilation (the candidate imports its generated file), leaving the drop direction as the uncovered one. (6) No CI regeneration-diff gate exists (rainix.yaml is the stock reusable; snapshot-generate-cmd runs only at tag release), and such a gate could not catch it anyway since the stale files are byte-identical. Escape is narrow — the group 2 source anchor fires if the dropped contract's source changes, so no wrong pin can ship — but a drop with no source change followed by a release cut permanently omits the contract from that tag's frozen record, leaves its released-suites lib without the tag, keeps RegistryDeployChainTest from ever being handed it, and SnapshotAlreadyFrozen makes the hole unrepairable. Severity MEDIUM is correct: no funds and no incorrect published pin, but a permanent gap in the release record this repo exists to publish plus lost chain verification for that release. Proposed fix is sound: checkedCandidateSuites() is inherited internal and reachable from a Build subclass harness, and test/script/Build.t.sol mirrors by subject path.

</details>

### LINKED ISSUE #57: The alias-lib emitters (writeAliasLib, aliasImportBlock, aliasLibraryBlock) have no test
**Audit finding `cov-04`** — dimension 2, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploySnapshot.sol:305-397`

## Problem

`grep -rn 'writeAliasLib\|aliasImportBlock\|aliasLibraryBlock' test/` returns nothing. These three functions emit `src/lib/LibAddressRegistryDeploy.sol` and `src/lib/LibMigrationRegistryDeploy.sol` — the consumer-facing pin files that `LibAddressRegistry.resolve` and every `LibMigrationRegistry` entry point read, and the import path the library's own docstring says "never moves, so consumers are unaffected by which snapshot it names".

Their siblings for the released lib (`releasedImportBlock`, `releasedLibraryBlock`, `writeReleasedSuitesLib`) have five dedicated tests including a regenerate-and-compare against the real committed file. The alias emitters have none of that: nothing pins the aliased names, nothing pins that the `dir` argument selects which snapshot is aliased (so an emitter that ignored `dir` and always wrote `candidate/` would be invisible), nothing pins the derived library name or the `src/lib/Lib<Contract>Deploy.sol` output path, and nothing regenerates the committed alias libs to prove they are not stale or hand-edited.

## Proposed fix

Add to `test/src/lib/LibRainDeploySnapshot.t.sol`:

```solidity
    /// The generated-file prefix both emitters share. Split so `reuse lint`
    /// reads this as a fixture rather than as this file's own declaration.
    /// @return The prefix.
    function expectedFilePrefix() internal pure returns (string memory) {
        return string.concat(
            "// SPDX-License",
            "-Identifier: LicenseRef-DCL-1.0\n",
            "// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd\n",
            "pragma solidity ^0.8.25;\n\n",
            "// THIS FILE IS AUTOGENERATED BY THE BUILD SCRIPT. DO NOT EDIT BY HAND.\n\n"
        );
    }

    /// The import block MUST alias BOTH pins out of the snapshot the `dir`
    /// names. The directory is what selects which snapshot a consumer's stable
    /// import path resolves to, so an emitter that ignored it would point every
    /// consumer at the rolling candidate whatever it was asked for.
    function testAliasImportBlockAliasesTheSnapshotItIsGiven() external pure {
        assertEq(
            LibRainDeploySnapshot.aliasImportBlock(
                EMITTED_CONTRACT, "ADDRESS_REGISTRY", LibRainDeploySnapshot.CANDIDATE
            ),
            "import {\n    DEPLOYED_ADDRESS as ADDRESS_REGISTRY_ADDR,\n    BYTECODE_HASH as ADDRESS_REGISTRY_HASH\n} from \"../generated/candidate/AddressRegistry.sol\";\n\n"
        );

        // A frozen release is aliased by the same rule, from its own tag.
        assertEq(
            LibRainDeploySnapshot.aliasImportBlock(EMITTED_CONTRACT, "ADDRESS_REGISTRY", "0_1_7"),
            "import {\n    DEPLOYED_ADDRESS as ADDRESS_REGISTRY_ADDR,\n    BYTECODE_HASH as ADDRESS_REGISTRY_HASH\n} from \"../generated/0_1_7/AddressRegistry.sol\";\n\n"
        );

        // The prefix is the caller's, so two contracts in one repo cannot
        // collide on the aliased names.
        assertTrue(
            vm.contains(
                LibRainDeploySnapshot.aliasImportBlock(
                    "MigrationRegistry", "MIGRATION_REGISTRY", LibRainDeploySnapshot.CANDIDATE
                ),
                "DEPLOYED_ADDRESS as MIGRATION_REGISTRY_ADDR"
            )
        );
    }

    /// The library block MUST re-export exactly the two pins, under the
    /// prefixed names consumers import, from the aliases the import block
    /// introduced.
    function testAliasLibraryBlockExportsBothPins() external pure {
        assertEq(
            LibRainDeploySnapshot.aliasLibraryBlock(
                EMITTED_CONTRACT, "ADDRESS_REGISTRY", "LibAddressRegistryDeploy"
            ),
            string.concat(
                "/// @title LibAddressRegistryDeploy\n",
                "/// @notice The deterministic Zoltu deploy address and code hash of\n/// `AddressRegistry`, aliased from its generated snapshot so that snapshot stays the\n",
                "/// single source of truth. The import path never moves, so consumers are\n",
                "/// unaffected by which snapshot it names.\n",
                "library LibAddressRegistryDeploy {\n",
                "    address constant ADDRESS_REGISTRY_DEPLOYED_ADDRESS = ADDRESS_REGISTRY_ADDR;\n",
                "    bytes32 constant ADDRESS_REGISTRY_DEPLOYED_CODEHASH = ADDRESS_REGISTRY_HASH;\n}\n"
            )
        );
    }

    /// The alias lib MUST land at `src/lib/Lib<Contract>Deploy.sol` holding
    /// exactly the prefix, import and library the emitters produce.
    ///
    /// Run against this repo's REAL committed file, so a stale or hand-edited
    /// alias lib is a test failure rather than a silent one. Restored BEFORE
    /// the assertions, because forge-std assertions revert.
    function testWriteAliasLibWritesTheCommittedLibAtItsPath() external {
        string memory path = "src/lib/LibAddressRegistryDeploy.sol";
        string memory before = vm.readFile(path);

        string memory written = LibRainDeploySnapshot.writeAliasLib(
            vm, EMITTED_CONTRACT, "ADDRESS_REGISTRY", LibRainDeploySnapshot.CANDIDATE
        );
        string memory emitted = vm.readFile(path);

        string memory expected = string.concat(
            expectedFilePrefix(),
            LibRainDeploySnapshot.aliasImportBlock(
                EMITTED_CONTRACT, "ADDRESS_REGISTRY", LibRainDeploySnapshot.CANDIDATE
            ),
            LibRainDeploySnapshot.aliasLibraryBlock(
                EMITTED_CONTRACT, "ADDRESS_REGISTRY", "LibAddressRegistryDeploy"
            )
        );

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.writeFile(path, before);

        assertEq(written, path);
        assertEq(emitted, expected);
        assertEq(emitted, before);
    }

    /// The second deployed contract's alias lib too — the library name and the
    /// output path are DERIVED from the contract name, so one contract cannot
    /// prove the derivation.
    function testWriteAliasLibDerivesThePathFromTheContract() external {
        string memory path = "src/lib/LibMigrationRegistryDeploy.sol";
        string memory before = vm.readFile(path);

        string memory written = LibRainDeploySnapshot.writeAliasLib(
            vm, "MigrationRegistry", "MIGRATION_REGISTRY", LibRainDeploySnapshot.CANDIDATE
        );
        string memory emitted = vm.readFile(path);

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.writeFile(path, before);

        assertEq(written, path);
        assertEq(emitted, before);
    }
```

---

<details>

Verified against source. The coverage claim is literally true: `grep -rn 'writeAliasLib|aliasImportBlock|aliasLibraryBlock' test/` returns nothing, the only caller is script/Build.sol:130, and the sibling released-lib emitters do have five dedicated tests in test/src/lib/LibRainDeploySnapshot.t.sol including a regenerate-and-byte-compare against the committed src/lib/LibAddressRegistryReleased.sol. Three exported functions of a library published for other deploy repos have zero direct coverage, and the `dir` parameter plus the library-name/output-path derivation are asserted nowhere. I could not refute that.

However the description's danger claims are largely already guarded, which is why the severity is wrong. src/abstract/RegistryDeploySuites.sol:115,143 builds each candidate DeploySuite with storedDeployedAddress/storedBytecodeHash taken FROM the alias libs while creationCode/storedRuntimeCode come from the candidate snapshot directly; RainDeployVerifySnapshot.checkInternallyConsistent (line 103) then requires storedDeployedAddress == zoltuAddress(creationCode) and storedBytecodeHash == derived hash, and testSnapshotMatchesSource anchors that creation code to type(AddressRegistry).creationCode. So a stale, hand-edited, or wrong-snapshot alias lib is already a CI failure by value. The emitted symbol names, library name and output path are pinned by compilation, since src/lib/LibAddressRegistry.sol and LibMigrationRegistry.sol import those exact symbols and LibAddressRegistry.t.sol / LibMigrationRegistry.t.sol:91 testDeployMatchesPins check the pins against a real Zoltu deploy of current source. `dir` is only ever passed CANDIDATE and src/generated/ holds only candidate/, so an emitter ignoring it is indistinguishable here and any other value would not compile.

Residual risk is therefore a test-coverage gap on a codegen helper whose consumer-visible output is already pinned elsewhere — no production value at risk that is not otherwise checked. LOW, not MEDIUM. The proposed fix is sound and applies cleanly (EMITTED_CONTRACT and CANDIDATE already exist in that test file).

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/45.
Closes https://github.com/rainlanguage/rain.deploy/issues/57 — the alias-lib emitter tests below (`testAliasImportBlockAliasesBothPins`, `testAliasLibraryBlockReExportsBothPins`, `testWriteAliasLibWritesTheLibAtItsPath`) are that issue in full.

Tests only. No `src/` or `script/` source change.

## What was actually uncovered

The issue carries two findings on `script/Build.sol`, and their two proposed
fixes contradict each other. `DEPLOY-3`'s verification says a test must
deliberately NOT invoke `run()`/`cutRelease()`; `cov-01`'s proposed fix is
built entirely around calling `run()`. This PR implements `DEPLOY-3` as
written and implements `cov-01` as its own verification block narrowed it,
which is not what its proposed-fix section says.

### `DEPLOY-3` — the two lists

`generatedContracts()` and `RegistryDeploySuites.candidateSuites()` are two
hard-coded arrays in two files. A candidate dropped from the generator by a
refactor or a merge still compiles, because its committed snapshot is still
there and still imported. `cutRelease()` then freezes only what the generator
names, `regenerateLibs()` writes a released-suites lib only for those, and the
release permanently omits that contract — `SnapshotAlreadyFrozen` makes the
hole unrepairable once the tag is cut.

Nothing downstream could see it. `testEveryFrozenSnapshotIsReleased` walks
record -> declaration, so a record entry never written is invisible to it.
`RegistryDeployChainTest` is never handed the omitted suite.
`testEveryCandidateHasASnapshot` compares the candidate DIRECTORY against the
declaration, and the stale committed file keeps both sides equal.

`test/script/Build.t.sol` asserts the two lists are the same SET, matched by
suite key in both directions, plus the three properties of the generator's own
entries that decide which file each contract's pins land in.

### `cov-01` — narrowed to the alias-lib emitter

`cov-01`'s verification corrected it MEDIUM -> LOW and refuted two of its three
consequences: `generatedContracts()` is not a second declaration of the
metadata, and the candidate snapshots and released libs are already anchored by
three test groups. What survived is one thing — `writeAliasLib`,
`aliasImportBlock` and `aliasLibraryBlock` are referenced by nothing outside
their own file, the only generator path here with zero execution by the suite,
so the committed `src/lib/Lib*Deploy.sol` was never proven to be what the
generator emits.

Closed in `test/src/lib/LibRainDeploySnapshot.t.sol` the way this repo already
closes exactly this shape: the two block emitters are pinned against literal
expected text, and `writeAliasLib` is driven against the REAL contract and
compared to the committed file, restored BEFORE the assertions because
forge-std assertions revert.

## Why nothing calls `run()` or `cutRelease()`

`run()` writes `src/lib/LibAddressRegistryReleased.sol`, which
`LibRainDeploySnapshotTest.testWriteReleasedSuitesLibWritesTheLibAtItsPath`
already reads, writes and restores. forge runs test contracts in parallel, so a
second writer of that file is a race — and one that yields flaky reds, not a
check. The alias test added here writes `Lib*Deploy.sol`, which no other test
writes and which every other reference to resolves as a compile-time symbol
import, so it adds no contended file.

`cov-01` also names `run()`'s two-step sequencing as unexercised. That
sequencing is not load-bearing: `regenerateLibs` reads nothing
`regenerateCandidates` writes. The alias libs are pure text from names, and the
released libs read only tag-shaped directories — `frozenSnapshotPaths` filters
on `isTag`, so `candidate/` falls out. The ordering that IS load-bearing is
`cutRelease()`'s, and `freeze` already enforces it structurally by taking
`regenerate` as a function pointer and running it first.

## Gates

- 175 pass / 47 fail, against a 168 / 47 baseline on `main`. All 47 are the
  pre-existing `vm.createSelectFork` failures from unset `*_RPC_URL`; neither
  the count nor the membership moved.
- `forge build` clean, `forge fmt --check` clean, `reuse lint` compliant,
  `pre-commit run --all-files` passing.

## QA

- Discriminating tests:
  `testGeneratedContractsAreExactlyTheDeclaredCandidates`,
  `testGeneratedContractNameMatchesTheArtifactPath`,
  `testGeneratedContractNamesAreUnique`,
  `testConstantPrefixesAreNonEmptyAndUnique`,
  `testAliasImportBlockAliasesBothPins`,
  `testAliasLibraryBlockReExportsBothPins`,
  `testWriteAliasLibWritesTheLibAtItsPath`.
  All seven are new, so all seven fail on base by not existing; each was
  additionally verified to fail on a break of the specific thing it guards,
  with the mutation applied to committed source and reverted after (table
  below). The last three in the alias set are mutually discriminating: the two
  pure block tests survive the `filePrefix()`, path-derivation and hand-edit
  mutations, and only the write test kills them.

- Mutations applied (line -> mutation -> killing test):
  - `script/Build.sol:83,87-91` -> `MigrationRegistry` entry deleted, length
    2 -> 1 -> `testGeneratedContractsAreExactlyTheDeclaredCandidates`
    (`a candidate is not generated, or a generated contract is not declared: 1 != 2`)
  - `script/Build.sol:87-91` -> entry 1 keeps `MigrationRegistry` but carries
    `addressRegistryCandidate()` ->
    `testGeneratedContractsAreExactlyTheDeclaredCandidates`
    (`declared candidate is not generated: migration-registry`)
  - `script/Build.sol:88` -> `contractName` -> `MigrationRegistryV2` ->
    `testGeneratedContractNameMatchesTheArtifactPath`
    (`generated name is not the artifact's contract`)
  - `script/Build.sol:85,88` -> both entries named `AddressRegistry` ->
    `testGeneratedContractNamesAreUnique`
    (`two generated entries name one contract`)
  - `script/Build.sol:89` -> `constantPrefix` emptied ->
    `testConstantPrefixesAreNonEmptyAndUnique` (`empty constant prefix`)
  - `script/Build.sol:85,89` -> both entries share `ADDRESS_REGISTRY` ->
    `testConstantPrefixesAreNonEmptyAndUnique`
    (`two entries share a constant prefix`)
  - `src/lib/LibRainDeploySnapshot.sol:313` -> `BYTECODE_HASH` alias dropped
    from `aliasImportBlock` -> `testAliasImportBlockAliasesBothPins` AND
    `testWriteAliasLibWritesTheLibAtItsPath`
  - `src/lib/LibRainDeploySnapshot.sol:346` -> `_DEPLOYED_CODEHASH` bound to
    `_ADDR` -> `testAliasLibraryBlockReExportsBothPins` AND
    `testWriteAliasLibWritesTheLibAtItsPath`
  - `src/lib/LibRainDeploySnapshot.sol:390` -> `LibCodeGen.filePrefix()`
    dropped from the concat -> ONLY
    `testWriteAliasLibWritesTheLibAtItsPath` (emitted file starts at
    `import {`)
  - `src/lib/LibRainDeploySnapshot.sol:384` -> path derived as
    `Lib<Contract>Deployed` -> ONLY
    `testWriteAliasLibWritesTheLibAtItsPath`
    (`LibAddressRegistryDeployed.sol != LibAddressRegistryDeploy.sol`)
  - `src/lib/LibAddressRegistryDeploy.sol` (the COMMITTED generated file)
    -> a doc line hand-edited -> ONLY
    `testWriteAliasLibWritesTheLibAtItsPath`, full-file string diff. This is
    the case `cov-01` exists for: a hand-edited generated file is now a test
    failure.

  Nothing survived its own mutation. Tree confirmed clean after every revert.

- Oracle: independent of the implementation in all three cases.
  - The list-agreement property's oracle is `checkedCandidateSuites()` — the
    deploy declaration, a different file from the generator's list, which is
    the whole point of the finding.
  - `contractName`'s oracle is the candidate's own `artifactPath`, split on
    `":"`, not the generator's own name field.
  - The alias emitters' oracle is literal expected text written out in the
    test, NOT reconstructed from the emitters under test. This matters: the
    pre-existing `testWriteReleasedSuitesLibWritesTheLibAtItsPath` builds its
    expectation by calling the same `releasedImportBlock`/`releasedLibraryBlock`
    the writer calls, so a mutation there moves both sides and survives. The
    alias tests avoid that, and `testWriteAliasLibWritesTheLibAtItsPath`'s
    oracle is the committed file on disk, which is fixed independently of any
    emitter change.

- Category check: the issue asks for (A) coverage of `script/Build.sol`, which
  had none, (B) an assertion that `generatedContracts()` covers every declared
  candidate, and (C) `cov-01`'s coverage gap. Covered A and B in
  `test/script/Build.t.sol`. Covered C as its own verification block narrowed
  it — the alias-lib emitter — and NOT as its proposed-fix section wrote it,
  because that section calls `run()`, which `DEPLOY-3`'s verification block in
  the same issue rules out as a parallel-write race. The `cutRelease()` write
  path is explicitly deferred by the issue itself to cov-02/cov-03, which need
  `freeze` to take a record root; it is out of scope here and stays uncovered.

### DIFF
--- a/test/concrete/BuildHarness.sol
+++ b/test/concrete/BuildHarness.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: LicenseRef-DCL-1.0
+// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+pragma solidity =0.8.25;
+
+import {Build, GeneratedContract} from "../../script/Build.sol";
+import {DeployCandidate} from "../../src/abstract/RainDeploySuitesBase.sol";
+
+/// @title BuildHarness
+/// @notice An external seam onto the two internal declarations `BuildTest`
+/// compares, so a test can hold both at once.
+///
+/// A harness rather than a change to `Build`: `generatedContracts()` is the
+/// script's own declaration and has no caller outside it, and widening it to
+/// `public` to be testable would put a second entry point on a script whose
+/// whole surface is `run()` and `cutRelease()`.
+contract BuildHarness is Build {
+    /// The generator's list.
+    /// @return The generated contracts.
+    function externalGeneratedContracts() external pure returns (GeneratedContract[] memory) {
+        return generatedContracts();
+    }
+
+    /// The deploy declaration's list, through the same guarded reader every
+    /// other consumer of the declaration uses.
+    /// @return The declared candidates.
+    function externalCandidateSuites() external pure returns (DeployCandidate[] memory) {
+        return checkedCandidateSuites();
+    }
+}
--- a/test/script/Build.t.sol
+++ b/test/script/Build.t.sol
@@ -3,31 +3,9 @@
 pragma solidity =0.8.25;
 
 import {Test} from "forge-std-1.16.1/src/Test.sol";
-import {Build, GeneratedContract} from "../../script/Build.sol";
+import {GeneratedContract} from "../../script/Build.sol";
 import {DeployCandidate} from "../../src/abstract/RainDeploySuitesBase.sol";
-
-/// @title BuildHarness
-/// @notice An external seam onto the two internal declarations, so a test can
-/// hold both at once and compare them.
-///
-/// A harness rather than a change to `Build`: `generatedContracts()` is the
-/// script's own declaration and has no caller outside it, and widening it to
-/// `public` to be testable would put a second entry point on a script whose
-/// whole surface is `run()` and `cutRelease()`.
-contract BuildHarness is Build {
-    /// The generator's list.
-    /// @return The generated contracts.
-    function externalGeneratedContracts() external pure returns (GeneratedContract[] memory) {
-        return generatedContracts();
-    }
-
-    /// The deploy declaration's list, through the same guarded reader every
-    /// other consumer of the declaration uses.
-    /// @return The declared candidates.
-    function externalCandidateSuites() external pure returns (DeployCandidate[] memory) {
-        return checkedCandidateSuites();
-    }
-}
+import {BuildHarness} from "../concrete/BuildHarness.sol";
 
 /// @title BuildTest
 /// @notice `script/Build.sol`'s own declaration.
--- a/test/script/Build.t.sol
+++ b/test/script/Build.t.sol
@@ -1,0 +1,168 @@
+// SPDX-License-Identifier: LicenseRef-DCL-1.0
+// SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+pragma solidity =0.8.25;
+
+import {Test} from "forge-std-1.16.1/src/Test.sol";
+import {Build, GeneratedContract} from "../../script/Build.sol";
+import {DeployCandidate} from "../../src/abstract/RainDeploySuitesBase.sol";
+
+/// @title BuildHarness
+/// @notice An external seam onto the two internal declarations, so a test can
+/// hold both at once and compare them.
+///
+/// A harness rather than a change to `Build`: `generatedContracts()` is the
+/// script's own declaration and has no caller outside it, and widening it to
+/// `public` to be testable would put a second entry point on a script whose
+/// whole surface is `run()` and `cutRelease()`.
+contract BuildHarness is Build {
+    /// The generator's list.
+    /// @return The generated contracts.
+    function externalGeneratedContracts() external pure returns (GeneratedContract[] memory) {
+        return generatedContracts();
+    }
+
+    /// The deploy declaration's list, through the same guarded reader every
+    /// other consumer of the declaration uses.
+    /// @return The declared candidates.
+    function externalCandidateSuites() external pure returns (DeployCandidate[] memory) {
+        return checkedCandidateSuites();
+    }
+}
+
+/// @title BuildTest
+/// @notice `script/Build.sol`'s own declaration.
+///
+/// `Build`'s NatSpec argues that "one list, three readers" makes a contract
+/// silently absent from a release impossible. That is true of the readers
+/// INSIDE `Build` — the regeneration, both lib writers and the freeze all read
+/// `generatedContracts()` — and false of the boundary: what this repo deploys
+/// is declared in `RegistryDeploySuites.candidateSuites()`, and the generator's
+/// list is a second hard-coded array in a second file.
+///
+/// A candidate dropped from `generatedContracts()` by a refactor or a merge
+/// still compiles, because its committed snapshot is still there and still
+/// imported. `cutRelease()` then freezes only the contracts the generator
+/// names, `regenerateLibs()` writes a released-suites lib only for those, and
+/// the release permanently omits that contract. Nothing downstream can see it:
+/// `testEveryFrozenSnapshotIsReleased` walks record -> declaration, so a record
+/// entry never written is invisible to it; `RegistryDeployChainTest` is never
+/// handed the omitted suite; and `testEveryCandidateHasASnapshot` compares the
+/// candidate DIRECTORY against the declaration, where the stale committed file
+/// keeps both sides equal. `SnapshotAlreadyFrozen` then makes the hole
+/// unrepairable, because the tag has been cut.
+///
+/// So the agreement of the two lists is asserted here, along with the three
+/// properties of the generator's own entries that decide which file each
+/// contract's pins are written into.
+///
+/// Deliberately nothing here calls `run()` or `cutRelease()`. Both write
+/// `src/lib/Lib*Released.sol`, which `LibRainDeploySnapshotTest` also writes,
+/// and forge runs test contracts in parallel — two contracts writing one file
+/// is a race, not a check. Everything below is pure.
+contract BuildTest is Test {
+    /// The harness the two declarations are read through.
+    BuildHarness internal sBuild;
+
+    function setUp() external {
+        sBuild = new BuildHarness();
+    }
+
+    /// PROPERTY: the generator's list and the deploy declaration are the SAME
+    /// SET of contracts, matched both ways. A candidate the generator does not
+    /// name is a contract silently absent from every release cut from here, and
+    /// neither the frozen-record check nor the chain check can see it, because
+    /// both are only ever handed what was written.
+    ///
+    /// Matched by the suite key rather than by index, because `Build` reaches
+    /// candidates by name for exactly this reason and a positional pass would
+    /// go green on a reordered list that had silently swapped two contracts.
+    function testGeneratedContractsAreExactlyTheDeclaredCandidates() external view {
+        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
+        DeployCandidate[] memory candidates = sBuild.externalCandidateSuites();
+
+        assertEq(
+            generated.length, candidates.length, "a candidate is not generated, or a generated contract is not declared"
+        );
+
+        for (uint256 i = 0; i < generated.length; i++) {
+            bool found = false;
+            for (uint256 j = 0; j < candidates.length; j++) {
+                if (
+                    keccak256(bytes(generated[i].candidate.snapshot.suite))
+                        == keccak256(bytes(candidates[j].snapshot.suite))
+                ) {
+                    assertEq(
+                        keccak256(generated[i].candidate.snapshot.creationCode),
+                        keccak256(candidates[j].snapshot.creationCode),
+                        "generated entry carries another contract's creation code"
+                    );
+                    found = true;
+                    break;
+                }
+            }
+            assertTrue(
+                found, string.concat("generated contract is not a declared candidate: ", generated[i].contractName)
+            );
+        }
+
+        for (uint256 j = 0; j < candidates.length; j++) {
+            bool found = false;
+            for (uint256 i = 0; i < generated.length; i++) {
+                if (
+                    keccak256(bytes(generated[i].candidate.snapshot.suite))
+                        == keccak256(bytes(candidates[j].snapshot.suite))
+                ) {
+                    found = true;
+                    break;
+                }
+            }
+            assertTrue(found, string.concat("declared candidate is not generated: ", candidates[j].snapshot.suite));
+        }
+    }
+
+    /// PROPERTY: `contractName` is the name the snapshot path and both
+    /// generated libs are built from, and it MUST be the contract the
+    /// candidate's artifact path names. A disagreement writes one contract's
+    /// pins into another contract's file, which every downstream check then
+    /// reads as self-consistent.
+    function testGeneratedContractNameMatchesTheArtifactPath() external view {
+        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
+        for (uint256 i = 0; i < generated.length; i++) {
+            string[] memory components = vm.split(generated[i].candidate.snapshot.artifactPath, ":");
+            assertEq(
+                components[components.length - 1],
+                generated[i].contractName,
+                "generated name is not the artifact's contract"
+            );
+        }
+    }
+
+    /// PROPERTY: no two entries name one contract. Two entries sharing a
+    /// `contractName` write one snapshot file twice and freeze one file for two
+    /// declarations, so the second silently replaces the first.
+    function testGeneratedContractNamesAreUnique() external view {
+        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
+        for (uint256 i = 0; i < generated.length; i++) {
+            for (uint256 j = i + 1; j < generated.length; j++) {
+                assertNotEq(
+                    generated[i].contractName, generated[j].contractName, "two generated entries name one contract"
+                );
+            }
+        }
+    }
+
+    /// PROPERTY: the constant prefix is non-empty and distinct per contract. It
+    /// names every constant both generated libs export, so a shared prefix is
+    /// two libs exporting colliding names.
+    function testConstantPrefixesAreNonEmptyAndUnique() external view {
+        GeneratedContract[] memory generated = sBuild.externalGeneratedContracts();
+        for (uint256 i = 0; i < generated.length; i++) {
+            assertGt(bytes(generated[i].constantPrefix).length, 0, "empty constant prefix");
+            for (uint256 j = i + 1; j < generated.length; j++) {
+                assertNotEq(
+                    generated[i].constantPrefix, generated[j].constantPrefix, "two entries share a constant prefix"
+                );
+            }
+        }
+    }
+}
--- a/test/src/lib/LibRainDeploySnapshot.t.sol
+++ b/test/src/lib/LibRainDeploySnapshot.t.sol
@@ -229,6 +229,103 @@ contract LibRainDeploySnapshotTest is Test {
         assertTrue(exists);
     }
 
+    /// The alias lib this repo's `AddressRegistry` snapshot is re-exported
+    /// through, and the only file the alias emitter writes here.
+    ///
+    /// `writeAliasLib` derives its path from the contract name and takes no
+    /// root, so driving it against a real contract writes the COMMITTED file.
+    /// That is what makes an alias lib whose text has drifted from what the
+    /// generator emits a failure rather than a diff nobody looks at — and the
+    /// alias emitter is otherwise the one generator path in this repo that
+    /// nothing anywhere executes, because only `script/Build.sol` calls it and
+    /// nothing may call that.
+    string constant ALIAS_LIB_PATH = "src/lib/LibAddressRegistryDeploy.sol";
+
+    /// The import block MUST alias exactly the two pins an alias lib
+    /// re-exports, out of the snapshot directory it is pointed at.
+    ///
+    /// Both constants are renamed on import because the lib exports its own
+    /// names for them in the same file, so a dropped alias is not a compile
+    /// error but a constant silently bound to whatever else is in scope. All
+    /// three parameters are asserted to reach the output distinctly: a block
+    /// that ignored `dir` would alias the rolling snapshot forever, including
+    /// where a frozen tag was asked for.
+    function testAliasImportBlockAliasesBothPins() external pure {
+        assertEq(
+            LibRainDeploySnapshot.aliasImportBlock(
+                "AddressRegistry", "ADDRESS_REGISTRY", LibRainDeploySnapshot.CANDIDATE
+            ),
+            "import {\n" "    DEPLOYED_ADDRESS as ADDRESS_REGISTRY_ADDR,\n"
+            "    BYTECODE_HASH as ADDRESS_REGISTRY_HASH\n" "} from \"../generated/candidate/AddressRegistry.sol\";\n\n"
+        );
+
+        assertEq(
+            LibRainDeploySnapshot.aliasImportBlock("MockDeployable", "MOCK_DEPLOYABLE", "0_1_7"),
+            "import {\n" "    DEPLOYED_ADDRESS as MOCK_DEPLOYABLE_ADDR,\n" "    BYTECODE_HASH as MOCK_DEPLOYABLE_HASH\n"
+            "} from \"../generated/0_1_7/MockDeployable.sol\";\n\n"
+        );
+    }
+
+    /// The library block MUST re-export BOTH pins under the consumer-facing
+    /// names, from the aliases the import block bound.
+    ///
+    /// The address and the code hash are what every consumer of this repo
+    /// resolves and verifies against, and they are the whole content of the
+    /// file. A block that emitted one of them, or bound one to the other's
+    /// alias, is a lib that compiles and lies.
+    function testAliasLibraryBlockReExportsBothPins() external pure {
+        assertEq(
+            LibRainDeploySnapshot.aliasLibraryBlock("AddressRegistry", "ADDRESS_REGISTRY", "LibAddressRegistryDeploy"),
+            "/// @title LibAddressRegistryDeploy\n"
+            "/// @notice The deterministic Zoltu deploy address and code hash of\n"
+            "/// `AddressRegistry`, aliased from its generated snapshot so that snapshot stays the\n"
+            "/// single source of truth. The import path never moves, so consumers are\n"
+            "/// unaffected by which snapshot it names.\n" "library LibAddressRegistryDeploy {\n"
+            "    address constant ADDRESS_REGISTRY_DEPLOYED_ADDRESS = ADDRESS_REGISTRY_ADDR;\n"
+            "    bytes32 constant ADDRESS_REGISTRY_DEPLOYED_CODEHASH = ADDRESS_REGISTRY_HASH;\n" "}\n"
+        );
+
+        assertEq(
+            LibRainDeploySnapshot.aliasLibraryBlock("MockDeployable", "MOCK_DEPLOYABLE", "LibMockDeployableDeploy"),
+            "/// @title LibMockDeployableDeploy\n"
+            "/// @notice The deterministic Zoltu deploy address and code hash of\n"
+            "/// `MockDeployable`, aliased from its generated snapshot so that snapshot stays the\n"
+            "/// single source of truth. The import path never moves, so consumers are\n"
+            "/// unaffected by which snapshot it names.\n" "library LibMockDeployableDeploy {\n"
+            "    address constant MOCK_DEPLOYABLE_DEPLOYED_ADDRESS = MOCK_DEPLOYABLE_ADDR;\n"
+            "    bytes32 constant MOCK_DEPLOYABLE_DEPLOYED_CODEHASH = MOCK_DEPLOYABLE_HASH;\n" "}\n"
+        );
+    }
+
+    /// The alias lib MUST land at the path derived from the contract name,
+    /// holding exactly the header, import block and library block the emitters
+    /// produce.
+    ///
+    /// Run against this repo's REAL contract and its real rolling snapshot, so
+    /// what it writes is the committed generated file and the assertion is that
+    /// the committed file IS what the generator emits today. Nothing else in
+    /// the suite can see that: the alias lib's VALUES are anchored by the pins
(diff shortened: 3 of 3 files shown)


thedavidmeister:
Reviewed 1c9d3f4: ready — Closes #45 and #57. The harness is the right call rather than the lazy one: `generatedContracts()` is the script's own declaration with no caller outside it, and widening it to `public` to make it testable would add a second entry point to a script whose entire surface is `run()` and `cutRelease()`. A test-only subclass keeps the production surface at two.

Reading the declaration side through `checkedCandidateSuites()` rather than around it means the comparison runs through the same guarded reader every other consumer uses, so the test cannot pass on a path nothing else takes.

PR #96, opened 2026-08-15
freeze: refuse a release tag that does not follow the record

### LINKED ISSUE #84: `freeze` has no release-tag monotonicity guard, so a mistyped `sol-v*` tag writes an out-of-order record the append-only rule then makes permanent
**Audit finding `HAZ6-04`** — dimension 6, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploySnapshot.sol:742-771`

## Problem

HAZARD: the release tag is a human-typed string that becomes an immutable, permanently-verified directory, and the only thing standing between a typo and that directory is whether a directory of that exact name already exists.

SHAPE: `freeze` runs three guards — `deployTag` refuses a non-`X.Y.Z` version, `SnapshotAlreadyFrozen` refuses a tag whose directory exists, `EmptyRelease` refuses a release naming no contracts. There is no guard that the tag is AHEAD of what is already frozen. `SnapshotAlreadyFrozen` is existence-based, so any version whose directory happens to be absent is freezable, including one below the newest release and one already published to Soldeer. The library already contains the exact ordering predicate this needs — `recordPrecedes` (445-465) compares tags numerically per component precisely because "`0_10_0` follows `0_9_0` as a release and precedes it as a string" — and `freeze` does not consult it. This is the org's monotonic-strictly-greater rule unenforced at the only place a release identity is minted.

PRODUCTION SCENARIO: main sits at release `0_2_0`. An operator cutting the next release pushes `sol-v0.1.9` — a fat finger, or a maintenance-branch habit. `rainix-tag-release`'s guard passes (the tag is on main), its regex passes (`0.1.9` is `MAJOR.MINOR.PATCH`), it seds `version = "0.1.9"` into `foundry.toml`, and `cutRelease()` finds no `src/generated/0_1_9/` and freezes one. The frozen bytes are the current candidate — byte-identical to `0_2_0`'s, since source has not moved — so the derived address is already live on all five chains and the pre-publish `forge test` chain gate PASSES. Soldeer accepts `rain-deploy~0.1.9` as a new version, and the workflow commits `src/generated/0_1_9/` back to main. Main now permanently carries a release that never happened, sorted BELOW the newest one; `releasedSuites()` declares `address-registry@0_1_9` and `migration-registry@0_1_9`; `RainDeployVerifyChain` will demand both stay live on every network Rain ever supports, forever; and `foundry.toml` claims the last released version is 0.1.9 when it is 0.2.0. Removing the directory is the only way out and it is exactly what the `frozen-snapshots-append-only` gate blocks, so the mistake is unwindable only by suspending the invariant the whole design rests on.

STRUCTURAL FIX: make the tag prove it follows the record, using the comparison the library already defines.

## Proposed fix

`src/lib/LibRainDeploySnapshot.sol`. Extract the tag comparison out of `recordPrecedes` so `freeze` can use it, then guard on it.

```diff
+/// Thrown when a release tag does not strictly follow every tag already in the
+/// record. The record is append-only, so a tag frozen out of order can never be
+/// removed: `releasedSuites()` declares it forever and the chain group demands
+/// it be live on every supported network forever. Refused at the one place a
+/// release identity is minted, rather than discovered afterwards.
+/// @param tag The tag being cut.
+/// @param newestFrozenTag The newest tag already in the record.
+error NonMonotonicRelease(string tag, string newestFrozenTag);
+
+    /// Whether release tag `a` strictly precedes release tag `b`, by version
+    /// component rather than by text — `0_10_0` follows `0_9_0` as a release
+    /// and precedes it as a string. Both MUST be `isTag`, so every component is
+    /// a run of digits `parseUint` reads.
+    /// @param vm The Vm instance for string operations.
+    /// @param a A release tag.
+    /// @param b A release tag.
+    /// @return Whether `a` strictly precedes `b`.
+    function tagPrecedes(Vm vm, string memory a, string memory b) internal pure returns (bool) {
+        string[] memory left = vm.split(a, "_");
+        string[] memory right = vm.split(b, "_");
+        for (uint256 i = 0; i < left.length; i++) {
+            uint256 leftComponent = vm.parseUint(left[i]);
+            uint256 rightComponent = vm.parseUint(right[i]);
+            if (leftComponent != rightComponent) {
+                return leftComponent < rightComponent;
+            }
+        }
+        return false;
+    }

     function recordPrecedes(Vm vm, string memory a, string memory b) internal pure returns (bool) {
-        string[] memory left = vm.split(tagForRecordPath(vm, a), "_");
-        string[] memory right = vm.split(tagForRecordPath(vm, b), "_");
-        for (uint256 i = 0; i < left.length; i++) {
-            uint256 leftComponent = vm.parseUint(left[i]);
-            uint256 rightComponent = vm.parseUint(right[i]);
-            if (leftComponent != rightComponent) {
-                return leftComponent < rightComponent;
-            }
-        }
+        string memory leftTag = tagForRecordPath(vm, a);
+        string memory rightTag = tagForRecordPath(vm, b);
+        if (tagPrecedes(vm, leftTag, rightTag)) {
+            return true;
+        }
+        if (tagPrecedes(vm, rightTag, leftTag)) {
+            return false;
+        }
         // tie break on the path, byte for byte, unchanged
```

Then in `freeze`, alongside the existing guards and BEFORE `regenerate()`:

```diff
         string memory tag = deployTag(vm);
         string memory frozenDir = dirForSnapshot(tag);
         if (vm.exists(frozenDir)) {
             revert SnapshotAlreadyFrozen(tag, frozenDir);
         }
         if (contractNames.length == 0) {
             revert EmptyRelease(tag);
         }
+        // The record is append-only, so an out-of-order tag is unremovable.
+        // Every existing tag must strictly precede this one.
+        string[] memory existing = sortedRecordPaths(vm, frozenSnapshotPaths(vm, LIB_FS_ROOT));
+        if (existing.length > 0) {
+            string memory newest = tagForRecordPath(vm, existing[existing.length - 1]);
+            if (!tagPrecedes(vm, newest, tag)) {
+                revert NonMonotonicRelease(tag, newest);
+            }
+        }
```

Tests in `test/src/lib/LibRainDeploySnapshot.t.sol`:

```solidity
    /// A tag BELOW the newest frozen release MUST be refused. The record is
    /// append-only, so freezing one is permanent.
    function testFreezeRefusesATagBelowTheNewestRelease() external {
        // record holds 0_2_0; foundry.toml version 0.1.9
        vm.expectRevert(abi.encodeWithSelector(NonMonotonicRelease.selector, "0_1_9", "0_2_0"));
        this.externalFreeze(contractNames);
        assertFalse(vm.exists(LibRainDeploySnapshot.dirForSnapshot("0_1_9")));
    }

    /// Refused BEFORE anything is written: a partial record is a frozen tag,
    /// which `SnapshotAlreadyFrozen` then refuses the retry of.
    function testFreezeLeavesNothingBehindOnANonMonotonicTag() external { /* as above, assert dir absent */ }

    /// The comparison MUST be by version component, not by text: `0_10_0`
    /// follows `0_9_0` and must be accepted, `0_9_0` after `0_10_0` refused.
    function testMonotonicityIsNumericNotLexicographic() external {
        assertTrue(LibRainDeploySnapshot.tagPrecedes(vm, "0_9_0", "0_10_0"));
        assertFalse(LibRainDeploySnapshot.tagPrecedes(vm, "0_10_0", "0_9_0"));
    }

    /// Equality is NOT strictly greater and MUST be refused — the fail-safe
    /// boundary. (`SnapshotAlreadyFrozen` also catches it; both must hold, so
    /// neither guard is load bearing alone.)
    function testFreezeRefusesTheNewestTagItself() external { /* expect SnapshotAlreadyFrozen */ }

    /// The FIRST release has nothing to follow and MUST be accepted.
    function testFreezeAcceptsTheFirstReleaseWithAnEmptyRecord() external { /* empty record, 0_0_1 accepted */ }

    /// A tag greater in a LATER component MUST be accepted (0_1_6 -> 0_1_7),
    /// and a major bump too (0_9_9 -> 1_0_0).
    function testFreezeAcceptsAnyStrictlyGreaterTag() external { /* both */ }

    /// A fuzzed pair MUST be accepted exactly when strictly greater.
    function testFreezeMonotonicityMatchesTagOrder(uint8 a, uint8 b, uint8 c) external { /* ... */ }
```

---

<details>

Could not refute. Verified against source and CI.

CODE READ CORRECTLY. `freeze` (src/lib/LibRainDeploySnapshot.sol:742-771) runs exactly the three guards the finding names — `deployTag`→`tagForVersion`'s strict `X.Y.Z` (149-161), `SnapshotAlreadyFrozen` on `vm.exists(frozenDir)` (745-747), `EmptyRelease` (748-750) — then `regenerate()`, then reads and writes. There is no ordering guard, and the doc block at 726-733 enumerates the guards without claiming ordering is deliberately unguarded, so this is not a documented org stance. `recordPrecedes` (445-465) does exist and does compare tags numerically per component for precisely the ordering reason the finding quotes, and `freeze` never calls it — the ordering predicate is present but only used by `sortedRecordPaths` (474-484) for emission order.

NOT GUARDED ELSEWHERE. Read the actual reusable workflow (`/home/gildlab/code/rainix-work/.github/workflows/rainix-tag-release.yaml`, which this repo pins via `.github/workflows/package-release.yaml` → `rainix-tag-release.yaml@main`). Its only tag checks are (a) `guard`: the tag commit is an ancestor of `origin/main`, and (b) `grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'` on the version stripped from the tag prefix. It then `sed -i`s that version into `foundry.toml` BEFORE running `snapshot-generate-cmd` (`forge script ./script/Build.sol --sig "cutRelease()"`), so the human-typed tag is literally the input `deployTag` reads. Nothing in the workflow compares the new version to the previous one. The `frozen-snapshots-append-only` action runs AFTER the commit and only forbids modifying existing snapshots, so a brand-new lower-numbered directory passes it — and is then what makes the mistake permanent, exactly as claimed.

THE CHAIN GATE DOES NOT CATCH IT. `RainDeployVerifyChain` (src/abstract/RainDeployVerifyChain.sol:134) checks `releasedSuites()` addresses are live. The bogus tag freezes the CURRENT candidate bytes, which are the bytes the manual deploy just put on chain (the workflow's own documented ordering: deploy first, then tag), so the pre-publish gate passes and the release lands on main and Soldeer. The generated `releasedSuites()` key is `<suite>@<tag>`, so no `DuplicateDeploySuite`, and `sortedRecordPaths` orders it numerically below the newer release — everything stays internally consistent, which is why nothing red-lines.

NOT TESTED. `grep "function test" test/src/lib/LibRainDeploySnapshot.t.sol` — 21 tests, the only `freeze` ones being `testFreezeRefusesAnEmptyRelease` and `testFreezeLeavesNothingBehindWhenThereIsNothingToFreeze`. No monotonicity test anywhere. There is also no version-vs-record identity test in this repo (`deployTag` has no caller in `src/abstract/`), so the regressed `[package].version` on main is neither prevented nor detected.

SEVERITY CORRECTED MEDIUM → LOW. Value at risk in production is bookkeeping, not consensus or funds: the phantom release's pins are correct and live on every chain, so a consumer that pins `rain-deploy~0.1.9` gets working, chain-verified contracts; nothing on chain is wrong, no key or money is touched, and the next legitimate release proceeds unimpeded (the existence-based guard has no quarrel with it). The residual harm is a permanently misordered record entry, one wasted Soldeer version, and a `[package].version` on main that names the wrong release until the next cut. Unwinding is an org-internal policy override of its own CI action, not a lock — costly and ugly, but not value at risk. Entry point is a single human typo in a rare, deliberate, manually dispatched operation.

FIX CAVEAT (does not affect the verdict). The proposed test suite is not implementable as written: `freeze` reads the real `foundry.toml` (`fs_permissions` grants it read-only, so a test cannot vary the version) and the proposed guard reads the real `LIB_FS_ROOT`, while `LibRainDeploySnapshotTest` drives fixture records under `test/generated` precisely because writing fixtures into `src/generated` would break contracts forge runs in parallel. A landed fix needs the record root (and ideally the tag) parameterised the way `frozenSnapshotPaths` and `writeReleasedSuitesLib` already are, or the guard will be as unexercisable as the finding says an unenforced rule is.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/84.

## The gap

`freeze` ran three guards — the version is a strict `X.Y.Z`, the tag is not already frozen, the release names at least one contract — and not one of them is about ORDER. So a `sol-v0.1.9` cut after `0.2.0` passes every one of them: its directory is absent precisely because that release never happened, and the bytes it freezes are the current candidate's, already live on every chain, so the pre-publish chain gate is green too.

The record is APPEND-ONLY, which is what turns that from wrong into permanent. `releasedSuites()` is generated from the record and declares the phantom release forever, `RainDeployVerifyChain` then demands its addresses stay live on every supported network forever, and the `frozen-snapshots-append-only` gate is exactly what stops the directory being deleted. The only exit is suspending the invariant the whole design rests on.

The ordering predicate was already in the library. `recordPrecedes` compares tags numerically per component precisely because `0_10_0` follows `0_9_0` as a release and precedes it as a string — `freeze` just never asked it.

## The change

`src/lib/LibRainDeploySnapshot.sol`:

- **`tagPrecedes(vm, a, b)`** — the numeric per-component comparison, split out of `recordPrecedes` behaviour-preservingly. `recordPrecedes` now asks `tagPrecedes` in both directions and falls through to its byte-for-byte path tie break only when neither tag precedes the other, i.e. the two files are the same release. One definition of "follows" answers the emission order and the guard alike; two spellings of it is how a record sorts one way and admits a tag the other way.
- **`newestFrozenTag(vm, recordRoot)`** — the greatest tag any file in a record sits under, compared as a version. `""` when nothing is frozen, which is a real state (every deploy repo before its first release, including this one) rather than a failure, and cannot collide with a release because it is not a tag `isTag` admits or a freeze could write. A scan, not the last of `sortedRecordPaths`: the path tie break that orders two contracts frozen under one tag has nothing to say about which tag is newest.
- **`checkReleaseFollowsRecord(vm, recordRoot, tag)`** — reverts `NonMonotonicRelease(tag, newestFrozenTag)` unless the tag strictly follows the newest release. Strictly, so the newest tag itself is refused: equality is not "follows". `SnapshotAlreadyFrozen` also refuses that one, and both must hold, so neither guard is load bearing alone.
- **`freeze`** calls it before `regenerate()`, alongside the existing guards.

## Design decisions

**The record root and the tag are parameters on the GUARD, not on `freeze`.** `freeze` passes `LIB_FS_ROOT` unconditionally. A caller-chosen read root on `freeze` would be a silent guard-disable with no backstop, because `freeze` writes into `LIB_FS_ROOT` regardless — the guard could be pointed away from the record it is protecting while the write still lands in it. Parameterising the guard is what makes the refusal reachable at all: the issue's own verification pass flagged that `freeze` reads the real `foundry.toml` (read-only under `fs_permissions`) and would read the real `LIB_FS_ROOT`, so a guard hung only off `freeze` would be as unexercisable as the unenforced rule it replaces. Same reason `frozenSnapshotPaths` takes a root and `tagForVersion` is split out of `deployTag`.

**The fuzzed guard test reads a committed `.txt` fixture.** forge 1.7.2-nightly runs the cases of a fuzz test CONCURRENTLY, so a test that builds and tears down its own filesystem fixture races itself — one case removes the record another case is reading. A record that is simply there is the only kind a fuzz can be put against. `.txt` is deliberate, not an oversight: a release directory holds every file in it and the record walk has no extension filter, so the walk finds this exactly as it finds a real snapshot, while the compiler never reads a fixture as source. The reasoning is in the fixture file itself.

Every non-fuzzed guard test still gets a record root of its own, for the reason `RELEASED_FIXTURE_ROOT` is not `FIXTURE_ROOT`: forge runs the tests in a contract concurrently, and a guard test sees the wrong newest release the moment another test's fixture is in the tree it is reading.

**`tagPrecedes` is a behaviour-preserving split.** No caller of `recordPrecedes` changes answer.

**README and CLAUDE.md are untouched.** Neither documents the freeze guard list.

## Tests

`test/src/lib/LibRainDeploySnapshot.t.sol`, all driven through an external wrapper so `vm.expectRevert` lands at the right call depth — the accepting cases go through it too, so both answers come from the same call.

- `testTagPrecedesComparesVersionsNotText` — `0_9_0` < `0_10_0` and not the reverse; every component, most significant difference deciding; and strictly, a tag does not precede itself.
- `testTagPrecedesMatchesComponentOrder` (fuzz) — over every pair of tags, the order is the order of the components read as numbers; exactly one of the two orderings holds unless they are the same release, when neither does.
- `testNewestFrozenTagIsTheGreatestReleaseNotAPositionInTheWalk` — `10_0_0` among `0_1_0`/`9_0_0` is neither the first nor the last by text, so an answer read off either end of the walk is wrong. The rolling candidate and a leftover scratch directory are in the tree and are not releases.
- `testNoFrozenReleaseAcceptsAnyFirstRelease` — both spellings of nothing released (root absent, root holding only the candidate) accept any first release. A guard that refused here would refuse the first release of every deploy repo that inherits it.
- `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease` — the production shape: `0_1_9` against a record holding `0_1_0` and `0_2_0`. Greater than a release that IS in the record, and refused against the NEWEST one.
- `testCheckReleaseFollowsRecordRefusesTheNewestTagItself` — the fail-safe boundary.
- `testCheckReleaseFollowsRecordAcceptsAnyStrictlyGreaterTag` — patch, major, and the `0_9_0` → `0_10_0` pair a text comparison gets wrong.
- `testCheckReleaseFollowsRecordAcceptsExactlyTheStrictlyGreater` (fuzz) — over every tag, against the committed record, accepts exactly the strictly greater and refuses every other one, naming both tags. Re-asserts the fixture's newest release on every case rather than assuming it: the whole answer is relative to the newest release, so a fixture that had quietly grown one would be a test that had quietly changed what it asks.

## QA

- **Discriminating tests:** `testTagPrecedesComparesVersionsNotText`, `testTagPrecedesMatchesComponentOrder` (fuzz), `testNewestFrozenTagIsTheGreatestReleaseNotAPositionInTheWalk`, `testNoFrozenReleaseAcceptsAnyFirstRelease`, `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease`, `testCheckReleaseFollowsRecordRefusesTheNewestTagItself`, `testCheckReleaseFollowsRecordAcceptsAnyStrictlyGreaterTag`, `testCheckReleaseFollowsRecordAcceptsExactlyTheStrictlyGreater` (fuzz) — each fails on base, in the strongest form: `tagPrecedes`, `newestFrozenTag` and `checkReleaseFollowsRecord` do not exist before this diff, so the suite does not compile against it. Per-behaviour discrimination is therefore the mutation matrix below, run against the landed code, which is the evidence that each test kills the specific line it claims.
- **Mutations applied** (`mutation-probe`, baseline green at 159 tests / 0 failed / 14 suites, 9 of 10 killed):

  | # | line → mutation | verdict / killing test |
  |---|---|---|
  | M01 | `LibRainDeploySnapshot.sol:468` `return leftComponent < rightComponent` → `>` | KILLED — `testTagPrecedesComparesVersionsNotText`, `testSortedRecordPathsOrdersTagsAsVersions`, `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease`, +2 |
  | M02 | `:464` `i < left.length` → `i < 1` (most significant component only) | KILLED — `testTagPrecedesComparesVersionsNotText`, `testCheckReleaseFollowsRecordAcceptsAnyStrictlyGreaterTag`, +3 |
  | M03 | `:471` `return false` → `return true` (a tag precedes itself) | KILLED — `testCheckReleaseFollowsRecordRefusesTheNewestTagItself`, `testTagPrecedesComparesVersionsNotText`, `testSortedRecordPathsOrdersTagsAsVersions` |
  | M04 | `:490-492` tag-order branch `return true` → `return false` (falls through to the path tie break) | KILLED — `testSortedRecordPathsOrdersTagsAsVersions`, `testRecordPathsForContractSelectsOneContractInTagOrder` |
  | M05 | `:493-495` reverse tag-order branch `return false` → `return true` | KILLED — `testSortedRecordPathsOrdersTagsAsVersions` |
  | M06 | `:774` drop `\|\| tagPrecedes(...)` (first release in the walk, not the greatest) | KILLED — `testNewestFrozenTagIsTheGreatestReleaseNotAPositionInTheWalk`, `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease` |
  | M07 | `:774` `tagPrecedes(vm, newest, tag)` → `tagPrecedes(vm, tag, newest)` (oldest release) | KILLED — `testNewestFrozenTagIsTheGreatestReleaseNotAPositionInTheWalk`, `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease` |
  | M08 | `:818` drop the `!` (accepts exactly the tags it should refuse) | KILLED — `testCheckReleaseFollowsRecordRefusesATagBelowTheNewestRelease`, `testCheckReleaseFollowsRecordRefusesTheNewestTagItself`, `testCheckReleaseFollowsRecordAcceptsAnyStrictlyGreaterTag` |
  | M09 | `:818` drop `bytes(newest).length > 0 &&` (refuses every first release) | KILLED — `testNoFrozenReleaseAcceptsAnyFirstRelease`, `testFreezeLeavesNothingBehindWhenThereIsNothingToFreeze` |
  | M10 | `:867` comment out `checkReleaseFollowsRecord(vm, LIB_FS_ROOT, tag)` in `freeze` | **SURVIVED** |

  **M10 survived and is not papered over.** It is `freeze`'s CALL to the guard, and it is structural, not a missing test. `freeze` takes its two inputs from the environment and neither can be moved by a test: the tag comes from `[package].version`, which `fs_permissions` grants `read` only, and the record root is `LIB_FS_ROOT`, which no test may put a release into because contracts forge runs in parallel walk that same root (the reason the fixtures live under `test/`, stated in `foundry.toml`). Both ends of that are checkable: `src/generated/` today holds only `candidate`, so `newestFrozenTag` returns `""` and the guard is a correct no-op — removing the call is unobservable. And once a release does exist, `[package].version` and the newest frozen tag move in lockstep, so `deployTag` equals the newest tag and `SnapshotAlreadyFrozen` (line 858) fires ahead of the guard (line 867) anyway. Killing M10 would take either a writable version input or a fixture release inside the real record, and the second is exactly the thing that breaks the parallel suites. The guard's own behaviour is fully covered by M06-M09; what is unproven by test is only that `freeze` is wired to it, which the diff shows directly.

  Three fork suites (`RainDeployVerifyChainTest`, `RainDeployVerifyChainCandidateTest`, `LibRainDeployTest`) were excluded from the probe run: they need `<NETWORK>_RPC_URL`, which CI's rpc-preflight binds and the probe host has not got. None of the three references `LibRainDeploySnapshot`, so none could have killed a mutant in it. CI runs them.
- **Oracle:** the ordering rule as #84 states it and ordinary semver precedence, derived in the test independently of the implementation. Both fuzz tests compute the expected answer from the raw `uint8` components by comparison (`aMajor != bMajor ? aMajor < bMajor : ...`), never by calling the code under test, so the assertion differs from the implementation's own answer whenever the implementation is wrong. The revert oracle is the exact `NonMonotonicRelease(tag, newest)` payload, both arguments asserted, not just the selector.
- **Category check:** #84 asks for (a) the tag comparison extracted so `freeze` can use it, (b) a guard refusing any tag not strictly following the record, (c) numeric-not-lexicographic ordering, (d) equality refused, (e) the first release accepted on an empty record, (f) strictly-greater accepted, (g) a fuzzed case. Covered: (a) `tagPrecedes`, (b) `checkReleaseFollowsRecord` called from `freeze`, (c) `testTagPrecedesComparesVersionsNotText` + `testTagPrecedesMatchesComponentOrder`, (d) `testCheckReleaseFollowsRecordRefusesTheNewestTagItself`, (e) `testNoFrozenReleaseAcceptsAnyFirstRelease`, (f) `testCheckReleaseFollowsRecordAcceptsAnyStrictlyGreaterTag`, (g) both fuzz tests. The issue spells its example tests against `freeze` itself; they are landed against the guard instead, which is what the issue's own verification section concludes is required — it flags that `freeze` reads the real read-only `foundry.toml` and the real `LIB_FS_ROOT`, so a guard reachable only through `freeze` "will be as unexercisable as the finding says an unenforced rule is". The issue's `testFreezeLeavesNothingBehindOnANonMonotonicTag` is the one example with no direct equivalent, for the same reason M10 survives; the property it wants is held structurally, by the guard sitting ahead of `regenerate()` and every write.

### DIFF
--- a/src/lib/LibRainDeploySnapshot.sol
+++ b/src/lib/LibRainDeploySnapshot.sol
@@ -35,6 +35,16 @@ error SnapshotAlreadyFrozen(string tag, string dir);
 /// @param tag The release tag that was being cut.
 error EmptyRelease(string tag);
 
+/// Thrown when a release tag does not strictly follow every tag already frozen
+/// in the record. The record is append-only, so a tag frozen out of order can
+/// never be removed: `releasedSuites()` declares it forever and the chain group
+/// then demands it be live on every supported network forever. Refused at the
+/// one place a release identity is minted, rather than discovered afterwards by
+/// whoever reads the record and finds a release nobody cut.
+/// @param tag The tag being cut.
+/// @param newestFrozenTag The newest tag already in the record.
+error NonMonotonicRelease(string tag, string newestFrozenTag);
+
 /// @title LibRainDeploySnapshot
 /// @notice Which release is being built, where its record lives, and how it is
 /// frozen. Release machinery, not code generation.
@@ -427,30 +437,61 @@ library LibRainDeploySnapshot {
         return string.concat(contractForRecordPath(vm, path), "_", tagForRecordPath(vm, path));
     }
 
-    /// Whether record file `a` is emitted before record file `b`: by release
-    /// tag, then by path.
+    /// Whether release tag `a` strictly precedes release tag `b`.
     ///
     /// Tags compare as VERSIONS rather than as text — `0_10_0` follows `0_9_0`
     /// as a release and precedes it as a string, so a text comparison misorders
-    /// every record that outlives a single-digit minor. Both tags are `isTag`,
-    /// so every component is a run of digits `parseUint` reads.
+    /// every record that outlives a single-digit minor. Both MUST be `isTag`,
+    /// so every component is a run of digits `parseUint` reads and both have
+    /// the same number of them.
     ///
-    /// The tie break is the path, byte for byte, so two contracts frozen under
-    /// one tag have an order at all. The walk's own order is the filesystem's,
-    /// and a generated file that changes with it is a diff on every build.
+    /// The ONE definition of what "follows" means for a release, asked by the
+    /// emission order and by the freeze's ordering guard alike. Two spellings
+    /// of it is how a record sorts one way and admits a tag the other way.
+    ///
+    /// Taking the tags rather than record paths is what makes the guard
+    /// reachable: a release being cut has no record file yet, and a comparison
+    /// that could only be asked with two paths could only be asked about
+    /// releases that already happened — the same reason `tagForVersion` is
+    /// split out of `deployTag`.
     /// @param vm The Vm instance for string operations.
-    /// @param a A record file.
-    /// @param b A record file.
-    /// @return Whether `a` precedes `b`.
-    function recordPrecedes(Vm vm, string memory a, string memory b) internal pure returns (bool) {
-        string[] memory left = vm.split(tagForRecordPath(vm, a), "_");
-        string[] memory right = vm.split(tagForRecordPath(vm, b), "_");
+    /// @param a A release tag.
+    /// @param b A release tag.
+    /// @return Whether `a` strictly precedes `b`.
+    function tagPrecedes(Vm vm, string memory a, string memory b) internal pure returns (bool) {
+        string[] memory left = vm.split(a, "_");
+        string[] memory right = vm.split(b, "_");
         for (uint256 i = 0; i < left.length; i++) {
             uint256 leftComponent = vm.parseUint(left[i]);
             uint256 rightComponent = vm.parseUint(right[i]);
             if (leftComponent != rightComponent) {
                 return leftComponent < rightComponent;
             }
+        }
+        return false;
+    }
+
+    /// Whether record file `a` is emitted before record file `b`: by release
+    /// tag (`tagPrecedes`), then by path.
+    ///
+    /// The tie break is the path, byte for byte, so two contracts frozen under
+    /// one tag have an order at all. The walk's own order is the filesystem's,
+    /// and a generated file that changes with it is a diff on every build. It
+    /// applies only when neither tag precedes the other, i.e. the two files are
+    /// the same release — a tag that follows the other is ordered by that and
+    /// never by the text of its path.
+    /// @param vm The Vm instance for string operations.
+    /// @param a A record file.
+    /// @param b A record file.
+    /// @return Whether `a` precedes `b`.
+    function recordPrecedes(Vm vm, string memory a, string memory b) internal pure returns (bool) {
+        string memory leftTag = tagForRecordPath(vm, a);
+        string memory rightTag = tagForRecordPath(vm, b);
+        if (tagPrecedes(vm, leftTag, rightTag)) {
+            return true;
+        }
+        if (tagPrecedes(vm, rightTag, leftTag)) {
+            return false;
         }
 
         bytes memory aBytes = bytes(a);
@@ -711,6 +752,74 @@ library LibRainDeploySnapshot {
         return path;
     }
 
+    /// The newest release in a record: the greatest tag any of its files sits
+    /// under, compared as a version.
+    ///
+    /// The empty string when the record holds no release at all. That is a real
+    /// state — a repo before its first release, including this one — rather
+    /// than a failure, and it cannot be confused with a release because it is
+    /// not a tag `isTag` admits or a freeze could ever write.
+    ///
+    /// A scan rather than the last of `sortedRecordPaths`: the newest release
+    /// is what this is asked for, and the path tie break that orders two files
+    /// frozen under one tag has nothing to say about which tag is newest.
+    /// @param vm The Vm instance for file operations.
+    /// @param recordRoot The record root — `LIB_FS_ROOT` for a repo's real
+    /// record.
+    /// @return newest The newest frozen tag, or `""` if nothing is frozen.
+    function newestFrozenTag(Vm vm, string memory recordRoot) internal view returns (string memory newest) {
+        string[] memory paths = frozenSnapshotPaths(vm, recordRoot);
+        for (uint256 i = 0; i < paths.length; i++) {
+            string memory tag = tagForRecordPath(vm, paths[i]);
+            if (bytes(newest).length == 0 || tagPrecedes(vm, newest, tag)) {
+                newest = tag;
+            }
+        }
+    }
+
+    /// Refuse a release tag that does not strictly follow every tag already in
+    /// the record.
+    ///
+    /// The record is APPEND-ONLY, which is what makes an out-of-order tag
+    /// permanent rather than merely wrong: `releasedSuites()` is generated from
+    /// the record and declares it forever, `RainDeployVerifyChain` then demands
+    /// its addresses stay live on every supported network forever, and the
+    /// append-only gate is exactly what stops the directory being deleted. The
+    /// only exit is suspending the invariant the whole design rests on.
+    ///
+    /// Nothing upstream is a substitute. The tag is a human-typed `sol-v*` that
+    /// `rainix-tag-release` seds straight into `foundry.toml` after checking
+    /// only that it is on `main` and shaped `X.Y.Z`, so a fat-fingered `0.1.9`
+    /// cut after `0.2.0` reaches this call as the release being made.
+    /// `SnapshotAlreadyFrozen` is existence-based and has no quarrel with it —
+    /// its directory is absent precisely because that release never happened —
+    /// and the frozen bytes are the current candidate, already live on chain,
+    /// so every downstream check passes and the record simply carries a release
+    /// sorted below the newest one.
+    ///
+    /// Strictly greater, so the newest tag itself is refused too: equality is
+    /// not "follows". `SnapshotAlreadyFrozen` also refuses that one, and both
+    /// must hold — neither guard is load bearing alone.
+    ///
+    /// The record root and the tag are parameters, so the refusal is reachable
+    /// without a record on disk to re-cut or a `foundry.toml` to rewrite, for
+    /// the same reason `frozenSnapshotPaths` takes a root and `tagForVersion`
+    /// is split out of `deployTag`. A guard nobody has seen fire is a guard
+    /// nobody knows works, which is the whole complaint against an unenforced
+    /// rule.
+    /// @param vm The Vm instance for file operations.
+    /// @param recordRoot The record root — `LIB_FS_ROOT` for a repo's real
+    /// record.
+    /// @param tag The release tag being cut. MUST be `isTag`.
+    function checkReleaseFollowsRecord(Vm vm, string memory recordRoot, string memory tag) internal view {
+        string memory newest = newestFrozenTag(vm, recordRoot);
+        // A repo that has released nothing has nothing for a first release to
+        // follow, so there is no tag it may not be.
+        if (bytes(newest).length > 0 && !tagPrecedes(vm, newest, tag)) {
+            revert NonMonotonicRelease(tag, newest);
+        }
+    }
+
     /// Regenerate the rolling snapshot and freeze it as this release's record,
     /// in that order, in one call.
     ///
@@ -730,6 +839,9 @@ library LibRainDeploySnapshot {
     /// - the release must name at least one contract, because a release with no
     ///   record is not a release and freezing one wedges the tag exactly as a
     ///   partial write does
+    /// - the tag must strictly follow every release already frozen
+    ///   (`checkReleaseFollowsRecord`), because the record it is appended to
+    ///   cannot be reordered or removed afterwards
     /// - every named contract must have a rolling snapshot, once regenerated
     ///
     /// The frozen copy is the bytes just regenerated, read back from disk, so
@@ -748,6 +860,11 @@ library LibRainDeploySnapshot {
         if (contractNames.length == 0) {
             revert EmptyRelease(tag);
         }
+        // The record this release is appended to is the one it is written into,
+        // so the root is `LIB_FS_ROOT` and not a caller's choice: a guard that
+        // could be pointed at another tree is a guard that could be pointed
+        // away from the record it is protecting.
+        checkReleaseFollowsRecord(vm, LIB_FS_ROOT, tag);
 
         regenerate();
 
--- a/test/fixture-record/1_1_1/MockDeployable.txt
+++ b/test/fixture-record/1_1_1/MockDeployable.txt
@@ -1,0 +1,13 @@
+SPDX-License-Identifier: LicenseRef-DCL-1.0
+SPDX-FileCopyrightText: Copyright (c) 2020 Rain Open Source Software Ltd
+
+A record fixture holding exactly one release, `1_1_1`, for the freeze's
+ordering guard to be fuzzed against. Committed rather than written by the test
+that reads it: forge runs the cases of a fuzz test CONCURRENTLY, so a fixture a
+fuzz builds and tears down races itself — one case removes the record another
+case is reading. A record that is simply there is the only kind a fuzz can be
+put against.
+
+Deliberately not `.sol`. A release directory holds every file in it and there is
+no extension to filter on, so the record walk finds this exactly as it finds a
+real snapshot, while the compiler never reads a fixture as source.
--- a/test/src/lib/LibRainDeploySnapshot.t.sol
+++ b/test/src/lib/LibRainDeploySnapshot.t.sol
@@ -8,6 +8,7 @@ import {DeploySuite} from "../../../src/abstract/RainDeploySuitesBase.sol";
 import {
     EmptyRelease,
     LibRainDeploySnapshot,
+    NonMonotonicRelease,
     NothingToFreeze,
     UnreleasableVersion
 } from "../../../src/lib/LibRainDeploySnapshot.sol";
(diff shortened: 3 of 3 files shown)


thedavidmeister:
@coderabbitai review

thedavidmeister:
Reviewed 2d8f41c: ready — Closes #84. `recordPrecedes` now compares TAGS rather than record paths, and that is what makes the new `NonMonotonicRelease` guard reachable at all: a release being cut has no record file yet, so a comparison that could only be asked with two paths could only be asked about releases that already happened. Same reasoning that splits `tagForVersion` out of `deployTag`.

Keeping one definition of what 'follows' means — asked by the emission order and by the freeze guard alike — is the part that matters. Two spellings is how a record sorts one way and admits a tag the other way.

Worth landing before the next tag specifically. The record is append-only and `freeze` refuses a re-cut with `SnapshotAlreadyFrozen`, so a tag frozen out of order can never be removed: `releasedSuites()` would declare it forever and the chain group would demand it live on every supported network forever. With #119 already landed, the two ways to mint a bad release identity — a padded component and an out-of-order tag — are both closed ahead of `sol-v0.1.6`.

PR #97, opened 2026-08-15
test: drive freeze's write path and its append-only guard

### LINKED ISSUE #46: `freeze()`'s write path and its `SnapshotAlreadyFrozen` append-only guard are untested
Two coverage findings with one blocker and one fix: `freeze` can only be pointed at the real `src/generated/` record, so neither its write path nor its append-only guard can be driven from a test.

---

# Finding `cov-03`

**Audit finding `cov-03`** — dimension 2, severity MEDIUM. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploySnapshot.sol:745-747`

## Problem

`grep -rn SnapshotAlreadyFrozen test/` returns nothing. The error is declared, documented as the thing that makes a release "cut once", and no test ever reaches it. It is the only protection on the immutability of `src/generated/<tag>/`, which is the record consumers pin their bytecode against: if it stopped firing, re-running `cutRelease()` would silently overwrite a published release's frozen snapshot with whatever the candidate currently holds.

Every other guard in this library has a test whose stated reason is that "a guard nobody has seen fire is a guard nobody knows works" — this one is the exception, and it is the guard with the most durable consequence. It is untestable today for the same reason as cov-02: `freeze` can only be pointed at `src/generated/`.

## Proposed fix

With the root seam from cov-02 in place, add to `test/src/lib/LibRainDeploySnapshot.t.sol`:

```solidity
    /// Where the re-cut fixture's record is built. Its own tree: this one is
    /// deliberately left frozen between the two calls, so no other test may
    /// share it.
    string constant RECUT_FIXTURE_ROOT = "test/generated-recut";

    /// External wrapper so `vm.expectRevert` lands at the right call depth.
    /// @param root The record root to freeze into.
    /// @param contractNames The contracts to freeze.
    function externalFreezeAt(string memory root, string[] memory contractNames) external {
        LibRainDeploySnapshot.freeze(vm, root, noRegeneration, contractNames);
    }

    /// A release is cut ONCE. Re-cutting a tag that already has a record MUST
    /// be refused, naming the tag and the directory, and MUST leave the
    /// original record exactly as it was — it is what consumers of that
    /// release pin against, and a second cut would replace it with whatever
    /// the candidate currently is.
    function testFreezeRefusesARecutRelease() external {
        string memory tag = LibRainDeploySnapshot.deployTag(vm);
        string memory frozenDir = string.concat(RECUT_FIXTURE_ROOT, "/", tag);
        string memory frozenPath = string.concat(frozenDir, "/", FREEZE_CONTRACT, ".sol");

        writeFixture(
            string.concat(RECUT_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol")
        );
        string[] memory contractNames = new string[](1);
        contractNames[0] = FREEZE_CONTRACT;

        LibRainDeploySnapshot.freeze(vm, RECUT_FIXTURE_ROOT, noRegeneration, contractNames);
        string memory firstCut = vm.readFile(frozenPath);

        // The candidate moves on, exactly as source does between releases.
        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.writeFile(
            string.concat(RECUT_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol"),
            freshRolling()
        );

        vm.expectRevert(abi.encodeWithSelector(SnapshotAlreadyFrozen.selector, tag, frozenDir));
        this.externalFreezeAt(RECUT_FIXTURE_ROOT, contractNames);

        // Read while the fixture is still there, asserted once it is gone.
        string memory afterRefusal = vm.readFile(frozenPath);
        string[] memory record = LibRainDeploySnapshot.frozenSnapshotPaths(vm, RECUT_FIXTURE_ROOT);

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.removeDir(RECUT_FIXTURE_ROOT, true);

        assertEq(afterRefusal, firstCut);
        assertNotEq(keccak256(bytes(afterRefusal)), keccak256(bytes(freshRolling())));
        assertEq(record.length, 1);
    }

    /// The refusal is about the DIRECTORY existing, not about what is in it, so
    /// an empty `<tag>/` left behind by anything refuses the real cut too —
    /// which is why `EmptyRelease` and the write ordering exist.
    function testFreezeRefusesATagDirectoryThatIsEmpty() external {
        string memory tag = LibRainDeploySnapshot.deployTag(vm);
        string memory frozenDir = string.concat(RECUT_FIXTURE_ROOT, "-empty/", tag);
        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.createDir(frozenDir, true);
        writeFixture(
            string.concat(
                RECUT_FIXTURE_ROOT, "-empty/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol"
            )
        );

        string[] memory contractNames = new string[](1);
        contractNames[0] = FREEZE_CONTRACT;

        vm.expectRevert(
            abi.encodeWithSelector(SnapshotAlreadyFrozen.selector, tag, frozenDir)
        );
        this.externalFreezeAt(string.concat(RECUT_FIXTURE_ROOT, "-empty"), contractNames);

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.removeDir(string.concat(RECUT_FIXTURE_ROOT, "-empty"), true);
    }
```

`SnapshotAlreadyFrozen` must be added to the existing import block at the top of the file.

---

<details>

Could not refute. Verified against source:

- /home/gildlab/code/rain-deploy-audit/src/lib/LibRainDeploySnapshot.sol:742-750 — `freeze` opens with `if (vm.exists(frozenDir)) revert SnapshotAlreadyFrozen(tag, frozenDir);` at exactly the cited lines 745-747. The error is declared at line 29 with the documented "a release is cut once" rationale, and the freeze NatSpec (line 721) names it as the guard that makes the record append-only.
- `grep -rn SnapshotAlreadyFrozen` over the whole repo hits only src/lib/LibRainDeploySnapshot.sol (declaration, two doc mentions, the revert). Nothing in test/, nothing in script/.
- The only two `freeze` call sites in tests are /home/gildlab/code/rain-deploy-audit/test/src/lib/LibRainDeploySnapshot.t.sol:648 (`externalFreeze(new string[](0))` → EmptyRelease) and :671 (`NoSuchContract` → NothingToFreeze). Both run with `src/generated/0_1_5/` absent (`ls src/generated` = `candidate` only; foundry.toml version = 0.1.5, nothing released yet), so the exists-check is false in both. Deleting the guard entirely leaves both tests green — the mutant survives, so the gap is real and not covered indirectly.
- The test contract's own header states the standard the finding invokes ("a guard nobody has seen fire is a guard nobody knows works"), and the other two freeze guards each have a dedicated test, so this is the stated-standard exception rather than an imported expectation.
- `freeze` genuinely has no root parameter (it derives paths via `dirForSnapshot`/`pathForSnapshot`, hardcoded to `LIB_FS_ROOT = "src/generated"`), unlike `frozenSnapshotPaths`/`writeReleasedSuitesLib`/`recordPathsForContract`, which all take a `recordRoot` precisely so they can be tested off the real tree — so the finding's account of why it is hard to test is accurate. (Strictly it is testable without the seam by creating `src/generated/0_1_5` and removing it, but that races the other test contracts forge runs in parallel against the real record — the documented reason FIXTURE_ROOT exists at :41-45 — so the seam is the right fix.)

Severity corrected MEDIUM → LOW. Value at risk is bounded: this is dev-time release tooling, not deployed contract code; no release has been cut yet, so no frozen record currently exists to overwrite; `cutRelease()` is a manual, once-per-release human action against a committed tree, so an overwrite shows up as a modified `src/generated/<tag>/` file in the release diff; and a record overwritten with a different candidate derives a different address, which `RainDeployVerifyChain` (group 4, released-only) red-lines because that address is not live on any supported network. The guard is worth a test, but its failure is neither silent nor unrecoverable in production.

</details>

---

# Finding `cov-02`

**Audit finding `cov-02`** — dimension 2, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploySnapshot.sol:742-771`

## Problem

`LibRainDeploySnapshotTest` drives only the two guards that fire BEFORE anything is written (`EmptyRelease`, `NothingToFreeze`), and both do so with `noRegeneration()`. Nothing anywhere calls `freeze` through to a successful cut, so none of this is exercised: the rolling snapshots are read back and copied into `<tag>/`, the copies are the bytes the regeneration just wrote, and the resulting directory is a release the record walk finds.

The ordering is the property that matters most and is the one entirely unasserted. The docstring states that `freeze` takes the regeneration as an argument and runs it FIRST precisely because freezing a stale candidate is silent — and with `noRegeneration()` as the only regeneration ever passed, a `freeze` that read the rolling files before calling `regenerate()` would pass every existing test. A stale freeze publishes a release whose recorded bytes are not the bytes that release deploys.

The write path is structurally undrivable today: `dirForSnapshot` hardcodes `src/generated/`, so a success-path test would create a real `<tag>/` directory that `RainDeployVerifySnapshot.testEveryFrozenSnapshotIsReleased` walks from contracts forge runs in parallel. That is why the gap exists, and the fix has to open the same seam `frozenSnapshotPaths` and `writeReleasedSuitesLib` already have.

## Proposed fix

1. Give `freeze` the record root the rest of the library already takes as a parameter, keeping the real caller unchanged:

```solidity
    // src/lib/LibRainDeploySnapshot.sol
    function freeze(Vm vm, function() internal regenerate, string[] memory contractNames) internal {
        freeze(vm, LIB_FS_ROOT, regenerate, contractNames);
    }

    /// @param root The record root — `LIB_FS_ROOT` for a repo's real record. A
    /// parameter for the same reason `frozenSnapshotPaths` takes one: a writer
    /// that can only be pointed at the real record can only be tested against
    /// it, and the real record is one a test must not leave a release in.
    function freeze(Vm vm, string memory root, function() internal regenerate, string[] memory contractNames)
        internal
    {
        string memory tag = deployTag(vm);
        string memory frozenDir = string.concat(root, "/", tag);
        if (vm.exists(frozenDir)) {
            revert SnapshotAlreadyFrozen(tag, frozenDir);
        }
        if (contractNames.length == 0) {
            revert EmptyRelease(tag);
        }

        regenerate();

        string[] memory records = new string[](contractNames.length);
        for (uint256 i = 0; i < contractNames.length; i++) {
            string memory rollingPath = string.concat(root, "/", CANDIDATE, "/", contractNames[i], ".sol");
            if (!vm.exists(rollingPath)) {
                revert NothingToFreeze(rollingPath);
            }
            records[i] = vm.readFile(rollingPath);
        }

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.createDir(frozenDir, true);
        for (uint256 i = 0; i < contractNames.length; i++) {
            //forge-lint: disable-next-line(unsafe-cheatcode)
            vm.writeFile(string.concat(frozenDir, "/", contractNames[i], ".sol"), records[i]);
        }
    }
```

2. Add to `test/src/lib/LibRainDeploySnapshot.t.sol`:

```solidity
    /// Where the freeze fixture's record is built. Its own tree, for the same
    /// reason `RELEASED_FIXTURE_ROOT` is not `FIXTURE_ROOT`, and NOT
    /// `src/generated`: a transient `<tag>/` there is a release the inherited
    /// record check has to fail on, from contracts forge runs in parallel.
    string constant FREEZE_FIXTURE_ROOT = "test/generated-freeze";

    /// The contract the freeze fixture cuts.
    string constant FREEZE_CONTRACT = "Marked";

    /// What the regeneration writes. Split so `reuse lint` reads this as a
    /// fixture rather than as this file's own license declaration.
    function freshRolling() internal pure returns (string memory) {
        return string.concat(
            "// SPDX-License", "-Identifier: LicenseRef-DCL-1.0\n",
            "address constant DEPLOYED_ADDRESS = address(0xfresh0000000000000000000000000000000fee);\n"
        );
    }

    /// What is on disk before the call, and what a freeze that read before
    /// regenerating would copy.
    function staleRolling() internal pure returns (string memory) {
        return string.concat(
            "// SPDX-License", "-Identifier: LicenseRef-DCL-1.0\n",
            "address constant DEPLOYED_ADDRESS = address(0x5741e000000000000000000000000000000000ee);\n"
        );
    }

    /// A regeneration that really regenerates, so "the guard fired FIRST" is no
    /// longer the only thing a freeze test can observe.
    function regenerateFreezeFixture() internal {
        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.writeFile(
            string.concat(FREEZE_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol"),
            freshRolling()
        );
    }

    /// A freeze MUST copy the bytes the regeneration wrote, not the bytes that
    /// were on disk when it was called. Freezing a stale candidate is silent —
    /// the immutability check only fires on a re-cut, which is too late — so
    /// the ordering is the whole of what makes a release describe itself.
    function testFreezeCopiesTheRegeneratedRollingSnapshot() external {
        string memory tag = LibRainDeploySnapshot.deployTag(vm);
        writeFixture(
            string.concat(FREEZE_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol")
        );
        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.writeFile(
            string.concat(FREEZE_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol"),
            staleRolling()
        );

        string[] memory contractNames = new string[](1);
        contractNames[0] = FREEZE_CONTRACT;
        LibRainDeploySnapshot.freeze(vm, FREEZE_FIXTURE_ROOT, regenerateFreezeFixture, contractNames);

        string memory frozenPath = string.concat(FREEZE_FIXTURE_ROOT, "/", tag, "/", FREEZE_CONTRACT, ".sol");
        bool frozenExists = vm.exists(frozenPath);
        string memory frozen = frozenExists ? vm.readFile(frozenPath) : "";
        // The cut is a release the record walk finds, under the tag the version
        // maps to and nowhere else.
        string[] memory record = LibRainDeploySnapshot.frozenSnapshotPaths(vm, FREEZE_FIXTURE_ROOT);
        // And the rolling snapshot is the regenerated one, still in place.
        string memory rolling = vm.readFile(
            string.concat(FREEZE_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", FREEZE_CONTRACT, ".sol")
        );

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.removeDir(FREEZE_FIXTURE_ROOT, true);

        assertTrue(frozenExists);
        assertEq(frozen, freshRolling());
        assertNotEq(keccak256(bytes(frozen)), keccak256(bytes(staleRolling())));
        assertEq(rolling, freshRolling());
        assertEq(record.length, 1);
        assertEq(record[0], frozenPath);
    }

    /// A release naming SEVERAL contracts MUST freeze every one of them. A
    /// contract regenerated but absent from the record is a contract silently
    /// missing from the release, and a tag that never held it has nothing
    /// missing from it for anything downstream to notice.
    function testFreezeCutsEveryNamedContract() external {
        string memory tag = LibRainDeploySnapshot.deployTag(vm);
        string[] memory contractNames = new string[](2);
        contractNames[0] = FREEZE_CONTRACT;
        contractNames[1] = "MarkedSecond";
        for (uint256 i = 0; i < contractNames.length; i++) {
            writeFixture(
                string.concat(
                    FREEZE_FIXTURE_ROOT, "/", LibRainDeploySnapshot.CANDIDATE, "/", contractNames[i], ".sol"
                )
            );
        }

        LibRainDeploySnapshot.freeze(vm, FREEZE_FIXTURE_ROOT, noRegeneration, contractNames);

        string[] memory record = LibRainDeploySnapshot.frozenSnapshotPaths(vm, FREEZE_FIXTURE_ROOT);
        bool first = holdsPath(record, string.concat(FREEZE_FIXTURE_ROOT, "/", tag, "/", contractNames[0], ".sol"));
        bool second = holdsPath(record, string.concat(FREEZE_FIXTURE_ROOT, "/", tag, "/", contractNames[1], ".sol"));

        //forge-lint: disable-next-line(unsafe-cheatcode)
        vm.removeDir(FREEZE_FIXTURE_ROOT, true);

        assertTrue(first);
        assertTrue(second);
        assertEq(record.length, 2);
    }
```

---

<details>

Verified against source, not refuted.

The factual claims hold. `freeze` at src/lib/LibRainDeploySnapshot.sol:742-771 is called from exactly two places: script/Build.sol:120 (`cutRelease()`, a tag/dispatch-only path nothing in `forge test` invokes) and test/src/lib/LibRainDeploySnapshot.t.sol:39 (`externalFreeze`, which hardcodes `noRegeneration`). Grepping the tree for `freeze` yields only two tests — `testFreezeRefusesAnEmptyRelease` (:644) and `testFreezeLeavesNothingBehindWhenThereIsNothingToFreeze` (:661) — and both revert before `vm.createDir(frozenDir, true)`. `SnapshotAlreadyFrozen` is not even imported by the test file. So the `createDir` + copy loop (lines 765-770) is executed by no test, and `src/generated/` holds only `candidate/`, so the write path has never run in this repo at all.

The ordering claim is the sharper half and also holds: moving `regenerate();` (line 752) below the read loop is a mutation that survives the whole suite, since the only regeneration any test passes is a no-op. The test file concedes this at :30-33 ("a no-op is what makes 'the guard fired' and 'the guard fired FIRST' the same observation"), while the library header and script/Build.sol:106-108 assert the ordering as load-bearing. An asserted property with a surviving mutant is a real dimension-2 gap, and nothing in CLAUDE.md documents this path as deliberately unexercised.

Severity corrected MEDIUM -> LOW, on value-at-risk:

1. The silent failure mode it leans on (stale freeze) is independently prevented on every push. `RegistryDeploySnapshotTest` inherits `testSnapshotMatchesSource` (RainDeployVerifySnapshot.sol:253) and `testSnapshotInternallyConsistent` (:224); the candidate suite's `creationCode`/`storedRuntimeCode` are imported from the COMMITTED src/generated/candidate/*.sol while `sourceCreationCode` is `type(X).creationCode` (RegistryDeploySuites.sol:110-120). In any CI-green tree the on-disk candidate is byte-equivalent to what `regenerate()` would write, so read-before-regenerate is unobservable. A stale freeze is reachable only by tagging a commit whose CI is already red.
2. A write-path defect fails loudly where it actually runs: `cutRelease()` regenerates the released-suites libs FROM the record immediately after (Build.sol:120-121), then `testEveryFrozenSnapshotIsReleased` and `RainDeployVerifyChain` check that record against the declaration and the live chains inside the same rainix-tag-release run (.github/workflows/package-release.yaml). The cost of the gap is a red release job, not a silently bad published record.
3. Nothing here is deployed bytecode or on-chain authority; it is build-time file writing.

One defect in the proposed fix the parent should not land as written: it parameterizes where `freeze` WRITES, citing `frozenSnapshotPaths` and `writeReleasedSuitesLib` as precedent — but both parameterize a READ root (`writeReleasedSuitesLib` still writes to a hardcoded src/lib/Lib<X>Released.sol). `writeSnapshot`'s docstring (LibRainDeploySnapshot.sol:252-259) states the opposite rule for writers: "There is no output root to choose... writing this repo's record somewhere else is not [a thing]". And the fix replaces `pathForSnapshot(CANDIDATE, ...)` with a hand `string.concat(root, "/", CANDIDATE, ...)`, exactly what `pathForSnapshot`'s docstring (:178-181) forbids: "Two spellings of one path is how a freeze silently reads nothing." It is equivalent today only because LibFs.pathForContract is `src/generated/<name>.sol`, so any future drift would be silent. The seam should be opened via root-aware `dirForSnapshot`/`pathForSnapshot` variants that preserve the single spelling, not by concatenating inside `freeze`.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/46 — audit findings `cov-02` and `cov-03`.

`freeze` could only ever be pointed at the real `src/generated/` record, so neither its write path nor its `SnapshotAlreadyFrozen` append-only guard could be driven from a test. `grep -rn SnapshotAlreadyFrozen test/` returned nothing: the only protection on the immutability of `src/generated/<tag>/` — the record consumers pin their bytecode against — had never been seen to fire.

## The seam

`freeze` now takes the record root the rest of the library already takes, threaded through new root-aware `dirForSnapshot` / `pathForSnapshot` variants.

Four choices worth stating, because each has a cheaper alternative that is wrong:

1. **The root is a REQUIRED parameter, not an overload with a defaulting wrapper.** The issue proposed the overload. A `freeze` that can still be called without a root can still be tested only against the real record, and the real record is one a test must not leave a release in. `frozenSnapshotPaths`, `recordPathsForContract` and `writeReleasedSuitesLib` do not default their root either.

2. **Paths go through the root-aware helpers, not hand-concatenation inside `freeze`.** The issue's proposed patch concatenated `string.concat(root, "/", CANDIDATE, ...)` directly — which the issue's *own* verification block flagged as a defect, against `pathForSnapshot`'s docstring: "Two spellings of one path is how a freeze silently reads nothing." The two spellings are equivalent only while `LibFs.pathForContract` stays `src/generated/<name>.sol`, so any later drift would be silent. `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot` fuzzes the root-aware spelling against `LibFs`'s at the real root, so there is one path and not two.

3. **Fixture snapshot content is comment-only.** `forge test` compiles everything under `test/`, and a fixture left behind by a *failing* test is source the next run has to compile. A realistic fixture — declarations, as a real snapshot holds — turns one red test into a repo that will not build until somebody deletes a directory by hand. Found by deleting the `SnapshotAlreadyFrozen` guard and watching the next run fail to *compile* rather than fail its assertions.

4. **`testFreezeCutsEveryNamedContract` asserts each frozen file holds its OWN contract's bytes.** Asserting only that a file per name exists is satisfied exactly by a write loop that stores `records[0]` for every `i`, and by a read loop truncated to the first contract. Mutants m6, m7 and m8 below are those three, and the weaker assertion killed none of them.

`writeSnapshot`'s docstring gave "writing this repo's record somewhere else is not [a thing]" as the reason no writer takes a root; it is reconciled here. `freeze` COPIES within one tree — reading a rolling snapshot under the root it is handed and writing the frozen copy under that same root — which is not the same freedom as choosing where a snapshot is GENERATED.

## Tests added

- `testFreezeCopiesTheRegeneratedRollingSnapshot` — the freeze copies the bytes the REGENERATION wrote, not the bytes on disk when it was called, and the copy lands where the record walk finds it.
- `testFreezeCutsEveryNamedContract` — a release naming several contracts freezes every one, each from its own rolling snapshot.
- `testFreezeRefusesARecutRelease` — a re-cut is refused naming the tag and the directory, and leaves the first cut byte for byte.
- `testFreezeRefusesATagDirectoryThatIsEmpty` — the refusal is about the directory existing, not its contents.
- `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot` — the root-aware path and `LibFs`'s are one path wherever both can spell it.

## Suite

| | tests | passed | failed |
|---|---|---|---|
| branch point `86f8d96` | 215 | 168 | 47 |
| this branch `9f5327e` | 220 | 173 | 47 |

The 47 are identical on both sides and every one is `vm.createSelectFork: environment variable *_RPC_URL not found` — 38 in `LibRainDeployTest`, 7 in `RainDeployVerifyChainTest`, 2 in `RainDeployVerifyChainCandidateTest`. They are absent fork endpoints in the local environment, not results. **Zero non-environmental failures on either side**, and the 5 tests this branch adds all pass.

## Mutation ledger

Nine mutants, one per behaviour the diff adds or changes. Run twice through independent harnesses — a per-mutant driver over the whole suite, and the org's `mutation-probe` over a suite narrowed to exclude the three fork contracts (their env failures make a baseline red, and the probe refuses to run on one). Both agree: **9 killed, 0 survived, 0 no-run, 0 harness errors.**

| # | mutation | verdict | killed by |
|---|---|---|---|
| m1 | `dirForSnapshot(root, dir)` ignores the root it was handed | KILLED | `testSnapshotPathsAgreeWithTheWriter`, and all five freeze tests |
| m2 | `pathForSnapshot(root, …)` drops the `.sol` extension | KILLED | `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot`, `testSnapshotPathsAgreeWithTheWriter`, `testFreezeLeavesNothingBehindWhenThereIsNothingToFreeze` |
| m3 | `pathForSnapshot(root, …)` ignores the root, so reader and writer drift | KILLED | `testSnapshotPathsAgreeWithTheWriter`, `testFreezeCopiesTheRegeneratedRollingSnapshot`, `testFreezeCutsEveryNamedContract`, `testFreezeRefusesARecutRelease` |
| m4 | the `SnapshotAlreadyFrozen` guard is deleted outright | KILLED | `testFreezeRefusesARecutRelease`, `testFreezeRefusesATagDirectoryThatIsEmpty` |
| m5 | `regenerate()` runs AFTER the reads, freezing a stale candidate | KILLED | `testFreezeCopiesTheRegeneratedRollingSnapshot` |
| m6 | every frozen file gets the FIRST contract's record | KILLED | `testFreezeCutsEveryNamedContract` |
| m7 | the read loop stops after the first contract | KILLED | `testFreezeCutsEveryNamedContract` |
| m8 | the write loop stops after the first contract | KILLED | `testFreezeCutsEveryNamedContract` |
| m9 | the cut is written back into the candidate, leaving the tag empty | KILLED | `testFreezeCopiesTheRegeneratedRollingSnapshot`, `testFreezeCutsEveryNamedContract`, `testFreezeRefusesARecutRelease` |

m4 and m5 are the two the issue was filed about. m4 — the append-only guard — had **no** test at all before this branch and survived the entire suite. m5 is the ordering: with `noRegeneration` as the only regeneration any test passed, a `freeze` that read before it regenerated passed every existing test.

### Evidence each suite actually ran

A killed mutant and a mutant that never compiled both make `forge test` exit non-zero, and this repo's suite is red at baseline from the missing RPC endpoints, so neither exit code nor "it went red" is evidence. Per mutant, `LibRainDeploySnapshotTest` **ran all 25 of its tests** (25 pass / 0 fail at baseline) and the compile logged `Compiler run successful`:

| # | LibRainDeploySnapshotTest | whole suite |
|---|---|---|
| m1 | 25 ran — 18 passed, 7 failed | 166 passed, 54 failed |
| m2 | 25 ran — 22 passed, 3 failed | 170 passed, 50 failed |
| m3 | 25 ran — 21 passed, 4 failed | 169 passed, 51 failed |
| m4 | 25 ran — 23 passed, 2 failed | 171 passed, 49 failed |
| m5 | 25 ran — 24 passed, 1 failed | 172 passed, 48 failed |
| m6 | 25 ran — 24 passed, 1 failed | 172 passed, 48 failed |
| m7 | 25 ran — 24 passed, 1 failed | 172 passed, 48 failed |
| m8 | 25 ran — 24 passed, 1 failed | 172 passed, 48 failed |
| m9 | 25 ran — 22 passed, 3 failed | 170 passed, 50 failed |

Every whole-suite figure is `47 environmental + N real`, and every `N` is accounted for by the named tests above.

Two harness traps were hit and fixed rather than reported through:

- **Fixture residue faking kills.** A mutant that breaks the freeze leaves its fixture record on disk, because the tests only remove fixtures once they get far enough to do so. A leftover `test/generated-recut/<tag>/` makes the NEXT mutant revert `SnapshotAlreadyFrozen` for a reason that has nothing to do with that mutant. Clearing residue is part of the suite command, before each run.
- **A mutant that ignores the record root writes into the REAL record**, both an untracked cut and untracked rolling files beside the two real ones — so cleanup restores `src/generated` as well.

## QA

- **Discriminating tests:** `testFreezeCopiesTheRegeneratedRollingSnapshot`, `testFreezeCutsEveryNamedContract`, `testFreezeRefusesARecutRelease`, `testFreezeRefusesATagDirectoryThatIsEmpty`, `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot` — none can run against base at all: on `86f8d96` `freeze` takes no record root, so a success-path test could only cut into the real `src/generated/`, which `testEveryFrozenSnapshotIsReleased` fails on from contracts forge runs in parallel. That untestability IS finding `cov-02`, so "fails on base" is verified the only way it can be — by restoring each base behaviour as a mutation and confirming the test fails: m4 restores base's absent append-only guard (base had no test for it; it survived the whole suite), and m5 restores the read-before-regenerate ordering that base's `noRegeneration`-only tests could not observe. Both are killed below.
- **Mutations applied:** 9/9 killed, 0 survived, 0 no-run, 0 harness errors, cross-checked through two independent harnesses.
  - `dirForSnapshot` `string.concat(root, …)` -> `string.concat(LIB_FS_ROOT, …)` -> `testSnapshotPathsAgreeWithTheWriter` + all five freeze tests
  - `pathForSnapshot` drop `".sol"` -> `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot`, `testSnapshotPathsAgreeWithTheWriter`
  - `pathForSnapshot` `dirForSnapshot(root, dir)` -> `dirForSnapshot(LIB_FS_ROOT, dir)` -> `testSnapshotPathsAgreeWithTheWriter`, `testFreezeCopiesTheRegeneratedRollingSnapshot`, `testFreezeCutsEveryNamedContract`, `testFreezeRefusesARecutRelease`
  - `freeze` delete `if (vm.exists(frozenDir)) revert SnapshotAlreadyFrozen(...)` -> `testFreezeRefusesARecutRelease`, `testFreezeRefusesATagDirectoryThatIsEmpty`
  - `freeze` move `regenerate();` below the read loop -> `testFreezeCopiesTheRegeneratedRollingSnapshot`
  - `freeze` write `records[0]` for every `i` -> `testFreezeCutsEveryNamedContract`
  - `freeze` read loop `i < contractNames.length` -> `i < 1` -> `testFreezeCutsEveryNamedContract`
  - `freeze` write loop `i < contractNames.length` -> `i < 1` -> `testFreezeCutsEveryNamedContract`
  - `freeze` write to `pathForSnapshot(root, tag, …)` -> `pathForSnapshot(root, CANDIDATE, …)` -> `testFreezeCopiesTheRegeneratedRollingSnapshot`, `testFreezeCutsEveryNamedContract`, `testFreezeRefusesARecutRelease`
- **Oracle:** the properties are stated in the issue and in the library's own NatSpec independently of the code — the regeneration runs FIRST, a release is cut ONCE, a snapshot path has ONE spelling — and expected values are never read back from the code under test. Each frozen file is compared against the fixture bytes the *test itself* wrote (`rollingFor(contractName)`, distinct per contract), and the freshness assertion compares against `freshRolling()` while asserting `assertNotEq` against `staleRolling()`, so a `freeze` that copied the wrong source cannot satisfy it. `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot`'s oracle is the independent `LibFs` writer, fuzzed rather than pinned to a chosen name.
- **Category check:** the issue asks for two findings — `cov-02` (freeze's write path: the regeneration ordering, the copied bytes, the cut being a release the record walk finds, and a multi-contract release) and `cov-03` (the `SnapshotAlreadyFrozen` append-only guard, including the empty-`<tag>/` case). Covered: ordering `testFreezeCopiesTheRegeneratedRollingSnapshot`; multi-contract `testFreezeCutsEveryNamedContract`; re-cut refusal and the original record left intact `testFreezeRefusesARecutRelease`; empty tag directory `testFreezeRefusesATagDirectoryThatIsEmpty`. The issue's examples are not treated as exhaustive: the seam they both need is covered as a property in its own right by `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot`, and the read/write loop truncations (m7, m8) are covered though the issue names neither.

## Checks run locally

`forge fmt --check` clean; `pre-commit run --all-files` all green (deadnix, denofmt, nil, nixfmt, prettier-rainix, statix, taplo, yamlfmt).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release snapshot handling to consistently use the specified record location.
  * Prevented invalid release cuts, including recuts, missing tag directories, and incorrect snapshot ordering.

* **Enhancements**
  * Added support for generating and freezing snapshots at alternate record locations.
  * Improved handling of refreshed rolling snapshots and releases containing multiple contracts.

* **Tests**
  * Expanded coverage for custom snapshot locations, regeneration behavior, release validation, and multi-contract freezes.


thedavidmeister:
Reviewed 026ad155f7fa15738d422998e9fccd7e7e699b3c: ready — Closes #46.

Drives `freeze`'s write path and its append-only guard. 256 tests pass on the merge commit and `forge fmt --check` is clean.

Two conflicts with `main` had to be resolved by hand, and both are the shape where a clean auto-merge produces broken code:

**#96's guard read the wrong record.** It added `checkReleaseFollowsRecord(vm, LIB_FS_ROOT, tag)` inside `freeze` while `freeze` had no root parameter; this branch gives `freeze` a root. Git merged both happily and left the guard reading the real record while the append landed in the caller's tree. Fixed to read `root`, with `testFreezeChecksTheRecordItIsAppendingTo` added — and proven by mutation rather than argued: reverting the argument to `LIB_FS_ROOT` fails only that new test, and survived every pre-existing one.

**#113 deleted the helper this branch calls.** It moved the shared test helpers into `test/lib/LibStringSet.sol` and removed the local `holdsPath` definitions; this branch added a new call to `holdsPath` while that was in flight. One side deletes a definition, the other adds a call, neither touches the other's lines — so the merge was clean and the tree did not compile. Both branches were green on their own CI. Routed through `LibStringSet.holds`, which is where the other membership assertions in that file already point.

#114's invariant survives: exactly one occurrence of the literal `"src/generated"` in `LibRainDeploySnapshot.sol`, at the `LIB_FS_ROOT` definition.

All three rainix jobs green — including `static`, which this PR could not pass before #130 trimmed `CLAUDE.md` under the 4096-byte cap. **CodeRabbit actually reviewed this one** — `Review completed` rather than the rate-limit banner most of today's PRs carried — with 0 unresolved threads, so that zero is coverage.

PR #98, opened 2026-08-15
docs: name every RegistryDeploySuites inheritor, without a count

### LINKED ISSUE #55: RegistryDeploySuites's docstring enumerates three inheritors when five inherit it, and omits the two that use it as a generation template
**Audit finding `ABS-07`** — dimension 4, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RegistryDeploySuites.sol:24-28`

## Problem

```solidity
/// Three contracts inherit this and nothing else declares a suite:
///
/// - `script/Deploy.sol` broadcasts from it
/// - `RegistryDeploySnapshotTest` checks its records against its creation code
/// - `RegistryDeployChainTest` checks it against every chain
```

`grep -rn RegistryDeploySuites --include=*.sol .` returns five inheritors:

- `script/Deploy.sol:36` — `contract Deploy is RegistryDeploySuites, RainDeployBroadcast {}`
- `script/Build.sol:79` — `contract Build is Script, RegistryDeploySuites`
- `test/src/abstract/RegistryDeploySnapshot.t.sol:24`
- `test/src/abstract/RegistryDeployChain.t.sol:29`
- `test/src/lib/GeneratedSnapshotShape.t.sol:37` — `contract GeneratedSnapshotShapeTest is RegistryDeploySuites, Test`

The omission of `script/Build.sol` is the substantive one. `Build.sol` is not an incidental fifth reader: its own header (lines 65-68) says it inherits this declaration specifically so the suite key, artifact path and dependency list are spelled once — "a second copy of them here is a second copy that drifts" — and it is what generates `LibAddressRegistryReleased`/`LibMigrationRegistryReleased`, i.e. what produces the `releasedSuites()` this same file returns. A reader taking the docstring at its word concludes the generator restates the declaration rather than inheriting it, which inverts the one property the file is arguing for. CLAUDE.md also disagrees with the docstring, listing four ("inherited by `script/Deploy.sol`, `script/Build.sol` and both pins test contracts").

The count is also load-bearing prose rather than decoration: "Three contracts inherit this and nothing else declares a suite" is the claim that makes "the deploy script broadcasts one contract while the tests verify another is not a thing that can be true here". A count that is wrong is a claim nobody can check by reading it.

## Proposed fix

**`src/abstract/RegistryDeploySuites.sol`**, replacing lines 24-32:

```diff
-/// Three contracts inherit this and nothing else declares a suite:
+/// Five contracts inherit this and nothing else declares a suite:
 ///
 /// - `script/Deploy.sol` broadcasts from it
+/// - `script/Build.sol` generates the snapshots, the alias libs and the
+///   released-suites libs from it — the named candidates below are the
+///   templates it emits from, which is why it inherits the declaration rather
+///   than restating the key, the artifact path and the dependency list
 /// - `RegistryDeploySnapshotTest` checks its records against its creation code
 /// - `RegistryDeployChainTest` checks it against every chain
+/// - `GeneratedSnapshotShapeTest` checks the shape of the files that generation
+///   writes
 ///
 /// So "the deploy script broadcasts one contract while the tests verify
 /// another" is not a thing that can be true here. Not because something checks
 /// for it — because there is only one list, and all three read it.
```

with the closing line's "all three" corrected to "all five". Correct CLAUDE.md's matching sentence ("inherited by `script/Deploy.sol`, `script/Build.sol` and both pins test contracts") to name `GeneratedSnapshotShapeTest` as well, so the two descriptions of the same fact agree.

---

<details>

Confirmed against source. src/abstract/RegistryDeploySuites.sol:24-32 states "Three contracts inherit this and nothing else declares a suite" and closes "all three read it", but grep -rn RegistryDeploySuites --include=*.sol . returns five inheritors: script/Deploy.sol:36, script/Build.sol:79, test/src/abstract/RegistryDeploySnapshot.t.sol:24, test/src/abstract/RegistryDeployChain.t.sol:29, test/src/lib/GeneratedSnapshotShape.t.sol:37. The file contradicts its own header lower down: the header itself says the released libs "come from script/Build.sol", and addressRegistryCandidate's docstring says "Build.sol inherits this rather than restating them" — so the three-count is falsified twice inside the file it heads. script/Build.sol:65-68 confirms the omission is the substantive one ("which is why this inherits the declaration rather than restating it ... a second copy of them here is a second copy that drifts"), i.e. the docstring's omission inverts the very single-source property the file argues for. CLAUDE.md:300-302 independently disagrees, listing four inheritors and itself omitting GeneratedSnapshotShapeTest, so the two descriptions of the same fact already drift. No test, guard or convention enforces docstring inheritor counts (grep of test/ shows nothing asserting on this prose), and this is not a restatement of a documented org convention. Severity LOW stands: documentation-only with no production value at risk, but above INFO because it is a load-bearing, checkable claim that is false and self-contradicting within the same file. The proposed fix is correct in substance, though the enumerating-count style is itself the fragile part — the durable fix is to name the readers without a leading cardinal, and to fix CLAUDE.md:300-302 in the same pass.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/55 (audit finding `ABS-07`, dimension 4, severity LOW).

## The defect

`src/abstract/RegistryDeploySuites.sol` opened with "Three contracts inherit this and nothing else declares a suite" and closed "there is only one list, and all three read it". Five contracts inherit it:

- `script/Deploy.sol:36`
- `script/Build.sol:79`
- `test/src/abstract/RegistryDeploySnapshot.t.sol:24`
- `test/src/abstract/RegistryDeployChain.t.sol:29`
- `test/src/lib/GeneratedSnapshotShape.t.sol:37`

The omissions are the two that make the file's own argument. `script/Build.sol:65-68` states it inherits the declaration precisely so the suite key, artifact path and dependency list are spelled once — "a second copy of them here is a second copy that drifts" — and it is what emits the `releasedSuites()` this same file returns. A reader taking the docstring at its word concludes the generator restates the declaration, which inverts the single-source property the file exists to argue for. The header contradicted itself lower down too: it already said the released libs "come from `script/Build.sol`", and `addressRegistryCandidate`'s docstring already said "Build.sol inherits this rather than restating them".

## The fix

The issue's proposed diff corrected the cardinal to five. Per the ruling in the issue's verification block, the leading cardinal is itself the fragile part, so this drops it rather than re-pins it: a count is a claim a sixth inheritor falsifies silently, and nothing enforces it. The readers are named, the opening reads "These contracts inherit it, and nothing else declares a suite", and the close reads "they all read it".

`CLAUDE.md:300-302` described the same fact and disagreed with both the old docstring and the code, listing four inheritors and omitting `GeneratedSnapshotShapeTest`. It now names the same set, also without a count ("both pins test contracts" → "the pins test contracts", for the same reason).

## QA
- Discriminating tests: n/a — the diff is comment-only, in two files that emit no bytecode behaviour. There is no runtime behaviour to discriminate on, and no test in the repo asserts on docstring prose (grep of `test/` finds nothing reading these comments). Adding a test that pins the comment text would enshrine the enumeration this change exists to de-fragilise.
- Mutations applied: n/a — a comment-only diff has no executable lines to mutate. Compilation before and after produces identical bytecode.
- Oracle: `grep -rn RegistryDeploySuites --include=*.sol .` run against this branch — the compiler's own inheritance graph, derived independently of the issue text and of the prose being corrected. It returns exactly the five inheritors now named, and the docstring is checked against that grep rather than against the issue's proposed diff.
- Category check: the issue asks for (a) the docstring's inheritor list corrected, and (b) CLAUDE.md's matching sentence corrected so the two agree. Both covered. The issue's verification block additionally rules that the durable form names the readers without a leading cardinal — applied to the opening line, the closing "all three", and to CLAUDE.md's "both pins test contracts", so no count survives anywhere in either description.

## Verification

- `forge fmt --check` — passes
- `forge build` — succeeds; the only output is warnings already present on `main` (`GeneratedSnapshotShape.t.sol:126` mutability, `LibRainDeploy.sol:304` unsafe-typecast)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -299,7 +299,7 @@ purpose: the suites and the broadcast are both inherited. Run only via the
 
 **`src/abstract/RegistryDeploySuites.sol`** — this repo's own declaration, one
 named candidate per deployed registry, inherited by `script/Deploy.sol`,
-`script/Build.sol` and both pins test contracts.
+`script/Build.sol`, the pins test contracts and `GeneratedSnapshotShapeTest`.
 
 **`src/abstract/RainDeployVerify*.sol`** — the deploy-pin verification every
 deploy repo inherits instead of hand-writing.
--- a/src/abstract/RegistryDeploySuites.sol
+++ b/src/abstract/RegistryDeploySuites.sol
@@ -21,15 +21,21 @@ import {LibMigrationRegistryReleased} from "../lib/LibMigrationRegistryReleased.
 /// @title RegistryDeploySuites
 /// @notice Everything this repo deploys, declared ONCE.
 ///
-/// Three contracts inherit this and nothing else declares a suite:
+/// These contracts inherit it, and nothing else declares a suite:
 ///
 /// - `script/Deploy.sol` broadcasts from it
+/// - `script/Build.sol` generates the snapshots, the alias libs and the
+///   released-suites libs from it — the named candidates below are the
+///   templates it emits from, which is why it inherits the declaration rather
+///   than restating the key, the artifact path and the dependency list
 /// - `RegistryDeploySnapshotTest` checks its records against its creation code
 /// - `RegistryDeployChainTest` checks it against every chain
+/// - `GeneratedSnapshotShapeTest` checks the shape of the files that generation
+///   writes
 ///
 /// So "the deploy script broadcasts one contract while the tests verify
 /// another" is not a thing that can be true here. Not because something checks
-/// for it — because there is only one list, and all three read it.
+/// for it — because there is only one list, and they all read it.
 ///
 /// It lives in `src/` rather than `test/` for two reasons. `.soldeerignore`
 /// excludes `test/`, and a downstream `script/` has to import its own


thedavidmeister:
Reviewed 7c19d4e: ready — Closes #55. Docs only, and the count goes with it rather than being patched: 'three contracts inherit this' becomes an unnumbered list, so the next inheritor does not silently make the sentence wrong again. `script/Build.sol` and `GeneratedSnapshotShapeTest` are both real inheritors that were missing, and the Build entry says why it inherits rather than restating the key, artifact path and dependency list — which is the invariant the whole docstring is about.

PR #102, opened 2026-08-15
fix: `candidateContractName` is `pure`, not `view`

### LINKED ISSUE #78: solc warning 2018: `candidateContractName` declared `view` when it can be `pure`
**Audit finding `Q4-03`** — dimension 4, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`test/src/lib/GeneratedSnapshotShape.t.sol:126`

## Problem

`nix develop -c forge build` reports:

```
   --> test/src/lib/GeneratedSnapshotShape.t.sol:126:5
```

The body only calls `vm.split` (declared `pure` on `Vm`) on a memory argument; `vm` itself is a `constant` on `Test`, so nothing is read from state. This is the only solc warning the whole tree emits, which means the build is one warning away from clean — and a build that already prints a warning is one where the next, real warning is not noticed.

## Proposed fix

`test/src/lib/GeneratedSnapshotShape.t.sol` line 126:
```diff
-    function candidateContractName(DeployCandidate memory candidate) internal view returns (string memory) {
+    function candidateContractName(DeployCandidate memory candidate) internal pure returns (string memory) {
```
No caller changes: `pure` is accepted everywhere `view` was.

---

<details>

Verified against source at 440e90b, could not refute.

1. The warning is real and at the cited location. `nix develop -c forge build --force` on a clean tree prints exactly:
   `Warning (2018): Function state mutability can be restricted to pure --> test/src/lib/GeneratedSnapshotShape.t.sol:126:5`
   Line 126 of /home/gildlab/code/rain-deploy-audit/test/src/lib/GeneratedSnapshotShape.t.sol is `function candidateContractName(DeployCandidate memory candidate) internal view returns (string memory) {`, matching the finding verbatim.

2. The mechanism is as described. The body calls only `vm.split(...)` on a `memory` argument; `split` is declared `external pure` at dependencies/forge-std-1.16.1/src/Vm.sol:1301, and `vm` is a constant on `Test`. Nothing is read from state.

3. The proposed fix was applied and built: with line 126 changed to `internal pure`, solc reports `Compiler run successful!` with no warnings. The sole caller is line 164 in the same file (`declared[i] = candidateContractName(candidates[i]);`), inside a `view` test function — `pure` is accepted there, no caller change needed. Grep confirms those are the only two occurrences of the symbol in the tree. Working tree restored afterwards.

4. No convention refutes it. CLAUDE.md documents no mutability exemption, and the file's sibling helpers already follow the compiler: `holdsName` and `artifactPath` are `pure`, while `snapshotContractNames`/`nodeTypes`/`constantDeclarations` are `view` because they genuinely call `vm.readDir`/`vm.readFile`. `candidateContractName` is the odd one out, not a deliberate pattern.

One factual overreach in the rationale, not enough to refute: the finding says fixing this leaves the build "clean", but `forge build` also emits a forge-lint warning (`warning[unsafe-typecast]` at src/lib/LibRainDeploy.sol:304). The narrow claim — it is the only *solc* warning — is true; the broader "one warning away from a clean build" is not, since lint output remains after the fix.

Severity corrected LOW -> INFO: this is a compiler style hint on an internal helper in test code. Nothing deployed, no test semantics, no on-chain value at risk; the mutability keyword affects no generated behaviour. The stated harm (a standing warning masks the next one) is also weakened by the lint warning that survives the fix, so the build is noisy either way until that is addressed too.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/78

## The diff

```diff
-    function candidateContractName(DeployCandidate memory candidate) internal view returns (string memory) {
+    function candidateContractName(DeployCandidate memory candidate) internal pure returns (string memory) {
```

One keyword in `test/src/lib/GeneratedSnapshotShape.t.sol`, exactly the fix the issue proposed.

## Why it is correct

The body calls only `vm.split`, declared `external pure` at `dependencies/forge-std-1.16.1/src/Vm.sol:1301`, against a `memory` argument, and `vm` is a `constant` on `Test`. Nothing is read from state, which is why solc can prove the restriction and emits warning 2018.

`pure` is accepted everywhere `view` was, so the sole caller — `testEveryCandidateHasASnapshot` at line 164 of the same file, itself `view` — needs no change. Those two lines are the only occurrences of the symbol in the tree.

It also puts the helper back in line with its siblings in the same file, which already track what they actually do: `holdsName` and `artifactPath` are `pure`, while `snapshotContractNames`/`nodeTypes`/`constantDeclarations` are `view` because they genuinely call `vm.readDir`/`vm.readFile`. `candidateContractName` was the odd one out, not a deliberate pattern.

## QA

- **Discriminating tests:** none added — `n/a`, and deliberately. State mutability on an `internal` function is a compile-time restriction with no runtime representation: there is no `STATICCALL`, no enforcement, nothing a test can observe. Proven rather than asserted — building both spellings and comparing the artifact for `GeneratedSnapshotShapeTest` gives `bytecode: IDENTICAL (len 39636)` and `deployedBytecode: IDENTICAL (len 39522)`. A forge test that failed on base and passed here cannot exist, so writing one would mean asserting on source text, which this file deliberately does not do (it reads the AST precisely so formatting cannot affect it).
- **Mutations applied:** `test/src/lib/GeneratedSnapshotShape.t.sol:126` `pure` -> `view` (i.e. back to base). **Killed by the compiler, not by a test.** `nix develop -c forge build --force` on the mutant prints `Compiler run successful with warnings:` / `Warning (2018): Function state mutability can be restricted to pure --> test/src/lib/GeneratedSnapshotShape.t.sol:126:5`; on this branch it prints `Compiler run successful!` with no solc warning anywhere in the tree. The mutant SURVIVES the test suite — `forge test --match-contract GeneratedSnapshotShapeTest` is 5 passed / 0 failed under both spellings, with gas identical to the unit (32114, 137663, 249440, 137661, 31717). That survival is the correct result, not a coverage gap: it is the same fact as the identical bytecode above. Behavioural coverage of the function is `testEveryCandidateHasASnapshot`, which is unaffected because the behaviour is unchanged.
- **Oracle:** solc 0.8.25's own mutability analysis, which is independent of this repo — it derived the restriction and named the line before any change was made. Backed by `Vm.sol:1301` declaring `split` as `external pure`, i.e. the reason the analysis is right rather than a restatement of its output.
- **Category check:** the issue asks for exactly one thing — the solc warning 2018 at `GeneratedSnapshotShape.t.sol:126`, with `view -> pure` as the proposed fix and "no caller changes" as the stated ramification. Covered: warning gone, caller verified unchanged, no other occurrence of the symbol. The issue's own verification block additionally observed that a forge-lint `warning[unsafe-typecast]` at `src/lib/LibRainDeploy.sol:304` survives this fix, so the build is not warning-free afterwards. I confirmed it still does. That is a different finding about a different file and is deliberately NOT bundled here; this PR closes the solc-warning finding as filed.

## Other verification

- Full `nix develop -c forge test`: 168 passed, 47 failed. Every one of the 47 is `vm.createSelectFork: environment variable <CHAIN>_RPC_URL not found` — the fork tests CLAUDE.md documents as needing a gitignored `.env`. Environmental, untouched by this diff; CI has the secrets.
- `nix develop -c forge fmt --check` passes; pre-commit hooks pass.

### DIFF
--- a/test/src/lib/GeneratedSnapshotShape.t.sol
+++ b/test/src/lib/GeneratedSnapshotShape.t.sol
@@ -123,7 +123,7 @@ contract GeneratedSnapshotShapeTest is RegistryDeploySuites, Test {
     /// declaration whose snapshot is somebody else's.
     /// @param candidate The candidate to name.
     /// @return The contract name.
-    function candidateContractName(DeployCandidate memory candidate) internal view returns (string memory) {
+    function candidateContractName(DeployCandidate memory candidate) internal pure returns (string memory) {
         string[] memory components = vm.split(candidate.snapshot.artifactPath, ":");
         return components[components.length - 1];
     }


thedavidmeister:
Reviewed 9c0b1e1: ready — Closes #78. `view` to `pure` on a helper whose only call is `vm.split`, which forge-std declares `pure` on `VmSafe` — an external call to a pure function is permitted from a pure context, and `rainix / static` and `rainix / test` both passing is the proof it compiles under this repo's own solc.

PR #103, opened 2026-08-15
Seed `suiteNames`'s named return before the loop

### LINKED ISSUE #72: `suiteNames`'s named return is assigned only inside a loop, made total by a guard in another function
**Audit finding `CQ4-02`** — dimension 4, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RainDeploySuitesBase.sol:198-203`

## Problem

`suiteNames()` declares the named return `names` and assigns it only inside `for (uint256 i = 0; i < suites.length; i++)`. Judged from this function's body alone there is a reachable path — `suites.length == 0` — on which `names` is never assigned and the function still returns, handing back the empty string. What actually makes that path unreachable is `allSuites()`, which reverts `NoDeployCandidates` on an empty candidate list; that totality lives in a different function and is not visible here. Per the org's named-return rule this is the "assigned only inside a loop made total by a guard elsewhere" case. It matters here rather than being academic: `names` is the valid-key list carried by `UnknownDeploymentSuite`, i.e. the only thing a mistyped `DEPLOYMENT_SUITE` tells a deployer, and if the guard in `allSuites` were ever loosened this reports "no valid suites" as an empty string rather than failing.

## Proposed fix

Make the assignment total in this function's own body by seeding `names` before the loop and iterating from 1.

Before (src/abstract/RainDeploySuitesBase.sol:198-203):
```solidity
    function suiteNames() internal pure returns (string memory names) {
        DeploySuite[] memory suites = allSuites();
        for (uint256 i = 0; i < suites.length; i++) {
            names = i == 0 ? suites[i].suite : string.concat(names, ", ", suites[i].suite);
        }
    }
```

After:
```solidity
    function suiteNames() internal pure returns (string memory names) {
        DeploySuite[] memory suites = allSuites();
        // Assigned on every path out of this function, including the empty set,
        // rather than relying on `allSuites` refusing one.
        names = "";
        if (suites.length == 0) {
            return names;
        }
        names = suites[0].suite;
        for (uint256 i = 1; i < suites.length; i++) {
            names = string.concat(names, ", ", suites[i].suite);
        }
    }
```

---

<details>

Verified against source at src/abstract/RainDeploySuitesBase.sol:198-203. The code reads exactly as quoted: `suiteNames()` declares named return `names` and assigns it only via the ternary inside `for (uint256 i = 0; i < suites.length; i++)`. Judged from that function's body alone there is a path (suites.length == 0) that reaches the closing brace unassigned and returns the empty string. Totality is supplied elsewhere, and even further away than the finding claims: `allSuites()` (line 172) delegates to `checkedCandidateSuites()` (line 156), which is where `revert NoDeployCandidates()` actually lives — two functions from the declaration.

Refutation attempts, all failed:
- "A guard elsewhere already handles it": the org rule (rainlanguage-audit-skills HEAD, skills/audit/SKILL.md:130) defines this case AS "assigned only inside a loop or a branch, where a guard elsewhere is what makes every path assign it — provable, but only via that guard", rates it LOW, and uses LibRainDeploy.deployToNetworks in this same repo as the worked example. The guard's existence is the premise of the finding, not a refutation.
- "A test covers it": test/src/abstract/RainDeploySuitesBase.t.sol:154-169 (testNoCandidateReverts) asserts externalSuiteNames() reverts NoDeployCandidates on an empty declaration. That pins the guard's behaviour, not local provability of the return — it is exactly the test that keeps passing if the guard is later narrowed so it no longer covers this reader.
- "Deliberate documented convention": CLAUDE.md and the file's own NatSpec document the single-guard design (one place raises NoDeployCandidates) but say nothing about return declarations. The rule's enumerated non-findings (assigned unconditionally on entry, total if/else, assembly-written, tuple-component documentation, file-local named-return convention) do not cover this shape.
- Case check: not last-iteration-wins (accumulator), not shadowed-by-explicit-return, not never-assigned-on-a-reachable-path (the path is unreachable in practice), so case 1, LOW.

Severity LOW is correct on production value-at-risk: the affected value is the validSuites field of an already-reverting UnknownDeploymentSuite, and the unassigned path is currently unreachable.

Caveat on the attached fix (does not affect the verdict): the proposed `if (suites.length == 0) { return names; }` is unreachable dead code that writes into the source the very thing checkedCandidateSuites' NatSpec and testNoCandidateReverts forbid — a reader answering from an empty declaration. The rule's own prescribed remedy is the minimal one: seed `names = "";` at the top of the function (or hoist the invariant), leaving the loop and its `i == 0` seed untouched, with no early return.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/72 (audit finding `CQ4-02`, dimension 4, severity LOW).

## The finding

`suiteNames()` declared the named return `names` and assigned it only inside `for (uint256 i = 0; i < suites.length; i++)`. Judged from this function's body alone, `suites.length == 0` reaches the closing brace with `names` never assigned and returns the empty string.

What actually makes that path unreachable is `checkedCandidateSuites()` reverting `NoDeployCandidates` — two functions away, reached via `allSuites()`. So the return is total, but only via a guard this reader cannot see. That is the org rule's case 1: *assigned only inside a loop, where a guard elsewhere is what makes every path assign it*.

It matters rather than being academic because `names` is the `validSuites` field of `UnknownDeploymentSuite` — the only thing a mistyped `DEPLOYMENT_SUITE` tells a deployer. If the guard in `checkedCandidateSuites` were ever narrowed, this reports "no valid suites" as an empty string instead of failing.

## The change

```solidity
 function suiteNames() internal pure returns (string memory names) {
     DeploySuite[] memory suites = allSuites();
+    // Seeded here so every path out of this function assigns `names`,
+    // provable from this body alone. `allSuites` never yields an empty set,
+    // but that is a fact about another function.
+    names = "";
     for (uint256 i = 0; i < suites.length; i++) {
         names = i == 0 ? suites[i].suite : string.concat(names, ", ", suites[i].suite);
     }
 }
```

Four lines, one of them code.

**The loop and its `i == 0` seed are deliberately untouched**, and mutant M4 below is the evidence they must be: with `names` seeded empty and the ternary dropped, the list comes back as `", address-registry-0-0-1, second-address, ..."`. The ternary is what keeps the first element unprefixed, so folding it into the seed is a behaviour change, not a simplification.

## Why not the fix proposed in the issue body

The issue's "After" block also added `if (suites.length == 0) { return names; }`. The issue's own collapsed verification block flags and rejects that, and I agree on reading the source:

- It is unreachable dead code — `allSuites()` cannot return an empty array.
- It writes into the source the exact thing `checkedCandidateSuites`' NatSpec forbids: *"a reader that answers from an empty declaration is a reader through which the whole registry can be empty and green."*
- `testNoCandidateReverts` asserts `externalSuiteNames()` **reverts** on an empty declaration. An early return is a second, contradictory spelling of that rule sitting in the source.

So this PR implements the minimal remedy the verification block prescribes — seed only, no early return.

## QA

- **Discriminating tests**: none added, and adding one is not possible here — stating that plainly rather than inventing a test. This is a local-provability fix, not a behaviour fix: `names = "";` is only observable on `suites.length == 0`, which `checkedCandidateSuites` makes unreachable by reverting `NoDeployCandidates`. A test that reached the line would first have to disable the guard the repo exists to enforce. The three existing tests that already pin this function are audited below rather than duplicated, per "an existing test that already kills mutants is left as-is": `testSuiteNamesIsTheRegistry`, `testEmptySuiteIsUnknown`, `testUnknownSuiteNamesEveryValidSuite` (all in `test/src/abstract/RainDeploySuitesBase.t.sol`), plus `testNoCandidateReverts` pinning that the empty path reverts rather than answers.
- **Mutations applied**: harness proven live first — unmutated baseline ran **10 tests, 10 passed, 0 failed**, with `testSuiteNamesIsTheRegistry`, `testNoCandidateReverts` and `testEmptySuiteIsUnknown` each named `[PASS]` in the output. Each mutant was greped back out of the file after `sed` to confirm it actually applied, so a no-op edit cannot fake a survival. All against `forge test --match-path test/src/abstract/RainDeploySuitesBase.t.sol`:
  - **M1** — delete `names = "";` (the line this PR adds) → **SURVIVED** (10 passed, 0 failed). Expected and reported as such rather than papered over: the line's whole purpose is to make an *unreachable* path assign, so no test can distinguish its presence. This is the known outcome for this finding class, not a coverage gap.
  - **M2** — `i == 0` → `i == 1` → **KILLED**, 3 failed. `testSuiteNamesIsTheRegistry` (`second-address, address-registry-candidate, second-address-candidate != address-registry-0-0-1, second-address, ...`), `testEmptySuiteIsUnknown`, `testUnknownSuiteNamesEveryValidSuite`. Run *with* the new seed present, which is what proves the seed did not weaken the ternary's coverage.
  - **M3** — separator `", "` → `","` → **KILLED**, 3 failed, same three tests.
  - **M4** — drop the ternary, always `string.concat(names, ", ", suites[i].suite)` → **KILLED**, 3 failed, same three tests, output `", address-registry-0-0-1, ..."` with the leading comma. This is the mutant that proves the `i == 0` branch cannot be collapsed into the seed.
- **Oracle**: the expected key list is independent of `suiteNames`' implementation — it is the registry order declared in the test fixture's own `releasedSuites()`/`candidateSuites()` (`address-registry-0-0-1, second-address, address-registry-candidate, second-address-candidate`), written out longhand as a literal in the test rather than recomputed by concatenating in the assertion. For the empty case the oracle is `checkedCandidateSuites`' NatSpec, which states that every reader must refuse an empty declaration — that is what rules out the early return the issue body proposed.
- **Category check**: the issue asks for exactly one thing — make the assignment total within this function's own body. Covered. The issue body's proposed diff additionally contained an early return; that is deliberately **not** implemented, because the issue's own verification block identifies it as unreachable dead code contradicting `checkedCandidateSuites`' NatSpec and `testNoCandidateReverts`, and prescribes the seed-only remedy instead. No adjacent scope taken: `allSuites`, `checkedCandidateSuites` and `suiteByName` are untouched.

## Full verification run

- `nix develop -c forge build` — clean, exit 0.
- `nix develop -c forge test` with all five RPC endpoints set: **215 passed, 0 failed, 0 skipped** across 17 suites.
- `nix develop -c forge fmt --check` — clean, exit 0.

(Without RPCs, 47 fork tests fail on `vm.createSelectFork: environment variable ..._RPC_URL not found`. That is the requirement documented in CLAUDE.md, not this change — they pass once the endpoints are set, which is the 215/215 above.)

### DIFF
--- a/src/abstract/RainDeploySuitesBase.sol
+++ b/src/abstract/RainDeploySuitesBase.sol
@@ -197,6 +197,10 @@ abstract contract RainDeploySuitesBase {
     /// @return names The declared keys.
     function suiteNames() internal pure returns (string memory names) {
         DeploySuite[] memory suites = allSuites();
+        // Seeded here so every path out of this function assigns `names`,
+        // provable from this body alone. `allSuites` never yields an empty set,
+        // but that is a fact about another function.
+        names = "";
         for (uint256 i = 0; i < suites.length; i++) {
             names = i == 0 ? suites[i].suite : string.concat(names, ", ", suites[i].suite);
         }


thedavidmeister:
Closing: the ruling is that this repo does not name return values, so the fix is to remove the name rather than to seed the default it silently supplies.

This PR is also its own best argument for that. Its mutation run reports that deleting the line it adds SURVIVED the whole suite — the line exists to make an unreachable path assign, so no test can distinguish its presence. The remedy for a named-return hole is unkillable by any test, because the compiler was the only mechanism that could ever have caught the class, and naming the return is what turned it off.

Superseded by the sweep issue, which also covers #72 and the generator that emits named returns into `Lib*Released.sol`.

PR #104, opened 2026-08-15
Correct checkCodeHash's stated justification in LibMigrationRegistry

### LINKED ISSUE #52: checkCodeHash's stated justification is factually wrong: an unguarded call into an empty account reverts, it does not decode as zero
**Audit finding `MIGREG-1`** — dimension 5, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibMigrationRegistry.sol:104-110, 131-135, 153-157`

## Problem

All three entry points justify `checkCodeHash()` with a mechanism that does not exist. `applied` (104-110) says that without the check, "no registry on this chain" would be silently collapsed into "this migration has not been applied" and "send a caller down its pre-migration branch on every chain the registry was never deployed to". `head` (131-135) says "a call into an empty account returns nothing, which decodes as zero ... an unverified read would hand back a head that is not a head at all". `applyMigration` (153-157) says a write would produce "a record that went nowhere" while "the migration would have run".

None of that is what solc 0.8.25 does. I compiled and ran the exact shapes against this repo's own compiler settings (solc 0.8.25, optimizer, 100000 runs, cancun): a high-level call to an address with no code REVERTS in both directions. For the returning call (`head`/`applied`), solc skips the extcodesize check because return data is expected and the ABI decoder then reverts on 0 bytes of returndata; for the void call (`applyMigration`), solc emits the extcodesize existence check and reverts. Both came back `ok=0, returndata length 0` — an anonymous revert, never a zero answer and never a silently-swallowed write.

So the empty-address case the comments lead with is already safe without the guard; what the guard actually buys there is a NAMED, diagnosable error instead of a bare revert. The case that genuinely would answer a lie — and that the comments never state as the reason — is the one the tests do cover: unrelated code, or an EIP-7702 delegated EOA, occupying the deterministic address and free to return any timestamp, any head, or a silent success for a write. This matters beyond tidiness in a repo whose comments are the spec: a consumer reading these lines learns a false model of what an unverified registry read does, and a maintainer weighing the guard's cost is told it prevents a silent wrong branch when it prevents an unnamed revert. The same wrong mechanism is restated in the suite at test/src/lib/LibMigrationRegistry.t.sol:264-269, 283-286 and 300-303, so the tests document it too. (The parallel claim in src/lib/LibAddressRegistry.sol is not made — that file's rationale is only "a loud revert rather than a call into unknown code", which is accurate.)

## Proposed fix

Rewrite the three rationales to state the mechanism that is real. src/lib/LibMigrationRegistry.sol lines 104-110, before:

    /// Verifies the registry's code hash before reading, so a chain where the
    /// registry is absent, or where something else occupies its address, is a
    /// loud revert rather than a call into unknown code. That distinction is
    /// the whole point here: "no registry on this chain" and "this migration
    /// has not been applied" are different facts, and silently collapsing the
    /// first into the second would send a caller down its pre-migration branch
    /// on every chain the registry was never deployed to.

after:

    /// Verifies the registry's code hash before reading, so a chain where the
    /// registry is absent, or where something else occupies its address, is a
    /// NAMED revert rather than a call into unknown code. An absent registry
    /// reverts either way — solc reverts a high-level call whose returndata is
    /// too short to decode — but anonymously, saying nothing about which of the
    /// two it was. What the check actually forbids is the case that does NOT
    /// revert: code at the address that is not this registry, an EIP-7702
    /// delegation included, is free to answer zero to every migration and send
    /// every caller down its pre-migration branch.

Lines 131-135, before:

    /// Verifies the registry's code hash first for the same reason `applied`
    /// does, and more sharply: a call into an empty account returns nothing,
    /// which decodes as zero, and zero is the one value a head can never hold —
    /// so an unverified read would hand back a head that is not a head at all,
    /// on exactly the chains where nothing has been deployed.

after:

    /// Verifies the registry's code hash first for the same reason `applied`
    /// does, and more sharply: occupying code is free to answer any head it
    /// likes, including the zero no head can ever hold, so an unverified read
    /// can hand back something that is not a head at all. An empty address is
    /// not that case — there is no returndata for a `bytes32` to decode from,
    /// so it reverts unguarded — and the check is what gives it a name.

Lines 153-157, before:

    /// Verifies the registry's code hash before writing, so a migration is
    /// never "applied" into an empty address or into unknown code. A record
    /// that went nowhere is worse than no record at all: the migration would
    /// have run, and every reader would go on asserting the pre-migration
    /// state.

after:

    /// Verifies the registry's code hash before writing, so a migration is
    /// never "applied" into unknown code. A write into an EMPTY address fails
    /// unguarded — solc checks the callee exists when no return data is
    /// expected — so what this stops is the write that SUCCEEDS into something
    /// that is not the registry: a record that went nowhere is worse than no
    /// record at all, because the migration ran and every reader goes on
    /// asserting the pre-migration state.

Apply the same correction to the three test rationales at test/src/lib/LibMigrationRegistry.t.sol:264-269, 283-286 and 300-303, whose assertions are unaffected — they already assert the named `UnexpectedMigrationRegistryCodeHash`, which is exactly the behaviour the corrected wording claims.

---

<details>

Could not refute. I read /home/gildlab/code/rain-deploy-audit/src/lib/LibMigrationRegistry.sol in full, the sibling /home/gildlab/code/rain-deploy-audit/src/lib/LibAddressRegistry.sol, the whole of /home/gildlab/code/rain-deploy-audit/test/src/lib/LibMigrationRegistry.t.sol, and CLAUDE.md, then verified the mechanism empirically rather than from memory.

EMPIRICAL CHECK. I compiled a probe under this repo's exact settings (foundry.toml: solc = "0.8.25", optimizer = true, optimizer_runs = 100000, evm_version = "cancun") calling a typed interface at an address with zero code, in all three shapes the library uses — `returns (uint256)` (as `applied`), `returns (bytes32)` (as `head`), and void (as `applyMigration`). Result: `readU ok: 0 / rdlen: 0`, `readB ok: 0 / rdlen: 0`, `write ok: 0 / rdlen: 0`. All three REVERT anonymously. Nothing decodes as zero and no write silently succeeds. This matches solc's documented codegen: since 0.8.10 the extcodesize existence check is skipped when return data is expected, and the ABI decoder reverts on 0 bytes of returndata; for the void call the existence check is emitted and reverts. (Probe removed; test/ is back to abstract, concrete, lib, src.)

WHAT THE CITED LINES SAY. The calls are high-level typed calls — `IMigrationRegistryV1(...).applied(...)` at line 124-125, `.head(...)` at 140, `.applyMigration(...)` at 169-170 — not low-level `call`, so the finding did not misread the call shape.

- 131-135 is unambiguously false as written: "a call into an empty account returns nothing, which decodes as zero ... so an unverified read would hand back a head that is not a head at all, on exactly the chains where nothing has been deployed." Empty account, `bytes32` return, unguarded: reverts. This is the strongest leg and it stands on its own.
- 104-110: "loud revert" is correct, but the justification clause — "silently collapsing the first into the second would send a caller down its pre-migration branch on every chain the registry was never deployed to" — pins the claim to the ABSENT case with "never deployed to", and in that case there is no collapse, there is a revert.
- 153-157: "so a migration is never 'applied' into an empty address ... A record that went nowhere ... the migration would have run" — the empty-address half is already true unguarded.

The finding is also right that the mechanism that DOES answer a lie (occupying code / an EIP-7702 designator free to return any timestamp, head, or a silent write success) is never stated as the reason, even though it is exactly what the suite covers: testAppliedWrongCode/testHeadWrongCode/testApplyMigrationWrongCode (lines ~318-365) and testAppliedDelegatedCode/testHeadDelegatedCode/testApplyMigrationDelegatedCode (~367-430).

REFUTATION ROUTES I TRIED AND THEY FAILED. (a) Tests already cover it — they do cover the BEHAVIOUR (all three NoRegistry tests at 264-315 assert the named UnexpectedMigrationRegistryCodeHash, which is correct and unaffected), but they restate the same wrong mechanism in their own doc comments at 264-269 ("That call would succeed and return nothing, which `abi.decode` would read as zero"), 283-286 and 300-303, so the suite propagates the error rather than refuting it. (b) The error's own natspec at 79-81 ("an empty account's code hash is zero, never the expected value") is accurate and is not part of the finding. (c) The sibling LibAddressRegistry.sol makes no such claim — its rationale is only "a loud revert rather than a call into unknown code", which is true — so this is not an org-wide documented convention being restated. (d) No CLAUDE.md convention excuses it; the repo explicitly treats prose as the spec.

SEVERITY. Leaving LOW rather than raising it. Runtime behaviour is correct and fully tested; nothing is at risk on chain. The error also errs conservatively — it OVERSTATES what the guard buys, so no maintainer reading it would remove the check. What is genuinely at risk is a consumer of this published package writing their own unguarded registry read on the false belief that it answers zero. That is real but small, and LOW is the right tier for it.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/52 (audit finding `MIGREG-1`, dimension 5, severity LOW).

## What was wrong

All three `LibMigrationRegistry` entry points justified `checkCodeHash()` with a mechanism solc does not exhibit. The comments claimed an unguarded call into an empty account returns nothing, decodes as zero, and so silently sends a caller down its pre-migration branch — or, for the write, silently swallows the record.

## Verified, not assumed

I did not take the issue's word for the mechanism. I compiled a probe under this repo's own settings (`solc = "0.8.25"`, optimizer, `optimizer_runs = 100000`, `evm_version = "cancun"`) driving the three call shapes the library uses as high-level typed interface calls, exactly as the library makes them:

| target | `applied`-shaped (`uint256`) | `head`-shaped (`bytes32`) | `applyMigration`-shaped (void) |
|---|---|---|---|
| empty address | `ok=false`, rdlen 0 | `ok=false`, rdlen 0 | `ok=false`, rdlen 0 |
| occupying code returning 32 zero bytes | `ok=true` -> `0x00..` | `ok=true` -> `0x00..` | `ok=true` |
| EIP-7702 designator (23-byte code) | `ok=true` -> `0x00..` | - | - |

An absent registry **reverts in all three directions**. Nothing decodes as zero and no write silently succeeds. For the returning calls solc skips the `extcodesize` check because return data is expected, and the ABI decoder reverts on 0 bytes of returndata; for the void call solc emits the existence check and reverts.

So the empty-address case the comments led with was already safe without the guard — what the guard buys there is a **named, diagnosable** error instead of a bare revert. The case that genuinely answers a lie, and that the comments never stated as the reason, is the one the suite already covers: unrelated code or an EIP-7702 delegation occupying the deterministic address, free to answer any timestamp, any head, or a silent success for a write.

The probe was removed before committing; it is not part of this diff.

## The change

Comments only, in two files:

- `src/lib/LibMigrationRegistry.sol` — the three rationales on `applied`, `head` and `applyMigration` now state the real mechanism: the absent registry reverts anonymously either way, and the check is what names it and what forbids the occupying-code case that does *not* revert.
- `test/src/lib/LibMigrationRegistry.t.sol` — the three `*NoRegistry` test rationales restated the same wrong mechanism, so they propagated it; corrected the same way, each now pointing at the `*WrongCode` / `*DelegatedCode` tests that cover the case which actually returns a wrong answer.

No assertion, no behaviour and no bytecode changes. Every changed line starts with `///`, so the deterministic address and codehash pins are untouched. The tests already assert the named `UnexpectedMigrationRegistryCodeHash`, which is exactly the behaviour the corrected wording claims.

The `checkCodeHash()` natspec itself and the sibling `LibAddressRegistry.sol` already stated an accurate rationale and are unchanged.

## QA

- Discriminating tests: `n/a` — the diff changes only `///` comment prose. There is no code path whose behaviour a new test could discriminate, and every existing assertion is byte-for-byte unchanged. What the change asserts is a claim ABOUT the compiler, and that claim was verified empirically (see Oracle) rather than by a suite test. Existing coverage of the corrected mechanism is already present and passing: `testAppliedWrongCode`, `testHeadWrongCode`, `testApplyMigrationWrongCode`, `testAppliedDelegatedCode`, `testHeadDelegatedCode`, `testApplyMigrationDelegatedCode` cover the occupying-code case the corrected prose names, and the three `*NoRegistry` tests cover the absent case.
- Mutations applied: `n/a` — a docs-only diff has no executable lines to mutate. Confirmed structurally: every added and removed line in `git diff` begins with `///`, and the compiled artifacts are unchanged, which the repo's own codehash/snapshot pin tests independently attest (the snapshot and registry-deploy suites pass, and those fail on any bytecode movement).
- Oracle: the solc compiler itself, not the implementation and not the issue text. I built a throwaway probe under this repo's exact `foundry.toml` settings that makes the same three high-level typed calls against (a) an address with no code, (b) an address etched with `0x60206000f3` (returns 32 zero bytes to any call), and (c) an EIP-7702 designator `0xef0100 || delegate`, and recorded `success` and returndata length for each. Results are the table above. The empty-address row falsifies the old prose; the occupying-code and delegated rows establish the new prose. This is independent of `LibMigrationRegistry` — the probe does not import it.
- Category check: issue asks for (A) the three source rationales at `src/lib/LibMigrationRegistry.sol:104-110, 131-135, 153-157`, and (B) the same correction applied to the three test rationales at `test/src/lib/LibMigrationRegistry.t.sol:264-269, 283-286, 300-303`. Covered A and B, both in full. The issue's exact proposed replacement text was used verbatim for A; for B the issue specified the correction without giving text, so the wording is mine and states the same verified mechanism. Deliberately NOT covered, as outside the filed scope: pinning the compiler behaviour as a permanent regression test so the prose cannot drift back, and the pre-existing "Both entry points check" in the `checkCodeHash()` natspec (there are three). Say the word and I will file or fold in either.

## Checks

- `forge fmt --check` — exit 0.
- `forge test` — 168 passed, 47 failed, and **all 47** failures are `vm.createSelectFork: environment variable *_RPC_URL not found`, the local `.env` absence CLAUDE.md documents. Zero non-RPC failures.
- `LibMigrationRegistryTest` — 23 passed, 0 failed.
- pre-commit hooks passed.

### DIFF
--- a/src/lib/LibMigrationRegistry.sol
+++ b/src/lib/LibMigrationRegistry.sol
@@ -103,11 +103,13 @@ library LibMigrationRegistry {
     ///
     /// Verifies the registry's code hash before reading, so a chain where the
     /// registry is absent, or where something else occupies its address, is a
-    /// loud revert rather than a call into unknown code. That distinction is
-    /// the whole point here: "no registry on this chain" and "this migration
-    /// has not been applied" are different facts, and silently collapsing the
-    /// first into the second would send a caller down its pre-migration branch
-    /// on every chain the registry was never deployed to.
+    /// NAMED revert rather than a call into unknown code. An absent registry
+    /// reverts either way — solc reverts a high-level call whose returndata is
+    /// too short to decode — but anonymously, saying nothing about which of the
+    /// two it was. What the check actually forbids is the case that does NOT
+    /// revert: code at the address that is not this registry, an EIP-7702
+    /// delegation included, is free to answer zero to every migration and send
+    /// every caller down its pre-migration branch.
     ///
     /// The registry itself refuses the zero writer, and refuses the two ids a
     /// migration can never be, so those arrive as reverts from it rather than
@@ -129,10 +131,11 @@ library LibMigrationRegistry {
     /// if it has never applied one.
     ///
     /// Verifies the registry's code hash first for the same reason `applied`
-    /// does, and more sharply: a call into an empty account returns nothing,
-    /// which decodes as zero, and zero is the one value a head can never hold —
-    /// so an unverified read would hand back a head that is not a head at all,
-    /// on exactly the chains where nothing has been deployed.
+    /// does, and more sharply: occupying code is free to answer any head it
+    /// likes, including the zero no head can ever hold, so an unverified read
+    /// can hand back something that is not a head at all. An empty address is
+    /// not that case — there is no returndata for a `bytes32` to decode from,
+    /// so it reverts unguarded — and the check is what gives it a name.
     /// @param writer The namespace to read. Never the zero address.
     /// @return The head of `writer`'s namespace. Never zero.
     function head(address writer) internal view returns (bytes32) {
@@ -151,10 +154,12 @@ library LibMigrationRegistry {
     /// one account interleave into one chain.
     ///
     /// Verifies the registry's code hash before writing, so a migration is
-    /// never "applied" into an empty address or into unknown code. A record
-    /// that went nowhere is worse than no record at all: the migration would
-    /// have run, and every reader would go on asserting the pre-migration
-    /// state.
+    /// never "applied" into unknown code. A write into an EMPTY address fails
+    /// unguarded — solc checks the callee exists when no return data is
+    /// expected — so what this stops is the write that SUCCEEDS into something
+    /// that is not the registry: a record that went nowhere is worse than no
+    /// record at all, because the migration ran and every reader goes on
+    /// asserting the pre-migration state.
     ///
     /// The registry refuses the zero id, refuses a migration this caller has
     /// already applied, and refuses one applied onto anything but the
--- a/test/src/lib/LibMigrationRegistry.t.sol
+++ b/test/src/lib/LibMigrationRegistry.t.sol
@@ -262,11 +262,13 @@ contract LibMigrationRegistryTest is Test {
     }
 
     /// A chain with no registry deployed reverts on the code hash rather than
-    /// calling into an empty account. That call would succeed and return
-    /// nothing, which `abi.decode` would read as zero — "this migration has
-    /// not been applied", on every chain the registry was never deployed to,
-    /// which is exactly the silent pre-migration branch this library exists to
-    /// make impossible.
+    /// calling into an empty account, and the point of that is the NAME. An
+    /// empty account has no returndata for `abi.decode` to read, so the call
+    /// reverts unguarded too — anonymously, saying nothing about whether the
+    /// registry is absent or the migration unapplied. Those are different
+    /// facts, and this error is what tells them apart. The case that would
+    /// answer the lie instead of reverting is occupying code, covered by
+    /// `testAppliedWrongCode` and `testAppliedDelegatedCode`.
     function testAppliedNoRegistry(address writer, bytes32 migration) external {
         assertEq(LibMigrationRegistryDeploy.MIGRATION_REGISTRY_DEPLOYED_ADDRESS.code.length, 0);
 
@@ -280,10 +282,11 @@ contract LibMigrationRegistryTest is Test {
         this.externalApplied(writer, migration);
     }
 
-    /// Reading a head off a chain with no registry is refused for a sharper
-    /// version of the same reason: the empty-account read decodes as zero, and
-    /// zero is a value no head can ever hold, so an unverified read hands back
-    /// something that is not a head at all.
+    /// Reading a head off a chain with no registry is refused for the same
+    /// reason and with the same named error: an unguarded read reverts here
+    /// anyway, because there is no returndata for a `bytes32` to decode from,
+    /// so what the check adds is which chain state it was. A head that is not
+    /// a head is what occupying code can return, not what an empty one does.
     function testHeadNoRegistry(address writer) external {
         assertEq(LibMigrationRegistryDeploy.MIGRATION_REGISTRY_DEPLOYED_ADDRESS.code.length, 0);
 
@@ -297,10 +300,12 @@ contract LibMigrationRegistryTest is Test {
         this.externalHead(writer);
     }
 
-    /// Writing to a chain with no registry is refused for the mirror reason: an
-    /// `applyMigration` into an empty account is a migration that reports itself
-    /// applied and is not, which leaves every reader asserting the
-    /// pre-migration state forever.
+    /// Writing to a chain with no registry is refused by the code hash rather
+    /// than by solc's own existence check, which an `applyMigration` into an
+    /// empty account hits regardless — no return data is expected, so the
+    /// callee is checked to exist and the write reverts unnamed. A migration
+    /// that reports itself applied and is not is what a write into occupying
+    /// code does, and `testApplyMigrationWrongCode` is where that is covered.
     function testApplyMigrationNoRegistry(bytes32 expectedHead, bytes32 migration) external {
         assertEq(LibMigrationRegistryDeploy.MIGRATION_REGISTRY_DEPLOYED_ADDRESS.code.length, 0);
 


thedavidmeister:
Reviewed 4a30b34: ready — Closes #52. Comments only, and the correction is a real sharpening rather than a reword. The old text implied the codehash check is what turns an absent registry into a revert; it is not — solc already reverts a high-level call whose returndata is too short to decode, so an empty address reverts either way, just anonymously. What the check actually forbids is the case that does NOT revert: code at the address that is not this registry, an EIP-7702 delegation included, free to answer zero to every migration and send every caller down its pre-migration branch. Naming that as the hazard, and naming the empty case as the one the check merely gives a name to, is accurate.

PR #106, opened 2026-08-15
test: assert the empty-derivation early return forks nothing

### LINKED ISSUE #63: The "nothing to check does not touch five RPC endpoints" early return is asserted by nothing
**Audit finding `cov-10`** — dimension 2, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RainDeployVerifyChain.sol:110-129`

## Problem

`checkDeployedOnSupportedNetworks` returns early when `derived.length == 0`, and the comment gives the reason: forking to check nothing turns an outage into the failure of an assertion that has no subject, which is the one failure this contract is supposed to be legible against.

Nothing asserts it. `RegistryDeployChainTest` inherits the contract with an empty `releasedSuites()`, so the branch is TAKEN on every CI run, but the test passes identically if the early return is deleted — it would simply fork all five networks, find nothing to check on each, and pass. The property under test is the absence of the forks, and no test observes that.

That also makes this the repo's own state today: its only chain-anchored contract exercises this branch and asserts nothing about it, while five RPC endpoints are the stated failure mode it exists to keep separate.

## Proposed fix

Add to `test/src/abstract/RegistryDeployChain.t.sol` (which already inherits the empty declaration this is about):

```solidity
    /// Nothing to check MUST NOT touch an RPC endpoint. This repo has released
    /// nothing, so this is the branch every CI run takes: forking five networks
    /// to check nothing turns an outage into the failure of an assertion with
    /// no subject, which is the one failure this contract exists to stay
    /// legible against.
    ///
    /// The absence of a fork is what is asserted, because the pass is identical
    /// either way. `vm.activeFork()` reverts when nothing is selected, so the
    /// low-level call failing IS "no network was reached".
    function testChainWithNothingToCheckForksNothing() external {
        (bool activeBefore,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
        assertFalse(activeBefore, "a fork was selected before the call");

        // The inherited matrix, over this repo's own (empty) released set.
        this.testSuitesLiveOnEverySupportedNetwork();

        (bool activeAfter,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
        assertFalse(activeAfter, "the matrix forked a network with nothing to check");
    }

    /// And the early return is about having NOTHING to check, not about the
    /// networks: handed one derivation, the matrix does fork. Without this the
    /// test above is satisfied by a matrix that never forks at all.
    function testChainWithASubjectDoesFork() external {
        DerivedDeploy[] memory derived = new DerivedDeploy[](1);
        derived[0] = DerivedDeploy({
            suite: "forks-something",
            deployedAddress: address(this),
            bytecodeHash: address(this).codehash
        });
        vm.makePersistent(address(this));

        checkDeployedOnSupportedNetworks(derived);

        (bool active,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
        assertTrue(active, "the matrix checked a subject without forking");
    }
```

`DerivedDeploy` must be imported from `../../../src/abstract/RainDeployVerifyBase.sol`, and the contract needs `Test`'s assertions, which it already inherits through `RainDeployVerifyBase`.

---

<details>

Verified against source and could not refute it.

Code reads as claimed. `src/abstract/RainDeployVerifyChain.sol:110-129` is `checkDeployedOnSupportedNetworks`, whose first statement is `if (derived.length == 0) { return; }` under a comment that states the reason ("Nothing to check is not a reason to touch five RPC endpoints... turns an outage into the failure of an assertion that has no subject"). The cited lines say exactly what the finding says they say.

Coverage: exactly three contracts inherit `RainDeployVerifyChain` (`grep -rn RainDeployVerifyChain --include=*.sol`): `RainDeployVerifyChainTest` (`test/src/abstract/RainDeployVerifyChain.t.sol`, `ExampleDeploySuites` — 2 released suites), `RainDeployVerifyChainCandidateTest` (1 released suite), and `RegistryDeployChainTest` (`test/src/abstract/RegistryDeployChain.t.sol`, an empty body over `RegistryDeploySuites`). The first two never reach the branch; the third takes it on every run — `RegistryDeploySuites.releasedSuites()` concatenates `LibAddressRegistryReleased.releasedSuites()` and `LibMigrationRegistryReleased.releasedSuites()`, both generated as `new DeploySuite[](0)`, so `deriveDeployments` returns length 0.

Nothing observes the absence of the forks. `grep -rn "activeFork\|forks nothing\|derived.length" test/ src/ script/` returns only the source line itself, the loop bound, and a doc comment in `RegistryDeployChain.t.sol` asserting the property in prose. Deleting the early return leaves the entire suite passing (all five networks fork, find nothing to check, and return), so the mutant survives with no test killing it.

Refutation attempts that failed:
- "Already covered elsewhere": no. The two chain contracts that do fork have non-empty released sets; `testChainMatrixReachesTheLastSupportedNetwork` and `testChainMatrixCoversEverySupportedNetwork` assert the loop DOES run, which is the opposite branch.
- "Deliberate documented convention": `CLAUDE.md` and the `RegistryDeployChainTest` natspec both state "it forks nothing and passes" as a repo property — they document the behaviour, they do not document a decision to leave it unasserted. A documented property with no assertion is the gap, not a restatement of a convention.
- "Untestable property": `vm.activeFork()` reverts when no fork is selected, so absence of a fork is observable; the proposed mechanism works.
- "No value": weakest refutation, but it fails too — this abstract is published as the package's product and inherited by consumer repos that have released nothing, where the guard is the difference between chain verification needing five RPC endpoints and needing none.

Severity: LOW is correct. Nothing deployed depends on this; the value at risk is CI legibility and an unnecessary five-endpoint dependency for pre-release consumer repos, not on-chain value.

Fix caveat (does not affect validity): the proposed `testChainWithASubjectDoesFork` makes `RegistryDeployChainTest` itself fork all five networks, which contradicts the very identity that contract's natspec claims ("this forks nothing") and re-imposes the RPC dependency the guard exists to remove from it. The "does fork with a subject" half belongs in `RainDeployVerifyChainTest` (which already forks) or a dedicated empty-suites fixture, leaving only the no-fork assertion in `RegistryDeployChain.t.sol`.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/63 (audit `cov-10`, LOW).

## The gap

`checkDeployedOnSupportedNetworks` returns before touching any RPC endpoint when there is nothing to check, and nothing observed it. `RegistryDeployChainTest` has released nothing, so it takes that branch on every CI run — and passed identically with the early return deleted, simply forking all five supported networks, finding nothing to check on each, and passing. The property under test was the absence of the forks, which no test looked at.

## What is asserted

**`test/src/abstract/RegistryDeployChain.t.sol` — `testChainWithNothingToCheckForksNothing`.** `vm.activeFork()` reverts when no fork is selected, so a failing low-level call to it IS "no network was reached". Two departures from the issue's sketch, both to make the assertion mean what it says:

- It runs the whole inherited entry point rather than handing the matrix an empty array, so the derivation is inside what is asserted — a fork opened while deriving touches the same five endpoints for the same nothing.
- It asserts the empty released set as a **premise**, because that is not the property. The first release gives this contract a subject and the matrix will then fork for it, correctly; without the premise this would fail at that release reporting that a matrix with nothing to check forked, which would by then be false.

**`test/src/abstract/RainDeployVerifyChain.t.sol` — `testChainWithASingleSubjectDoesFork`.** The complement, placed per the issue's own verification-block caveat: it goes in the contract that already forks every supported network, never beside the empty case, because asserting it there would hand the one contract that exists to need no RPC endpoint the five-endpoint dependency the early return removes from it.

It runs at **one** subject, and the count is the point. Every other test in that contract runs the matrix over its two released suites, so a guard keyed anywhere below two returns early on every subject in the repo and no existing test sees it — `RainDeployVerifyChainCandidateTest` is the only other contract that reaches the matrix with a subject, at exactly one, and it forks on its own before calling, so a matrix that returned without forking is indistinguishable there.

No `src/` change — this PR is tests only.

## QA

- Discriminating tests: `testChainWithNothingToCheckForksNothing` (`test/src/abstract/RegistryDeployChain.t.sol`), `testChainWithASingleSubjectDoesFork` (`test/src/abstract/RainDeployVerifyChain.t.sol`) — both pass on the unmutated baseline at `ce1e371`, and each fails under the mutation of the exact line it is about, verified by applying the mutant to `src/abstract/RainDeployVerifyChain.sol` in the working tree and re-running `forge test` (the baseline was committed first, so the revert is `git checkout -- src/...`). Neither existed on base, so "fails on base" is verified as the mutation table below: on base the guard is unasserted, and the first mutant reproduces base's own untested state.
- Mutations applied:
  - `src/abstract/RainDeployVerifyChain.sol:115-117` → delete the `if (derived.length == 0) { return; }` guard entirely → killed by `testChainWithNothingToCheckForksNothing` (`FAIL: the matrix forked a network with nothing to check`).
  - `src/abstract/RainDeployVerifyChain.sol:115` → `derived.length == 0` → `derived.length < 2` → killed by `testChainWithASingleSubjectDoesFork` (`FAIL: the matrix checked a subject without forking`). The other **8** chain tests all PASSED under this mutant — including `testChainMatrixReachesTheLastSupportedNetwork`, `testChainMatrixCoversEverySupportedNetwork`, `testChainIgnoresAnUndeployedCandidate` and the new empty-set test — which is the evidence that one subject is the only length in the repo that discriminates, and why the complement is not a duplicate of the existing "the matrix does run" tests.
- Oracle: the natspec on `checkDeployedOnSupportedNetworks` itself, which states the guard's purpose ("Nothing to check is not a reason to touch five RPC endpoints"), plus `RegistryDeployChainTest`'s contract natspec and `CLAUDE.md`, which both state "it forks nothing and passes" as a repo property. The expected values come from that stated intent, not from the implementation: the observable is `vm.activeFork()`'s documented revert-when-nothing-selected behaviour, independent of anything `RainDeployVerifyChain` computes, and the fork counts are never read back from the code under test.
- Category check: issue asks for (A) an assertion that the empty-derivation case touches no RPC endpoint, and (B) the complement so (A) is not satisfied by a matrix that never forks at all — covered A and B. The issue's collapsed verification block additionally rules that (B) must NOT live in `RegistryDeployChain.t.sol`, since that would re-impose the five-endpoint dependency on the one contract the guard frees from it; (B) is therefore in `RainDeployVerifyChainTest`, which already forks, as that ruling directs.

### DIFF
--- a/test/src/abstract/RainDeployVerifyChain.t.sol
+++ b/test/src/abstract/RainDeployVerifyChain.t.sol
@@ -147,6 +147,44 @@ contract RainDeployVerifyChainTest is ExampleDeploySuites, RainDeployVerifyChain
         assertEq(block.chainid, lastChainId);
     }
 
+    /// The early return for an empty set is about having NOTHING to check, not
+    /// about the networks: handed ONE derivation, the matrix forks.
+    ///
+    /// The count is the whole point of the case. Every other test here runs the
+    /// matrix over this contract's TWO released suites, so a guard keyed
+    /// anywhere below two — `derived.length < 2` — returns early on every
+    /// subject in the repo and is caught by none of them:
+    /// `RainDeployVerifyChainCandidateTest` is the only other contract that
+    /// reaches the matrix with a subject, at exactly one, and it forks on its own
+    /// before calling, so a matrix that returned without forking is
+    /// indistinguishable there. `RegistryDeployChainTest`'s empty case reads as
+    /// satisfied under that guard, which is what makes one subject the length
+    /// that discriminates.
+    ///
+    /// This contract is where it belongs because it already forks every
+    /// supported network. Asserting it from the empty-set side would hand the
+    /// contract that exists to need no RPC endpoint the five-endpoint dependency
+    /// the early return removes from it.
+    function testChainWithASingleSubjectDoesFork() external {
+        (bool activeBefore,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
+        assertFalse(activeBefore, "a fork was selected before the call");
+
+        // One subject, and `setUp` left it etched persistently, so it is live
+        // with its recorded hash on whichever forks the matrix creates and the
+        // check itself passes on all of them.
+        DerivedDeploy[] memory derived = new DerivedDeploy[](1);
+        derived[0] = DerivedDeploy({
+            suite: "address-registry-0-0-1",
+            deployedAddress: ADDRESS_REGISTRY_DEPLOYED_ADDRESS,
+            bytecodeHash: ADDRESS_REGISTRY_BYTECODE_HASH
+        });
+
+        checkDeployedOnSupportedNetworks(derived);
+
+        (bool activeAfter,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
+        assertTrue(activeAfter, "the matrix checked a subject without forking");
+    }
+
     /// Code on a network that is not the code the version's creation code
     /// produces MUST fail hard, naming the network and BOTH hashes.
     ///
--- a/test/src/abstract/RegistryDeployChain.t.sol
+++ b/test/src/abstract/RegistryDeployChain.t.sol
@@ -26,4 +26,37 @@ import {RegistryDeploySuites} from "../../../src/abstract/RegistryDeploySuites.s
 /// it says this and nothing more: a missing deployment or an unreachable
 /// endpoint fails here alone, leaving every snapshot assertion to answer for
 /// itself.
-contract RegistryDeployChainTest is RegistryDeploySuites, RainDeployVerifyChain {}
+contract RegistryDeployChainTest is RegistryDeploySuites, RainDeployVerifyChain {
+    /// Nothing to check MUST NOT touch an RPC endpoint. This repo has released
+    /// nothing, so this is the branch every CI run takes: forking five networks
+    /// to check nothing turns an outage into the failure of an assertion with
+    /// no subject, which is the one failure this contract exists to stay
+    /// legible against.
+    ///
+    /// The ABSENCE of a fork is what is asserted, because the pass is identical
+    /// either way — the matrix that forks all five and finds nothing to check on
+    /// each of them passes too, and is the only thing this contract would have
+    /// done differently. `vm.activeFork()` reverts when nothing is selected, so
+    /// the low-level call failing IS "no network was reached".
+    ///
+    /// It runs the whole inherited entry point rather than handing the matrix an
+    /// empty array, so the derivation is inside what is asserted: a fork opened
+    /// while deriving would touch the same five endpoints for the same nothing.
+    ///
+    /// The empty released set is asserted rather than assumed, because it is the
+    /// premise and not the property. The first release gives this contract a
+    /// subject and the matrix will fork for it — correctly — so this fails at
+    /// that release naming what actually changed, instead of reporting that a
+    /// matrix with nothing to check forked, which would by then be false.
+    function testChainWithNothingToCheckForksNothing() external {
+        assertEq(releasedSuites().length, 0, "this repo has released something, so the matrix has a subject");
+
+        (bool activeBefore,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
+        assertFalse(activeBefore, "a fork was selected before the call");
+
+        this.testSuitesLiveOnEverySupportedNetwork();
+
+        (bool activeAfter,) = address(vm).call(abi.encodeWithSignature("activeFork()"));
+        assertFalse(activeAfter, "the matrix forked a network with nothing to check");
+    }
+}


thedavidmeister:
Reviewed 0b4b6f0: ready — Closes #63. The count is the whole finding and the test is placed to match it. Every other test in that contract runs the matrix over TWO released suites, so a guard keyed anywhere below two returns early on every subject in the repo and no existing test notices: the only other contract reaching the matrix with a subject does so at exactly one, and it forks on its own beforehand, which makes a matrix that returned without forking indistinguishable there. One subject is therefore the length that discriminates, and asserting `activeFork()` before and after is what turns "it returned" into "it forked".

Putting it in the contract that already forks every supported network is also right — asserting it from the empty-set side would hand the contract that exists precisely to need no RPC a five-endpoint dependency.

PR #111, opened 2026-08-15
Suppress unsafe-typecast lint on the range-checked uint160 cast

### LINKED ISSUE #49: forge-lint `unsafe-typecast` on an already-range-checked uint160 cast is the repo's only unsuppressed warning
Raised by two code-quality reviewers, one repo-global and one per-file.

---

# Finding `Q4-04`

**Audit finding `Q4-04`** — dimension 4, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:304`

## Problem

`nix develop -c forge build` reports:

```
warning[unsafe-typecast]: typecasts that can truncate values should be checked
    --> src/lib/LibRainDeploy.sol:304:38
304 |             address actual = address(uint160(word));
```

The cast is in fact safe — line 301 does `if (word > type(uint160).max) revert ResolvedAddressReadFailed(...)` — so this is a false positive, but an unsuppressed one: it is emitted on every build of every consumer of this library, and forge-lint output is exactly the kind of thing a repo needs to keep at zero so a genuinely unchecked cast added later is visible. The lint's own remedy (a justified `disable-next-line`) is not applied.

## Proposed fix

`src/lib/LibRainDeploy.sol`, immediately above line 304:
```diff
             if (word > type(uint160).max) {
                 revert ResolvedAddressReadFailed(network, target, i, returnData);
             }
+            // Casting to `uint160` is safe because the range check directly
+            // above rejects every word with dirty upper 96 bits, so nothing
+            // reaching here can truncate.
+            // forge-lint: disable-next-line(unsafe-typecast)
             address actual = address(uint160(word));
```

---

<details>

Could not refute; reproduced directly. Ran `nix develop -c forge build` (forge 1.7.2-nightly from the repo's flake) and it emits exactly the cited warning: `warning[unsafe-typecast]` at src/lib/LibRainDeploy.sol:304:38 on `address actual = address(uint160(word));`. It is the ONLY forge-lint warning in the entire build, so the finding's premise that the repo is otherwise at zero holds as measured fact, not assumption.

Source read confirms the cited lines say what the finding claims: line 300 `uint256 word = abi.decode(returnData, (uint256));`, line 301-303 `if (word > type(uint160).max) revert ResolvedAddressReadFailed(network, target, i, returnData);`, line 304 the cast. The finding correctly characterises this as a lint false positive (the cast IS safe), and its complaint is only that the suppression is missing.

Attempts to refute all failed:
- "Deliberate org convention" — the opposite. The repo's convention is site-local suppression WITH justification: src/lib/LibRainDeploySnapshot.sol:157 already carries `// forge-lint: disable-next-line(unsafe-typecast)`, plus five `unsafe-cheatcode` suppressions in that same file, and src/concrete/MigrationRegistry.sol:117-119 carries a justified `block-timestamp` suppression. Eighteen lines above the offending cast, LibRainDeploy.sol:283-286 uses `// slither-disable-next-line low-level-calls` with a comment explicitly stating it is excluded at the site "rather than repo-wide so a low-level call added anywhere else is still reported" — the identical reasoning the finding applies. Line 304 is the one site that skipped the house style.
- "Config already mutes it" — no `[lint]` section exists in foundry.toml; nothing in CLAUDE.md sanctions lint noise.
- "A test/guard elsewhere covers it" — inapplicable; the claim is about build output, not runtime behaviour, and the runtime range check the finding relies on exists as described.
- "The fix has bad ramifications" — none. The fix is comment-only; with `bytecode_hash = "none"` and `cbor_metadata = false`, comments cannot move creation code or the pinned code hashes, and LibRainDeploy.sol is a Vm-importing script library outside the pinned contracts' compile graph.

Severity LOW stands: no on-chain value at risk (the cast is genuinely safe), but it is a real convention violation that erodes the zero-warning baseline which would otherwise make a future genuinely-unchecked cast visible.

</details>

---

# Finding `LIBDEPLOY-05`

**Audit finding `LIBDEPLOY-05`** — dimension 4, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:300-304`

## Problem

`nix develop -c forge build --force` on this commit emits exactly one lint warning across the whole tree:

```
warning[unsafe-typecast]: typecasts that can truncate values should be checked
    --> src/lib/LibRainDeploy.sol:304:38
304 |             address actual = address(uint160(word));
```

The cast is safe — line 301 refuses `word > type(uint160).max` — and the repo's convention is to say so at the site: `src/lib/LibRainDeploySnapshot.sol:157` suppresses this exact lint with `// forge-lint: disable-next-line(unsafe-typecast)`, and `src/concrete/MigrationRegistry.sol:119` does the same for `block-timestamp`. Every other suppressible lint in `src/`, `script/` and `test/` carries its comment; this one site does not, so the build is permanently one warning noisy and the next genuinely unchecked cast added anywhere arrives into output that already has a warning in it.

## Proposed fix

Suppress it at the site with the reason, matching `LibRainDeploySnapshot.sol:157`.

```solidity
            uint256 word = abi.decode(returnData, (uint256));
            if (word > type(uint160).max) {
                revert ResolvedAddressReadFailed(network, target, i, returnData);
            }
            // The word is range checked immediately above, so the low 160 bits
            // are the whole of it and the cast keeps every one.
            // forge-lint: disable-next-line(unsafe-typecast)
            address actual = address(uint160(word));
```

Verify with `nix develop -c forge build --force`, which must then report no `warning[unsafe-typecast]` at all.

---

<details>

Confirmed by running the build. `nix develop -c forge build --force` at 440e90b emits `warning[unsafe-typecast]` at src/lib/LibRainDeploy.sol:304:38 on `address actual = address(uint160(word));`, exactly as quoted. The cast is safe — lines 301-303 revert `ResolvedAddressReadFailed` when `word > type(uint160).max` — so the finding's reading of the guard is correct. The site-suppression convention is real and in-repo: LibRainDeploySnapshot.sol:157 suppresses this same lint, MigrationRegistry.sol:119 suppresses `block-timestamp`, and there are 5 more `unsafe-cheatcode` suppressions in LibRainDeploySnapshot.sol plus 9 in test/src/lib/LibRainDeploySnapshot.t.sol. This is the only unsuppressed forge-lint warning in the tree. Nothing refutes it: foundry.toml has no lint config, .github/workflows has no lint step, and test/src/lib/LibRainDeploy.t.sol covers checkResolvedAddresses behaviourally but a test cannot silence a lint. Severity is wrong, though. Two reasons. First, the finding's stated impact is partly false: the same build ALSO emits an unsuppressed solc `Warning (2018): Function state mutability can be restricted to pure` at test/src/lib/GeneratedSnapshotShape.t.sol:126, so the build output is already non-clean regardless of this line, and the "next unchecked cast arrives into already-warning output" argument does not depend on fixing this site. Second and decisive: no production value is at risk. The cast is provably safe, no CI gate fails on it (forge build exits 0; rainix-sol-static runs slither/fmt/single-contract, not forge lint), and the proposed fix is a comment. Per severity-by-prod-impact this is INFO hygiene, not LOW.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/49

## What

`checkResolvedAddresses` decodes a static-call answer as a `uint256` word and
range checks it against `type(uint160).max` — reverting `ResolvedAddressReadFailed`
on a word with dirty upper 96 bits — before casting it to an address. forge-lint
cannot see that guard, so it warned `unsafe-typecast` on the cast on every build
of every consumer of this library.

The cast is safe. This suppresses the lint at the site with the reason, which is
the house style already in this same file eighteen lines above:

- `src/lib/LibRainDeploy.sol:282-286` — `slither-disable-next-line low-level-calls`,
  with a comment saying it is excluded at the site "rather than repo-wide so a
  low-level call added anywhere else is still reported"
- `src/lib/LibRainDeploySnapshot.sol:157` — this exact lint, same directive
- `src/concrete/MigrationRegistry.sol:117-119` — `block-timestamp`, same shape

Line 304 was the one suppressible site in `src/`, `script/` or `test/` that had
skipped it.

Comment-only. No behaviour changes.

## Verification

Reproduced and then confirmed fixed with the repo's own toolchain
(`nix develop -c forge build --force`, forge 1.7.2-nightly from the flake).

Before, on `main` at `86f8d96`:

```
warning[unsafe-typecast]: typecasts that can truncate values should be checked
    ╭▸ src/lib/LibRainDeploy.sol:304:38
    │
304 │             address actual = address(uint160(word));
    │                                      ━━━━━━━━━━━━━
```

After: no `warning[unsafe-typecast]` anywhere in the build.

- `forge fmt --check` passes.
- `forge test`: identical results on this branch and on `main` — 168 passing and
  47 failing on both, and the failing test-name sets are identical between the
  two runs. Every one of those 47 is `vm.createSelectFork: environment variable
  <NETWORK>_RPC_URL not found`, i.e. the fork tests this sandbox has no `.env`
  for, exactly as `CLAUDE.md` documents. Pre-existing and unrelated; CI supplies
  the endpoints.

The pinned deploy snapshots cannot move: `bytecode_hash = "none"` and
`cbor_metadata = false`, so comments are not in the creation code, and
`LibRainDeploy.sol` is a `Vm`-importing script library that is not in the pinned
contracts' compile graph at all. `GeneratedSnapshotShapeTest`,
`LibRainDeploySnapshotTest`, `RainDeployVerifySnapshotTest` and
`RegistryDeploySnapshotTest` all pass unchanged.

## QA

- **Discriminating tests**: `testCheckResolvedAddressesDirtyWordReverts(bytes32)`
  in `test/src/lib/LibRainDeploy.t.sol` — not a test this PR adds, but the test
  that makes the suppression honest. A `forge-lint: disable` is only ever as good
  as the claim in the comment beside it, and the claim here is that the range
  check rejects every word with dirty upper 96 bits. That fuzz test drives
  exactly those words and asserts `ResolvedAddressReadFailed`, so the
  justification is enforced rather than merely asserted. Its own docstring
  already names the failure mode a wrong suppression would hide: "truncating the
  word silently PASSES this check against an address the read never gave".

  For the comment itself the discriminating check is the build, run both ways on
  this exact toolchain — mutation 1 below.
- **Mutations applied**:
  1. The new comment block at `src/lib/LibRainDeploy.sol:304` -> deleted
     (`git checkout main -- src/lib/LibRainDeploy.sol`) -> `forge build --force`
     emits `warning[unsafe-typecast]` again (count 1); restored -> count 0. So
     this diff is what removes the warning, not something incidental to it.
  2. The guard itself, `src/lib/LibRainDeploy.sol:301-303`
     `if (word > type(uint160).max) { revert ResolvedAddressReadFailed(...); }`
     -> deleted entirely, which makes the cast genuinely truncating and the new
     suppression comment a lie -> **killed by**
     `testCheckResolvedAddressesDirtyWordReverts`:
     `[FAIL: next call did not revert as expected]`.

  Harness honesty for mutation 2, because a "killed" verdict is worthless if the
  run was fake: the unmutated baseline `[PASS]`ed at `runs: 256`, so the test
  really executes; the mutant reported `Compiler run successful!`, so its failure
  is an assertion failure and not a build error masquerading as one; the guard
  was confirmed absent in the mutant (`grep -c` = 0) and present again after
  restore (= 1); and `git status` is clean afterwards.
- **Oracle**: forge-lint's own output, independent of both the source comment and
  of me — it either emits `warning[unsafe-typecast]` for that span or it does
  not. forge-lint also printed the exact remedy it wanted
  (`// forge-lint: disable-next-line(unsafe-typecast)`), and the in-repo
  convention sites listed above fix the comment's wording shape, so neither the
  directive nor the style is my invention. For mutation 2 the oracle is the
  interface contract — a read answering a non-address word must surface as
  `ResolvedAddressReadFailed` naming network, target and index — taken from that
  error's declared purpose, not from the implementation.
- **Category check**: the issue carries two findings (`Q4-04`, `LIBDEPLOY-05`)
  proposing the same single fix — suppress this lint at this one site with a
  justification. Covered. Treating the category as "suppressible lint sites
  lacking their justified suppression" rather than only the one cited line, I
  checked it is a category of one: after this change `forge build --force` emits
  no `warning[...]` forge-lint diagnostic of any kind across `src/`, `script/`
  and `test/`. The one remaining build warning is a solc `Warning (2018)`, a
  different category, deliberately out of scope — see below.

## One correction to the issue

Both findings state this was the repo's only unsuppressed warning. It was the
only unsuppressed **forge-lint** warning, which is what this fixes, but the same
build also emits an unsuppressed solc warning:

```
   --> test/src/lib/GeneratedSnapshotShape.t.sol:126:5
```

`LIBDEPLOY-05`'s own verification block already noted this and used it to argue
the severity down. It is a different category — a real code change to a test
helper, not a missing suppression — and neither finding proposed fixing it, so it
is out of scope here and left alone. Worth its own issue if a zero-warning build
is the actual goal, since this PR alone does not achieve one.

### DIFF
--- a/src/lib/LibRainDeploy.sol
+++ b/src/lib/LibRainDeploy.sol
@@ -301,6 +301,12 @@ library LibRainDeploy {
             if (word > type(uint160).max) {
                 revert ResolvedAddressReadFailed(network, target, i, returnData);
             }
+            // Casting to `uint160` is safe because the range check directly
+            // above rejects every word with dirty upper 96 bits, so the low 160
+            // bits are the whole of the word and the cast keeps every one.
+            // Excluded at the site rather than repo-wide so an unchecked cast
+            // added anywhere else is still reported.
+            // forge-lint: disable-next-line(unsafe-typecast)
             address actual = address(uint160(word));
             if (actual != expectedAddresses[i]) {
                 revert UnexpectedResolvedAddress(network, target, i, expectedAddresses[i], actual);


thedavidmeister:
Reviewed cbb5e9fb8a2cf06ca068cc307d426d267a810721: ready — Closes #49. The suppression is justified by the line directly above it: `word > type(uint160).max` reverts first, so by the time the cast runs the upper 96 bits are known clear and `uint160(word)` keeps every bit there is. Scoped to the site with `disable-next-line` rather than repo-wide, so an unchecked cast added anywhere else still reports — same shape as the `slither-disable` scoping already used on the low-level call and the timestamp comparison in this repo.

PR #112, opened 2026-08-15
test: assert [rpc_endpoints] and [etherscan] are exactly supportedNetworks()

### LINKED ISSUE #47: The supported-network list lives in three places and only the RPC half is enforced
Two reviewers raised this from different hazard partitions. The second adds the `[etherscan]` consequence the first does not trace.

---

# Finding `H6-01`

**Audit finding `H6-01`** — dimension 6, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`foundry.toml:55-76`

## Problem

HAZARD: adding (or renaming) a supported network without a matching `[etherscan]` entry becomes an ordinary, undetected edit, and the mistake surfaces only after a broadcast has spent real money from a key-custody ceremony.

SHAPE: the same fact — "which networks this org deploys to" — is spelled three times and nothing joins them. `LibRainDeploy.supportedNetworks()` (src/lib/LibRainDeploy.sol:237-245) hardcodes five names; `[rpc_endpoints]` (foundry.toml:60-65) repeats them; `[etherscan]` (foundry.toml:71-76) repeats them a third time. Exactly one of the three pairings is enforced, and only incidentally: `testZoltuFactoryCodehash` (test/src/lib/LibRainDeploy.t.sol:210-216) forks every name in `supportedNetworks()`, so a name with no `[rpc_endpoints]` alias fails. Nothing in the repo ever reads `[etherscan]` — `grep parseToml` over src/, script/ and test/ finds only `.package.version`. foundry.toml:67-70 states the failure mode in a comment ("a deploy without this section broadcasts and then fails with no API key configured for the chain — after spending the gas") and then relies on a human to keep the list in step, which is precisely the answer the hazard rule rejects.

SCENARIO: a sixth network is added — one line in `supportedNetworks()`, one `[rpc_endpoints]` alias so the fork tests go green, `[etherscan]` missed. Every CI job passes: the snapshot group never forks, the chain group forks by RPC alias only, and slither/fmt/reuse have nothing to say. The `Manual sol artifacts` dispatch is then run, `rainix-manual-sol-artifacts` passes `--verify`, the Zoltu deploy broadcasts to all six chains, and the run fails at the verification step on the sixth — gas spent, deployment landed unverified, and the operator is now debugging a release workflow with a hot key.

STRUCTURAL FIX: read the TOML the deploy actually runs under and assert both sections are exactly `supportedNetworks()`, in both directions, so the three lists can only be the same list.

## Proposed fix

Add to `test/src/lib/LibRainDeploy.t.sol` (beside `testSupportedNetworks`, line 196). `fs_permissions` already grants `{ access = "read", path = "./foundry.toml" }`, so no config change is needed:

```solidity
/// PROPERTY: `[rpc_endpoints]` and `[etherscan]` are EXACTLY
/// `supportedNetworks()`. The deploy forks by the first and `--verify` resolves
/// the second, so a name missing from either is a broadcast that spends gas and
/// then fails, and a name in either that no suite deploys to is a config nobody
/// reads. Asserted in both directions so the three lists are one list.
function testEverySupportedNetworkIsFullyConfigured() external view {
    string memory toml = vm.readFile("foundry.toml");
    string[] memory networks = LibRainDeploy.supportedNetworks();

    for (uint256 i = 0; i < networks.length; i++) {
        assertTrue(
            vm.keyExistsToml(toml, string.concat(".rpc_endpoints.", networks[i])),
            string.concat("supported network has no [rpc_endpoints] alias: ", networks[i])
        );
        assertTrue(
            vm.keyExistsToml(toml, string.concat(".etherscan.", networks[i])),
            string.concat("supported network has no [etherscan] entry: ", networks[i])
        );
    }

    assertEq(
        vm.parseTomlKeys(toml, ".rpc_endpoints").length,
        networks.length,
        "[rpc_endpoints] declares a network nothing deploys to"
    );
    assertEq(
        vm.parseTomlKeys(toml, ".etherscan").length,
        networks.length,
        "[etherscan] declares a network nothing deploys to"
    );
}
```

---

<details>

Could not refute. Every cited fact checks out against source. supportedNetworks() (src/lib/LibRainDeploy.sol:237-245) hardcodes five names whose constants (lines 67-79) are exactly the config aliases; [rpc_endpoints] (foundry.toml:60-65) and [etherscan] (foundry.toml:71-76) restate the same list twice more. Grep over src/, script/, test/ for parseToml|readFile|keyExistsToml|etherscan finds nothing reading the [etherscan] section — only .package.version in src/lib/LibRainDeploySnapshot.sol:87, artifact JSON reads, and snapshot record reads. The RPC half is enforced only incidentally, by testZoltuFactoryCodehash (test/src/lib/LibRainDeploy.t.sol:210-216) and testChainMatrixCoversEverySupportedNetwork (test/src/abstract/RainDeployVerifyChain.t.sol:233-) forking each name. testSupportedNetworks (line 196) locks the library list only. The deploy path is real: .github/workflows/manual-sol-artifacts.yaml dispatches rainix-manual-sol-artifacts@main, and foundry.toml:67-70 documents the broadcast-then-fail-after-gas mode itself while relying on a human to keep the lists in step. fs_permissions already grants read of ./foundry.toml, so the proposed test needs no config change. The documenting comment is not a deliberate decision against enforcement, so the "documented org convention" escape does not apply.

Severity corrected MEDIUM -> LOW on value-at-risk. The consequence is gas for one small registry deploy on cheap chains plus a failed verification step at the end of a run. The contract still lands at its deterministic CREATE2 address with the pinned codehash, so consumer pins and bytecode are unaffected; recovery is adding the entry and running a standalone forge verify-contract, which needs an API key rather than the deploy key. Zoltu deploys are idempotent and permissionless, and deployToNetworks skips already-deployed addresses, so no re-ceremony over new value. Trigger is rare (adding a sixth network, which already forces an edit to testSupportedNetworks). No funds, no wrong code, fully recoverable: LOW.

</details>

---

# Finding `HZ6-03`

**Audit finding `HZ6-03`** — dimension 6, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`foundry.toml:60-76`

## Problem

HAZARD: adding one network requires editing `LibRainDeploy.supportedNetworks()` (src/lib/LibRainDeploy.sol:66-79, 237-245), `[rpc_endpoints]` (foundry.toml:60-65) and `[etherscan]` (foundry.toml:71-76), plus provisioning a `CI_DEPLOY_<NET>_ETHERSCAN_API_KEY` secret in the rainix org. Three lists of the same set, in two languages, with one of them enforced only as a side effect.

SHAPE: `testZoltuFactoryCodehash` (test/src/lib/LibRainDeploy.t.sol:210-216) forks every name `supportedNetworks()` returns, so a name with no `[rpc_endpoints]` alias fails there. Nothing anywhere reads `[etherscan]`. foundry.toml's own comment states the requirement — "One entry per `[rpc_endpoints]` alias, because the deploy goes to all of them" — and states the consequence — "a deploy without this section broadcasts and then fails with no API key configured for the chain — after spending the gas" — and then leaves it to memory. The repo already grants tests read access to foundry.toml (foundry.toml:38), so the cross-check is available and simply absent.

PRODUCTION SCENARIO: a sixth network is added to `supportedNetworks()` and `[rpc_endpoints]`, and `[etherscan]` is missed. Every test is green, because the only network assertion in the suite is satisfied by the RPC alias. `Manual sol artifacts` is dispatched, `rainix-manual-sol-artifacts` passes `--verify`, the Zoltu deploy broadcasts on the new chain, and the run then dies with no API key configured — real gas spent, an unverified contract on the explorer, and a workflow run that has to be diagnosed rather than simply retried. The same shape fires in reverse for a `[rpc_endpoints]` alias with no `supportedNetworks()` entry: that network is never deployed to and never verified, and nothing goes red.

STRUCTURAL FIX: make `supportedNetworks()` the one source and assert both foundry.toml sections against it, in both directions.

## Proposed fix

Add to `test/src/lib/LibRainDeploy.t.sol`, beside `testSupportedNetworks` (foundry.toml:38 already grants the read):

```solidity
/// PROPERTY: every supported network is FULLY configured, and nothing is
/// configured that is not supported. `supportedNetworks()` is the one source:
/// a fork alias without a network is a chain nothing deploys to or verifies,
/// and a network without an explorer key broadcasts and then fails `--verify`
/// after the gas is spent.
function testSupportedNetworksAreFullyConfigured() external view {
    string memory config = vm.readFile("foundry.toml");
    string[] memory networks = LibRainDeploy.supportedNetworks();

    for (uint256 i = 0; i < networks.length; i++) {
        assertTrue(
            vm.keyExistsToml(config, string.concat(".rpc_endpoints.", networks[i])),
            string.concat("supported network has no [rpc_endpoints] alias: ", networks[i])
        );
        assertTrue(
            vm.keyExistsToml(config, string.concat(".etherscan.", networks[i])),
            string.concat("supported network has no [etherscan] key: ", networks[i])
        );
    }

    // Counted as well as contained, so a section carrying a name
    // `supportedNetworks()` does not — a chain nothing ever deploys to or
    // verifies — is caught by the same property.
    assertEq(
        vm.parseTomlKeys(config, ".rpc_endpoints").length,
        networks.length,
        "[rpc_endpoints] and supportedNetworks() are different sets"
    );
    assertEq(
        vm.parseTomlKeys(config, ".etherscan").length,
        networks.length,
        "[etherscan] and supportedNetworks() are different sets"
    );
}
```

This is a `pure`-adjacent view test with no RPC dependency, so it fails on the PR that adds the network rather than at dispatch time. Note it is deliberately two-directional: containment alone passes for a section carrying an extra alias, and equal counts alone pass for a swapped pair.

---

<details>

SURVIVES. Every cited fact checks out against source at 440e90b.

Three lists confirmed: `LibRainDeploy.sol:66-79` (network name constants) + `:235-244` (`supportedNetworks()` building the 5-element array), `foundry.toml:60-65` `[rpc_endpoints]`, `foundry.toml:71-76` `[etherscan]`. Plus the org-side `CI_DEPLOY_<NET>_ETHERSCAN_API_KEY` secrets implied by `.github/workflows/manual-sol-artifacts.yaml` (`secrets: inherit` into `rainix-manual-sol-artifacts@main`).

The "unenforced" claim survives a grep of the whole suite: `grep -rn "etherscan|rpc_endpoints|parseTomlKeys|keyExistsToml|readFile" test/ src/ script/ --include=*.sol` finds no code anywhere that reads the `[etherscan]` section — only doc comments in `RainDeployVerifyChain.sol:10,23` and `RainDeployBroadcast.sol:61` that reference `[rpc_endpoints]` in prose. `testSupportedNetworks` (`test/src/lib/LibRainDeploy.t.sol:194-202`) asserts length 5 and order against the library constants only; it does not open foundry.toml. `testZoltuFactoryCodehash` (`:210-216`) forks every `supportedNetworks()` name, which does enforce the RPC-alias direction as a side effect exactly as the finding says, and enforces nothing about `[etherscan]` or about the reverse direction. `RainDeployVerifyChain.sol:119` iterates the same list but is released-only and currently has no released suite, so it is not a second net.

The proposed fix is mechanically available: `foundry.toml:38` already grants read on `./foundry.toml`, and `LibRainDeploySnapshot.sol:87` already reads/parses that file, so a view-only TOML cross-check needs no config change and no RPC.

Not a documented org convention: the foundry.toml comment at `:67-70` states the requirement and the consequence and then relies on memory, which is the hazard, not a deliberate design.

SEVERITY CORRECTED MEDIUM -> LOW. Severity is value-at-risk in production. A missing `[etherscan]` entry does not corrupt deployed state: the Zoltu factory is CREATE2 over calldata under a zero salt, so the contract still lands at the correct deterministic address with the correct codehash on the new chain; the pins, snapshots and consumer-visible addresses are unaffected. What is actually lost is explorer verification, a trivial amount of L2 gas, and a manual run that has to be diagnosed. Deploying is idempotent, so recovery is fix-the-toml and re-dispatch or verify out of band. The path is `workflow_dispatch` only, one suite per dispatch, human-driven and watched — not an automated pipeline that can silently repeat. The reverse direction (an `[rpc_endpoints]` alias with no `supportedNetworks()` entry) also requires missing the primary edit the whole repo centers on. Real defect, worth the test, but LOW blast radius.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/47

## What was wrong

"Which networks this org deploys to" was spelled three times with nothing joining them:

- `LibRainDeploy.supportedNetworks()` (`src/lib/LibRainDeploy.sol:237-245`)
- `[rpc_endpoints]` (`foundry.toml:60-65`)
- `[etherscan]` (`foundry.toml:71-76`)

Only one pairing was enforced, and only incidentally: `testZoltuFactoryCodehash` forks every name `supportedNetworks()` returns, so a name with no `[rpc_endpoints]` alias fails there. Nothing in the repo read `[etherscan]` at all — grep over `src/`, `script/` and `test/` for `readFile|parseToml|keyExists` finds only `.package.version`, artifact JSON reads and snapshot record reads.

The consequence is the one `foundry.toml:67-70` already documents and then leaves to memory: `rainix-manual-sol-artifacts` passes `--verify`, so a network in `supportedNetworks()` with no `[etherscan]` entry broadcasts, spends the gas, and *then* fails with no API key configured for the chain.

## The fix

`testSupportedNetworksAreFullyConfigured` in `test/src/lib/LibRainDeploy.t.sol` reads `foundry.toml` and asserts both sections are EXACTLY `supportedNetworks()`.

Asserted in **both directions, by membership**:

- every supported network has an `[rpc_endpoints]` alias and an `[etherscan]` key — a missing one is the broadcast-then-fail above;
- every `[rpc_endpoints]` alias and every `[etherscan]` key is a supported network — a stray one is config nothing ever reads.

One direction alone is not enough: containment forwards passes for a section carrying an alias nothing deploys to, and containment backwards passes for a network with no config at all. Membership both ways makes the sets equal, so the three lists can only be one list.

Membership rather than the key counts the issue proposed. It is strictly stronger — equal counts plus forward containment is only set equality if `supportedNetworks()` has no duplicates, which this test would then be leaning on `testSupportedNetworks` to guarantee — and the failure names the offending alias instead of only reporting that two numbers differ.

The raw file is read rather than forge's resolved config, because the values are `${VAR}` interpolations that only exist in CI. The keys are the whole contract here and they are in the text — so the test needs no RPC and fails on the PR that drifts, rather than at dispatch time with a hot key.

`fs_permissions` already granted `{ access = "read", path = "./foundry.toml" }` and `LibRainDeploySnapshot.deployTag` already reads that file the same way, so no config change was needed.

The only non-test change is a comment on `[etherscan]` naming the test that now enforces it, so the next person editing that section learns it is checked rather than rediscovering the hazard.

## Mutation evidence

Five mutants, one per drift direction. Every run was checked for `1 total tests` as well as for `[FAIL`, so a mutant that failed to compile could not be miscounted as killed.

| # | mutant | killed with |
|---|---|---|
| M1 | 6th network in `supportedNetworks()`, no config at all | `supported network has no [rpc_endpoints] alias: optimism` |
| M2 | 6th network + `[rpc_endpoints]` alias, `[etherscan]` missed — **the hazard in the issue** | `supported network has no [etherscan] key: optimism` |
| M3 | stray `[rpc_endpoints]` alias | `[rpc_endpoints] alias is not a supported network: optimism` |
| M4 | stray `[etherscan]` key | `[etherscan] key is not a supported network: optimism` |
| M5 | `[etherscan]` entry deleted | `supported network has no [etherscan] key: polygon` |

M2 is the scenario both findings describe end to end, and it is the one nothing in the repo caught before this test.

## Scope

Test-only plus one comment. No behaviour change: `supportedNetworks()`, `[rpc_endpoints]` and `[etherscan]` are all untouched and already agree.

## QA
- Discriminating tests: `testSupportedNetworksAreFullyConfigured` — fails on base under each of the five drifts below (verified by applying each mutation to a base tree via `.fixes/mutate.sh` and running `forge test --match-test`; each run asserted `1 total tests` actually RAN as well as `[FAIL`, so a non-compiling mutant cannot be miscounted as killed). It is a new test: before it nothing in the repo read `[etherscan]`, so the base suite is green on every mutant below.
- Mutations applied: `LibRainDeploy.supportedNetworks()` `new string[](5)` → `new string[](6)` + `networks[5] = "optimism"` → killed by `testSupportedNetworksAreFullyConfigured` (`supported network has no [rpc_endpoints] alias: optimism`); that same mutation plus `optimism` added to `[rpc_endpoints]` — the exact hazard the issue describes — → killed by `testSupportedNetworksAreFullyConfigured` (`supported network has no [etherscan] key: optimism`); `foundry.toml` `[rpc_endpoints]` += `optimism` with the library untouched → killed (`[rpc_endpoints] alias is not a supported network: optimism`); `foundry.toml` `[etherscan]` += `optimism` with the library untouched → killed (`[etherscan] key is not a supported network: optimism`); `foundry.toml` `[etherscan]` −= `polygon` → killed (`supported network has no [etherscan] key: polygon`). 5 mutants, 5 killed.
- Oracle: `LibRainDeploy.supportedNetworks()`, which is what the deploy actually iterates (`RainDeployBroadcast.deployNetworks()` defaults to it). The expected key sets are never restated in the test — they are read from `supportedNetworks()` at runtime and compared against keys parsed out of the `foundry.toml` text, so the test cannot drift from the list it is about. The consequence being guarded is stated independently by `foundry.toml:67-70` and by `rainix-manual-sol-artifacts` passing `--verify`.
- Category check: the issue asks for both `[rpc_endpoints]` and `[etherscan]` asserted against `supportedNetworks()` in both directions; covered — `[rpc_endpoints]` forward and reverse, `[etherscan]` forward and reverse. Implemented by membership rather than the proposed key counts: strictly stronger (count equality is only set equality given no duplicates in `supportedNetworks()`, which the count form would be leaning on `testSupportedNetworks` to supply) and it names the offending alias on failure instead of only reporting that two numbers differ. No `fs_permissions` change was needed, as both findings predicted.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -68,6 +68,10 @@ polygon = "${POLYGON_RPC_URL}"
 # these variable names, so a deploy without this section broadcasts and then
 # fails with no API key configured for the chain — after spending the gas. One
 # entry per `[rpc_endpoints]` alias, because the deploy goes to all of them.
+#
+# Both sections are checked against `LibRainDeploy.supportedNetworks()`, in both
+# directions, by `testSupportedNetworksAreFullyConfigured`. Adding a network is
+# an edit to all three or a red test, not a broadcast that discovers it.
 [etherscan]
 arbitrum = { key = "${CI_DEPLOY_ARBITRUM_ETHERSCAN_API_KEY}" }
 base = { key = "${CI_DEPLOY_BASE_ETHERSCAN_API_KEY}" }
--- a/test/src/lib/LibRainDeploy.t.sol
+++ b/test/src/lib/LibRainDeploy.t.sol
@@ -203,6 +203,70 @@ contract LibRainDeployTest is Test {
         assertEq(networks[4], LibRainDeploy.POLYGON);
     }
 
+    /// Whether `names` holds `name`. A config section is keyed, not ordered, so
+    /// membership is the only thing worth asserting across one.
+    /// @param names The names to search.
+    /// @param name The name to look for.
+    /// @return Whether it is there.
+    function holdsName(string[] memory names, string memory name) internal pure returns (bool) {
+        for (uint256 i = 0; i < names.length; i++) {
+            if (keccak256(bytes(names[i])) == keccak256(bytes(name))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// PROPERTY: `[rpc_endpoints]` and `[etherscan]` in `foundry.toml` are
+    /// EXACTLY `supportedNetworks()`, which makes the three lists one list.
+    ///
+    /// The deploy forks by the first and `--verify` resolves the second, so a
+    /// supported network missing from either broadcasts and then fails after
+    /// the gas is spent, and a section entry no supported network names is
+    /// config nothing ever reads. Both are the same defect — the lists having
+    /// drifted — so both directions are asserted, by membership: containment
+    /// one way alone passes for a section carrying an alias nothing deploys
+    /// to, and the other way alone passes for a network with no config at all.
+    ///
+    /// This is what makes the `[etherscan]` half enforced at all. The RPC half
+    /// is enforced only incidentally, by the fork tests, and only forwards.
+    ///
+    /// The raw file is read rather than forge's resolved config because the
+    /// values are `${VAR}` interpolations that exist only in CI. The KEYS are
+    /// the whole contract here, and they are in the text — so this needs no
+    /// RPC and fails on the PR that drifts rather than at dispatch time.
+    function testSupportedNetworksAreFullyConfigured() external view {
+        string memory config = vm.readFile("foundry.toml");
+        string[] memory networks = LibRainDeploy.supportedNetworks();
+
+        for (uint256 i = 0; i < networks.length; i++) {
+            assertTrue(
+                vm.keyExistsToml(config, string.concat(".rpc_endpoints.", networks[i])),
+                string.concat("supported network has no [rpc_endpoints] alias: ", networks[i])
+            );
+            assertTrue(
+                vm.keyExistsToml(config, string.concat(".etherscan.", networks[i])),
+                string.concat("supported network has no [etherscan] key: ", networks[i])
+            );
+        }
+
+        string[] memory rpcAliases = vm.parseTomlKeys(config, ".rpc_endpoints");
+        for (uint256 i = 0; i < rpcAliases.length; i++) {
+            assertTrue(
+                holdsName(networks, rpcAliases[i]),
+                string.concat("[rpc_endpoints] alias is not a supported network: ", rpcAliases[i])
+            );
+        }
+
+        string[] memory etherscanKeys = vm.parseTomlKeys(config, ".etherscan");
+        for (uint256 i = 0; i < etherscanKeys.length; i++) {
+            assertTrue(
+                holdsName(networks, etherscanKeys[i]),
+                string.concat("[etherscan] key is not a supported network: ", etherscanKeys[i])
+            );
+        }
+    }
+
     /// `ZOLTU_FACTORY_CODEHASH` MUST match the actual codehash of the Zoltu
     /// factory on every supported network. Every name in `supportedNetworks`
     /// MUST also be a configured fork alias, otherwise it cannot be deployed


thedavidmeister:
Reviewed b3e5c259ddfee95be9c0808e0c05e84f23342965: ready — Closes #47. Asserts both directions, which is what makes it worth having: containment one way alone passes for a section carrying an alias nothing deploys to, and the other way alone passes for a supported network with no config at all. Both are the same drift.

Reading the raw `foundry.toml` rather than forge's resolved config is right, not a shortcut — the values are `${VAR}` interpolations that only exist in CI, while the KEYS are the entire contract being asserted and they are in the text. So this needs no RPC and goes red on the PR that drifts rather than at dispatch time, which is exactly where the `[etherscan]` half was previously enforced by nothing at all.

Directly relevant today: this repo's two deploy dispatches broadcast and then failed at verification, and a missing `[etherscan]` key is one of the ways that happens after the gas is spent.

PR #116, opened 2026-08-15
docs: replace dangling Extrospection reference with the codehash mechanism

### LINKED ISSUE #65: Dangling reference to the "Rain Extrospection lib", which nothing in the tree uses
**Audit finding `D3-02`** — dimension 3, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`README.md:36-37`

## Problem

The Approach list claims "Bytecode integrity checks (e.g. via the Rain Extrospection lib) supported post-deploy." No file in `src/`, `script/`, `test/`, `foundry.toml`, `soldeer.lock` or `remappings.txt` mentions Extrospection — the only dependencies are `forge-std` and `rain-sol-codegen`. Post-deploy bytecode integrity in this repo is done with raw `address.codehash` comparisons (`LibRainDeploy.deployToNetworks` line 444, `RainDeployVerifyChain.checkDeployedOnNetwork` lines 93-98, `LibAddressRegistry.resolve` lines 46-51, `LibMigrationRegistry.checkCodeHash` lines 93-100). The bullet points a reader at an external library that plays no part in the design.

## Proposed fix

README.md lines 36-37, replace:

```markdown
- Bytecode integrity checks (e.g. via the Rain Extrospection lib) supported
  post-deploy.
```

with:

```markdown
- Bytecode integrity checks post-deploy: `codehash` is compared against the
  recorded pin on every network by `RainDeployVerifyChain`, and before every
  registry read by `LibAddressRegistry` and `LibMigrationRegistry`.
```

---

<details>

Could not refute. Verified at 440e90b:

- README.md:36-37 reads verbatim: "- Bytecode integrity checks (e.g. via the Rain Extrospection lib) supported / post-deploy." The quote and line numbers are correct.
- `grep -rni extrospection` across the whole tree (excluding .git/dependencies) hits ONLY README.md:36. Nothing in src/, script/, test/, foundry.toml, remappings.txt or soldeer.lock references it. remappings.txt has exactly two entries (forge-std-1.16.1, rain-sol-codegen-0.1.6) and soldeer.lock exactly those two dependencies, so the lib is not even available to be used.
- CLAUDE.md never mentions Extrospection either; it documents the codehash-based checks instead.
- The mechanism the finding names as the real one is what the source does: `RainDeployVerifyChain.checkDeployedOnNetwork` (src/abstract/RainDeployVerifyChain.sol:88-100) reverts NotDeployedOnNetwork on empty code then compares `derived.deployedAddress.codehash` to `derived.bytecodeHash`; LibRainDeploy.sol:444-445 reverts UnexpectedDeployedCodeHash; `LibAddressRegistry.resolve` (src/lib/LibAddressRegistry.sol:46-51) and `LibMigrationRegistry.checkCodeHash` (src/lib/LibMigrationRegistry.sol:92-99) gate every registry touch on a codehash pin. All four cited sites exist and say what the finding says.

Refutations considered and rejected: (a) "e.g. is merely illustrative" — the example points at a library absent from the repo, its docs and its dependency set, so it describes an integration that does not exist; (b) "a test or convention covers it" — no test asserts README prose, and no org convention documents Extrospection as an assumed sibling lib; (c) "the bullet is still true, integrity checks ARE supported" — true, which is why this is prose accuracy only, not a false behavioural claim.

Severity corrected LOW -> INFO. Value-at-risk in production is nil: the integrity checks are inherited automatically by any consumer of RainDeployVerifyChain/RainDeployVerifySnapshot and by every registry read, so a consumer never chooses or wires a bytecode-check mechanism and the stale example cannot cause a misconfiguration. The proposed replacement text is factually accurate against the four call sites and can be applied as written.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/65

The Approach list pointed readers at the Rain Extrospection lib as the example
mechanism for post-deploy bytecode integrity. Nothing in this repo uses it:
`grep -rni extrospection` over the tree hit only that one README line, and
`remappings.txt` / `soldeer.lock` carry exactly two dependencies
(`forge-std`, `rain-sol-codegen`), so the lib is not even available to be used.
The bullet described an integration that does not exist.

Replaced with what the source actually does — `codehash` comparison — naming
the contracts a reader can go read.

## Deviation from the issue's proposed text

The issue proposed `...and before every registry **read** by LibAddressRegistry
and LibMigrationRegistry`. That is accurate for `LibAddressRegistry`, whose only
function is the read `resolve`, but narrower than `LibMigrationRegistry`, which
gates the write `applyMigration` (line 168) as well as the reads `applied`
(122) and `head` (139). The library's own comment is explicit that this is the
point: "Both entry points check ... a read would branch a test on whatever that
code returned, and a write would record a migration somewhere nothing will ever
read it."

Shipping "read" verbatim would have swapped a stale-example inaccuracy for a
too-narrow one, so the wording is `access`, the property both libraries share.
The issue's own verification block used the same wider phrasing ("gate every
registry touch on a codehash pin"); only the proposed-fix line narrowed it.

## Verification

Every call site named in the new text was read at this branch's base
(`86f8d96`) and does what the text says:

- `src/abstract/RainDeployVerifyChain.sol:88-100` — `checkDeployedOnNetwork`
  reverts `NotDeployedOnNetwork` on empty code, then compares
  `derived.deployedAddress.codehash` against `derived.bytecodeHash`.
- `src/lib/LibAddressRegistry.sol:45-52` — `resolve` compares the registry's
  `codehash` against `ADDRESS_REGISTRY_DEPLOYED_CODEHASH` before the `get` call.
- `src/lib/LibMigrationRegistry.sol:93-100` — `checkCodeHash` compares against
  `MIGRATION_REGISTRY_DEPLOYED_CODEHASH`; called from all three entry points.

Post-change `grep -rni extrospection` over the tree (excluding `.git`,
`dependencies`, `out`) returns nothing.

Documentation only — no Solidity changed.

## QA
- Discriminating tests: n/a — no test asserts README prose, and adding one would
  pin wording rather than behaviour. The discriminating check is
  `git grep -ni extrospection <rev> -- README.md src script test foundry.toml
  remappings.txt soldeer.lock`, which fails on base (`86f8d96:README.md:36`) and
  returns no hits at this head.
- Mutations applied: n/a — docs-only diff, no executable line to mutate. Repo
  gates run: `reuse lint` exit 0 (71/71 files compliant); pre-commit
  (denofmt, no-consumer-prettier) passed on the commit.
- Oracle: the Solidity source, read independently of the README. Each contract
  the new text names was opened and confirmed to compare `codehash`:
  `RainDeployVerifyChain.checkDeployedOnNetwork` (src/abstract/
  RainDeployVerifyChain.sol:88-100), `LibAddressRegistry.resolve`
  (src/lib/LibAddressRegistry.sol:45-52), and `LibMigrationRegistry.checkCodeHash`
  (src/lib/LibMigrationRegistry.sol:93-100), whose three callers were enumerated
  by grep to establish that "access" and not "read" is the correct word.
- Category check: issue #65 asks one thing — remove the dangling Extrospection
  reference and describe the real mechanism. Covered. The grep that found the
  finding was re-run tree-wide post-change and is clean, so the category
  (dangling references to this absent lib) is exhausted, not just the cited line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,9 @@ Approach:
 - Hard guards against deploying to networks where dependencies are missing.
 - Pre-calculated addresses asserted against the creation code before deploying,
   and against the chain after: silent failures fail loudly.
-- Bytecode integrity checks (e.g. via the Rain Extrospection lib) supported
-  post-deploy.
+- Bytecode integrity checks post-deploy: `codehash` is compared against the
+  recorded pin on every network by `RainDeployVerifyChain`, and before every
+  registry access by `LibAddressRegistry` and `LibMigrationRegistry`.
 - An address registry, read at run time rather than compiled into creation code,
   and a post-deploy check that every target network's deployment took the
   address it was supposed to.


thedavidmeister:
Reviewed f1a6f834ed70364ca7ce9426cbadb81ba9bddadf: ready — Closes #65. README only. The old bullet pointed at the Rain Extrospection lib, which this repo does not use; the replacement names the mechanism it actually has — `codehash` against the recorded pin, by `RainDeployVerifyChain` on every network, and before every registry access by `LibAddressRegistry` and `LibMigrationRegistry`. I checked the second half: `LibMigrationRegistry.checkCodeHash()` is called by `applied`, `head` and `applyMigration`, so "before every registry access" is accurate rather than approximate.

PR #119, opened 2026-08-15
Refuse a leading zero on a release-version component

### LINKED ISSUE #51: isStrictTriple accepts leading zeros, so one release version has two directory spellings and SnapshotAlreadyFrozen never sees the second as a re-cut
**Audit finding `LIBDEPLOY-04`** — dimension 5, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploySnapshot.sol:103-125`

## Problem

`isStrictTriple` accepts any non-empty run of digits per component, so `0.01.5` is as strict a triple as `0.1.5` and `tagForVersion` maps it to the directory `0_01_5`, which `isTag` accepts as a release. Two directories then describe one release:

- `freeze` (742-750) refuses a re-cut by `vm.exists(dirForSnapshot(tag))`. With `0_1_5` already frozen, bumping `[package].version` to `0.01.5` makes `frozenDir` a path that does not exist, so the same release is frozen a second time under a second name — the append-only guarantee the error exists for does not fire.
- `recordPrecedes` (445-455) reads both tags through `vm.parseUint`, so `0_01_5` and `0_1_5` compare EQUAL as versions and fall through to the byte-for-byte path tie break, which orders them by the accident of a zero character rather than by release.
- both are then declared as separate released suites (`releasedLibraryBlock` appends the tag to the key), and both are demanded live on every network by the chain group.

The file states its own rule as "the ONE definition of the release-version shape" whose purpose is that two spellings cannot drift apart — but it admits two spellings of one version, which is the same class of defect one step down. The version is also the soldeer `[package].version`, where `0.01.5` is not a version any consumer can resolve.

## Proposed fix

Refuse a leading zero on a multi-digit component, so a component has exactly one spelling. `0` itself stays legal — `0.0.0` is a real version and is pinned by `testIsTagAcceptsWhatAFreezeCanWrite`.

In `isStrictTriple`, replace the digit branch (117-118):

```solidity
            } else if (char >= "0" && char <= "9") {
                // A component is one number, so it has one spelling. `01` and
                // `1` are the same release and would freeze to two directories,
                // neither of which `SnapshotAlreadyFrozen` sees as the other,
                // and which `recordPrecedes` cannot order because they compare
                // equal as versions.
                if (digitsInComponent == 1 && subjectBytes[i - 1] == "0") {
                    return false;
                }
                digitsInComponent++;
            } else {
```

(`i` is at least 1 wherever `digitsInComponent == 1`, because that digit was read at an earlier index.)

Tests (test/src/lib/LibRainDeploySnapshot.t.sol):

```solidity
    /// A component has ONE spelling. A padded component is the same release as
    /// its unpadded form: it freezes to a second directory the immutability
    /// check does not recognise as the first, and the two compare equal as
    /// versions so nothing can order them either.
    function testIsStrictTripleRefusesLeadingZeros() external pure {
        assertFalse(LibRainDeploySnapshot.isStrictTriple("0.01.5", "."));
        assertFalse(LibRainDeploySnapshot.isStrictTriple("01.1.5", "."));
        assertFalse(LibRainDeploySnapshot.isStrictTriple("0.1.05", "."));
        assertFalse(LibRainDeploySnapshot.isStrictTriple("00.0.0", "."));
        assertFalse(LibRainDeploySnapshot.isTag("0_01_5"));

        // The boundary: a single zero IS the number zero, and `0.0.0` is a real
        // version.
        assertTrue(LibRainDeploySnapshot.isStrictTriple("0.0.0", "."));
        assertTrue(LibRainDeploySnapshot.isStrictTriple("0.10.0", "."));
        assertTrue(LibRainDeploySnapshot.isStrictTriple("10.0.100", "."));
        assertTrue(LibRainDeploySnapshot.isTag("0_0_0"));
    }

    /// The refusal MUST be reachable through the release path, naming the
    /// version, rather than only through the predicate.
    function testTagForVersionRefusesLeadingZeros() external {
        string[3] memory bad = ["0.01.5", "01.1.5", "0.1.05"];
        for (uint256 i = 0; i < bad.length; i++) {
            vm.expectRevert(abi.encodeWithSelector(UnreleasableVersion.selector, bad[i]));
            this.externalTagForVersion(bad[i]);
        }
    }
```

The existing fuzz `testEveryFreezableVersionIsATagTheRecordFinds` keeps passing: `vm.toString` of a `uint256` never emits a leading zero.

---

<details>

Could not refute; every mechanical claim verified against source and empirically against a forge probe (since removed).

Source read in full: /home/gildlab/code/rain-deploy-audit/src/lib/LibRainDeploySnapshot.sol.
- `isStrictTriple` (103-125) counts digits per component and never inspects the first digit, so `0.01.5` is accepted. `isTag` is the same predicate with `_`. `tagForVersion` (146-158) maps byte-for-byte, so `0.01.5` -> `0_01_5`.
- `freeze` (742-750) gates a re-cut solely on `vm.exists(dirForSnapshot(tag))`, and `frozenSnapshotPaths` (217-250) admits any `isTag` directory, so `0_01_5` is a second, independently frozen, independently declared release of the same code.
- `recordPrecedes` (445-455) reads components with `vm.parseUint`.

Probe (ran under `nix develop -c forge test`, PASS, file deleted afterwards): `isStrictTriple("0.01.5", ".")` true, `isTag("0_01_5")` true, `tagForVersion("0.01.5") == "0_01_5"`, `vm.parseUint("01") == 1`, `vm.parseUint("00") == 0`, and `recordPrecedes(vm, "src/generated/0_01_5/A.sol", "src/generated/0_1_5/A.sol")` true — i.e. the two tags do compare EQUAL as versions and are ordered by the zero character in the byte-for-byte tiebreak, exactly as described.

Grepped the test tree: test/src/lib/LibRainDeploySnapshot.t.sol covers `0_1_7-rc1`, `0_1`, `0_1_`, `_1_7`, `""`, `0.1.7 `, `0..7`, `a.b.c` — nothing anywhere asserts anything about a leading zero, and the fuzz `testEveryFreezableVersionIsATagTheRecordFinds` builds versions with `vm.toString(uint256)`, which never emits one. No guard elsewhere: the only version canonicality check in the repo is this predicate (`deployTag` -> `tagForVersion`), and RainDeployVerifySnapshot checks declaration-vs-record, which a duplicate spelling satisfies on both sides. src/generated currently holds only `candidate/`, so the scenario is prospective, not present.

Not refuted on "correct behaviour" grounds either: the file states its own job as the ONE canonical release-version shape and already refuses cosmetic-only deviations (trailing space, rc suffix), so admitting two spellings of one version is inside its declared responsibility, and `recordPrecedes` establishes internally that tags are numbers — which is precisely what makes the two directories indistinguishable as releases while remaining distinct as paths.

Severity corrected LOW -> INFO. Value at risk in production is negligible: reaching it requires a maintainer to hand-write a non-semver `[package].version` (releases are triggered by a hand-typed `sol-v*` git tag), and the worst outcome is a duplicate frozen directory whose contracts are the same bytecode at the same CREATE2 addresses (so every chain check still passes) plus a Soldeer publish under an unresolvable version, which is loud. No consensus value, no address, and no consumer pin is wrong in any branch — the already-frozen bytes are never replaced, so the immutability the error protects is not itself breached; what is breached is "one release is cut once".

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/51

`isStrictTriple` accepted any non-empty run of digits per component, so `0.01.5`
was as strict a triple as `0.1.5`. That made one release have two spellings, and
every guard built on the predicate reads them as two different releases:

- `freeze` refuses a re-cut on `vm.exists(dirForSnapshot(tag))`. With `0_1_5`
  frozen, bumping `[package].version` to `0.01.5` names a directory that does not
  exist, so the same release is frozen a second time under a second name and
  `SnapshotAlreadyFrozen` never fires.
- `recordPrecedes` reads both tags through `vm.parseUint`, so `0_01_5` and
  `0_1_5` compare EQUAL as versions and fall through to the byte-for-byte path
  tiebreak — ordered by the accident of a zero character rather than by release.
- Both are then declared as separate released suites and both are demanded live
  on every network by the chain group.

The version is also the soldeer `[package].version`, where `0.01.5` is not a
version any consumer can resolve.

The file states its own job as "the ONE definition of the release-version shape",
whose purpose is that two spellings cannot drift apart. Admitting two spellings
of one version is that same defect one step down, so the fix belongs in the
predicate rather than in a second check beside it.

## The fix

A component is one number, so it gets one spelling: refuse a leading zero on a
multi-digit component. The check fires on the second digit of a component whose
first digit was `0`, which is the only position a leading zero can occupy —
`digitsInComponent` resets at every separator, so each component is checked
independently, and `i` is at least 1 wherever `digitsInComponent == 1` because
that digit was read at an earlier index.

A single `0` stays legal. It IS the number zero, `0.0.0` is a real version, and
it is already pinned by `testIsTagAcceptsWhatAFreezeCanWrite`.

Because `isTag` is the same predicate asked with `_`, the directory side moves
with the version side — which is the property that made one predicate the right
place for this.

The doc comment stated the old rule ("three non-empty runs of digits") and now
states the current one.

## Tests

- `testIsStrictTripleRefusesLeadingZeros` — the padded spellings are refused at
  every component position (`0.01.5`, `01.1.5`, `0.1.05`, `00.0.0`), the tag form
  `0_01_5` with them, and the boundary is pinned on the accept side: `0.0.0`,
  `0.10.0`, `10.0.100`, `0_0_0`. Zeros inside and at the end of a component are
  what separate this from a rule that would refuse `0.10.0`.
- `testTagForVersionRefusesLeadingZeros` — the refusal is reachable through the
  release path, reverting `UnreleasableVersion` naming the version, not only
  through the predicate.

The existing fuzz `testEveryFreezableVersionIsATagTheRecordFinds` keeps passing:
`vm.toString` of a `uint256` never emits a leading zero.

Nothing in `src/generated/` is affected — it holds only `candidate/`, and no
existing tag has a padded component, so no directory that the record counts today
stops being counted.

## QA
- Discriminating tests: `testIsStrictTripleRefusesLeadingZeros`, `testTagForVersionRefusesLeadingZeros` — each fails on base (verified by reverting the guard to the pre-fix source, mutant 1 below: predicate test `FAIL: assertion failed`, tagFor test `FAIL: next call did not revert as expected`; both `PASS` unmutated, 22 tests observed to run in every case)
- Mutations applied: `src/lib/LibRainDeploySnapshot.sol:125` → (1) guard deleted entirely, i.e. the exact pre-fix code → killed by BOTH new tests; (2) `digitsInComponent == 1` → `digitsInComponent >= 1` (rejects a zero at any position, so `10.0.100` is refused) → killed by `testIsStrictTripleRefusesLeadingZeros`; (3) `subjectBytes[i - 1] == "0"` → `char == "0"` (inspects the current digit instead of the component's first, so `0.10.0` is refused while `0.01.5` is accepted) → killed by BOTH new tests. No surviving mutants.
- Oracle: a version component is a NUMBER, so `01` and `1` denote the same release and a release must have one spelling — derived from semver and from the repo's own use of the value (`recordPrecedes` reads tags through `vm.parseUint`, which maps both spellings to the same integer; `[package].version` is the soldeer version, where `0.01.5` resolves for nobody). The accept/reject set is fixed by that rule, not by reading the predicate: zeros INSIDE and at the END of a component (`0.10.0`, `10.0.100`) are the same number either way and must stay accepted, which is what separates this from a rule that refuses any zero.
- Category check: issue asks (a) refuse a leading zero on a multi-digit component while keeping a lone `0` legal, (b) pin the predicate directly, (c) pin the refusal as reachable through the release path naming the version; covered (a) by the `:125` guard, (b) by `testIsStrictTripleRefusesLeadingZeros` (all three component positions, plus `00.0.0` and the tag form `0_01_5`, plus the accept-side boundary `0.0.0`/`0.10.0`/`10.0.100`/`0_0_0`), (c) by `testTagForVersionRefusesLeadingZeros` asserting `UnreleasableVersion` carries the offending version. Full suite green with fork RPCs configured: 217 passed, 0 failed, 17 suites.

### DIFF
--- a/src/lib/LibRainDeploySnapshot.sol
+++ b/src/lib/LibRainDeploySnapshot.sol
@@ -87,8 +87,8 @@ library LibRainDeploySnapshot {
         return tagForVersion(vm.parseTomlString(vm.readFile("foundry.toml"), ".package.version"));
     }
 
-    /// Whether `subject` is three non-empty runs of digits joined by exactly
-    /// two `separator`s.
+    /// Whether `subject` is three numbers joined by exactly two `separator`s,
+    /// each written without a leading zero.
     ///
     /// The ONE definition of the release-version shape. It is asked with `.`
     /// for a version out of `foundry.toml` and with `_` for the directory that
@@ -115,6 +115,16 @@ library LibRainDeploySnapshot {
                 separators++;
                 digitsInComponent = 0;
             } else if (char >= "0" && char <= "9") {
+                // A component is one number, so it has one spelling. `01` and
+                // `1` are the same release and would freeze to two directories,
+                // neither of which `SnapshotAlreadyFrozen` sees as the other,
+                // and which `recordPrecedes` cannot order because they compare
+                // equal as versions. `i` is at least 1 wherever
+                // `digitsInComponent == 1`, because that digit was read at an
+                // earlier index.
+                if (digitsInComponent == 1 && subjectBytes[i - 1] == "0") {
+                    return false;
+                }
                 digitsInComponent++;
             } else {
                 return false;
--- a/test/src/lib/LibRainDeploySnapshot.t.sol
+++ b/test/src/lib/LibRainDeploySnapshot.t.sol
@@ -99,6 +99,25 @@ contract LibRainDeploySnapshotTest is Test {
         assertFalse(LibRainDeploySnapshot.isTag("collision-guard"));
     }
 
+    /// A component has ONE spelling. A padded component is the same release as
+    /// its unpadded form: it freezes to a second directory the immutability
+    /// check does not recognise as the first, and the two compare equal as
+    /// versions so nothing can order them either.
+    function testIsStrictTripleRefusesLeadingZeros() external pure {
+        assertFalse(LibRainDeploySnapshot.isStrictTriple("0.01.5", "."));
+        assertFalse(LibRainDeploySnapshot.isStrictTriple("01.1.5", "."));
+        assertFalse(LibRainDeploySnapshot.isStrictTriple("0.1.05", "."));
+        assertFalse(LibRainDeploySnapshot.isStrictTriple("00.0.0", "."));
+        assertFalse(LibRainDeploySnapshot.isTag("0_01_5"));
+
+        // The boundary: a single zero IS the number zero, and `0.0.0` is a real
+        // version.
+        assertTrue(LibRainDeploySnapshot.isStrictTriple("0.0.0", "."));
+        assertTrue(LibRainDeploySnapshot.isStrictTriple("0.10.0", "."));
+        assertTrue(LibRainDeploySnapshot.isStrictTriple("10.0.100", "."));
+        assertTrue(LibRainDeploySnapshot.isTag("0_0_0"));
+    }
+
     /// EVERY version a freeze accepts MUST produce a directory the record
     /// recognises as a release. A version that could be frozen to a directory
     /// the record then ignores is exactly the orphan snapshot
@@ -179,6 +198,16 @@ contract LibRainDeploySnapshotTest is Test {
     /// to a directory the append-only gate ignores forever.
     function testTagForVersionRefusesNonStrict() external {
         string[9] memory bad = ["0.1.7-rc1", "0.1", "0.1.7.1", "", "a.b.c", ".1.7", "0..7", "0.1.", "0.1.7 "];
+        for (uint256 i = 0; i < bad.length; i++) {
+            vm.expectRevert(abi.encodeWithSelector(UnreleasableVersion.selector, bad[i]));
+            this.externalTagForVersion(bad[i]);
+        }
+    }
+
+    /// The refusal MUST be reachable through the release path, naming the
+    /// version, rather than only through the predicate.
+    function testTagForVersionRefusesLeadingZeros() external {
+        string[3] memory bad = ["0.01.5", "01.1.5", "0.1.05"];
         for (uint256 i = 0; i < bad.length; i++) {
             vm.expectRevert(abi.encodeWithSelector(UnreleasableVersion.selector, bad[i]));
             this.externalTagForVersion(bad[i]);


thedavidmeister:
@coderabbitai review

thedavidmeister:
Reviewed 4e318a870f160a86e1354be3e09374dc4a0d6cba: ready — Closes #51. Read the guard, not just the tests.

The rejection fires only when `digitsInComponent == 1` and the preceding byte was `0`, so a lone `0` is still the number zero and passes, `01` is refused on its second digit, and `10`/`100` are unaffected because the preceding byte is not `0` and the guard is skipped once the component is two digits long. `0.0.0` passing and `00.0.0` failing are both pinned. The `i - 1` read cannot underflow: reaching `digitsInComponent == 1` means a digit was already consumed at a lower index.

Worth landing before the next tag specifically. `freeze` refuses a re-cut with `SnapshotAlreadyFrozen`, and a padded component would freeze to a second directory that neither `SnapshotAlreadyFrozen` recognises as the first nor `recordPrecedes` can order, since they compare equal as versions. This closes that off ahead of `sol-v0.1.6`.

PR #122, opened 2026-08-15
test: cover every answer length the resolved-address read guard refuses

### LINKED ISSUE #62: A read answering more than one word is never exercised, so the word-length guard is only half covered
**Audit finding `cov-09`** — dimension 2, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:291-293`

## Problem

`checkResolvedAddresses` refuses a read whose `returnData.length != 0x20`. Two of the three cases are covered — `testCheckResolvedAddressesUnreadableTargetReverts` gives `length == 0` and `testCheckResolvedAddressesDirtyWordReverts` gives exactly one dirty word — but nothing gives a read that answers with MORE than one word, which is what a consumer-supplied read of a function returning `(address,address)`, a `string`, or any dynamic type produces.

The gap is mutation-visible: weakening the guard to `returnData.length < 0x20` survives the entire existing suite, and the multi-word answer would then be `abi.decode`d down to its first word and compared as if the read had answered one address — the exact case the `ResolvedAddressReadFailed` docstring says it exists for ("answers with something that is not a single address-sized word").

## Proposed fix

Add to `test/src/lib/LibRainDeploy.t.sol`. No new fixture is needed — the read is mocked on a target that really has code, so the answer's LENGTH is the only thing wrong with it:

```solidity
    /// A read that answers with MORE than one word has not answered with an
    /// address either, and MUST be reported as `ResolvedAddressReadFailed`.
    /// This is a consumer that pointed a read at a function returning two
    /// values, or a dynamic type — and the answer decodes perfectly well as a
    /// word, so nothing but the length says it is not an address.
    ///
    /// The first word is the address the caller expects, so a guard that
    /// accepted any answer of at least one word would PASS this against a read
    /// that never gave a single address.
    function testCheckResolvedAddressesMultiWordAnswerReverts(bytes32 name, address account, address extra)
        external
    {
        vm.assume(account != address(0));
        (, MockResolvedOwner consumer) = deployRegistryAndConsumer(name, account);

        bytes memory twoWords = abi.encode(account, extra);
        assertEq(twoWords.length, 0x40);
        vm.mockCall(address(consumer), abi.encodeWithSignature("iOwner()"), twoWords);

        vm.expectRevert(
            abi.encodeWithSelector(
                LibRainDeploy.ResolvedAddressReadFailed.selector,
                "test_network",
                address(consumer),
                uint256(0),
                twoWords
            )
        );
        this.externalCheckResolvedAddresses("test_network", address(consumer), ownerReadCalls(), expected(account));
    }

    /// An answer SHORTER than a word but not empty is refused too — a read that
    /// returned twenty raw bytes, as the Zoltu factory itself does, is not an
    /// ABI-encoded address.
    function testCheckResolvedAddressesShortAnswerReverts(bytes32 name, address account) external {
        vm.assume(account != address(0));
        (, MockResolvedOwner consumer) = deployRegistryAndConsumer(name, account);

        bytes memory raw = abi.encodePacked(bytes20(account));
        assertEq(raw.length, 20);
        vm.mockCall(address(consumer), abi.encodeWithSignature("iOwner()"), raw);

        vm.expectRevert(
            abi.encodeWithSelector(
                LibRainDeploy.ResolvedAddressReadFailed.selector,
                "test_network",
                address(consumer),
                uint256(0),
                raw
            )
        );
        this.externalCheckResolvedAddresses("test_network", address(consumer), ownerReadCalls(), expected(account));
    }
```

---

<details>

Confirmed against source and by executing the mutation.

Source at src/lib/LibRainDeploy.sol:291 is exactly `if (!success || returnData.length != 0x20)`, followed by `abi.decode(returnData, (uint256))` and a uint160 range check — so the length guard is the only thing standing between a multi-word answer and it being truncated to its first word and compared as an address.

Grepped every checkResolvedAddresses test in test/src/lib/LibRainDeploy.t.sol. All reads are `iOwner()` (one address word) or produce an empty return (no-code target, reverting selector). Length regimes actually exercised: 0 and exactly 0x20. No test produces returnData.length > 0x20, and none produces 0 < length < 0x20 either.

Mutation proof, not inference: I edited the guard to `returnData.length < 0x20` and ran `forge test --match-path test/src/lib/LibRainDeploy.t.sol --match-test CheckResolvedAddresses`. Result: 10 passed, 2 failed — and both failures are `vm.createSelectFork: environment variable ARBITRUM_RPC_URL / BASE_RPC_URL not found`, i.e. missing local .env RPC endpoints on the two fork-based OnNetworks tests, not mutation kills. Every non-fork test passed under the weakened guard, so the mutant survives. Source restored to `!= 0x20` afterwards and verified.

None of the refutation grounds apply: the behaviour exists as described, the cited lines say what the finding claims, no sibling test or documented convention covers the multi-word case, and this is not a restatement of an org convention. The docstring at lines 288-290 explicitly states the error's subject is a read that "answers with something that is not a single address-sized word", so the uncovered case is one the code intends to handle.

Severity LOW is correct: the guard as written is currently CORRECT, so there is no live production defect — the value at risk is only the regression risk of an unpinned guard in a deploy-verification helper that runs in scripts, not on-chain. Keeping LOW.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/62 — audit finding `cov-09`, dimension 2, severity LOW.

## What was uncovered

`checkResolvedAddresses` refuses a read whose answer is not exactly one word, at `src/lib/LibRainDeploy.sol:291`:

```solidity
if (!success || returnData.length != 0x20) {
    revert ResolvedAddressReadFailed(network, target, i, returnData);
}
```

Only two answer lengths were ever produced by the suite — `0`, from a target with no code and from a reverting selector, and exactly `0x20`, from a real address getter. Nothing produced an answer LONGER than a word, which is what a consumer-supplied read of a getter returning `(address, address)` or any dynamic type answers with, and nothing produced one shorter but non-empty either.

## What is covered now

Two fuzz tests in `test/src/lib/LibRainDeploy.t.sol`, written as the RULE rather than as two point cases:

- `testCheckResolvedAddressesLongAnswerReverts` fuzzes the whole tail past the first word, so an answer of `0x21` bytes is as much an instance as one of `0x40`. The answer's leading word is exactly the address the caller expects, so nothing but the length is wrong with it — a guard that accepted any answer of at least one word would decode that word and PASS against a read that never gave a single address.
- `testCheckResolvedAddressesShortAnswerReverts` fuzzes every length between an empty answer and a whole word. Length 20 is exactly the shape the Zoltu factory itself answers with.

`test/concrete/MockRawAnswerOwner.sol` is what makes those answers reachable from a contract rather than from a cheatcode, alongside the existing `MockDirtyWordOwner`. It returns in assembly because the length is the point: every Solidity return type ABI-encodes to a whole number of words, so nothing declared can answer with twenty bytes.

No source change. The guard is correct as written; this pins it.

## QA
- Discriminating tests: `testCheckResolvedAddressesLongAnswerReverts`, `testCheckResolvedAddressesShortAnswerReverts` — neither fails on base, and neither can: the guard is CORRECT on `main`, which is why issue 62 is a coverage finding at severity LOW with no live defect. What each discriminates against is a specific weakening of the guard that the pre-existing suite does not catch, verified by running the guard mutated and restored under a script that printed the guard source line and `git status` before every run (final `git status` clean, guard line back to `!= 0x20`), so no result can come from a stale or unmutated build. For each mutant below I ran the ten pre-existing `CheckResolvedAddresses` tests separately from the two new ones, so "which test kills it" is an observation and not an inference.
- Mutations applied: all at `src/lib/LibRainDeploy.sol:291`. (1) `returnData.length != 0x20` → `returnData.length < 0x20`, accepting every answer of at least one word: all 10 pre-existing tests PASS (mutant survives them — this is exactly the finding in issue 62), killed by `testCheckResolvedAddressesLongAnswerReverts` with `next call did not revert as expected`, because the long answer passed the weakened guard, was decoded down to its first word and compared as if it were the one address. (2) `returnData.length != 0x20` → `returnData.length == 0 || returnData.length > 0x20`, accepting exactly the `1..0x1f` regime and nothing else: all 10 pre-existing tests PASS, killed by `testCheckResolvedAddressesShortAnswerReverts` with `call reverted as expected, but without data` — the short answer reached `abi.decode`, which reverts with no data of its own, the bare undiagnosable revert `ResolvedAddressReadFailed` exists to replace. (3) Reported because it is the obvious mutant and it is NOT load-bearing: `→ returnData.length > 0x20` is killed by the new short test but ALSO by the pre-existing `testCheckResolvedAddressesUnreadableTargetReverts`, since a zero-length answer is `<= 0x20` too — so it proves nothing about the short regime, which is why mutant (2) was constructed to isolate it. Baseline `!= 0x20`: both new tests pass.
- Oracle: the ABI and the error's own docstring, not the implementation. An ABI-encoded `address` is exactly one 32-byte word, so every other answer length is by construction not an address; the docstring at `LibRainDeploy.sol:288-290` states the error's subject is a read that "answers with something that is not a single address-sized word". The addresses passed to the check are the fuzzer's own inputs, and the expected revert data is rebuilt from the answer bytes the test itself constructed — nothing is read back out of the code under test.
- Category check: issue 62 asks for the two uncovered length regimes — more than one word, and shorter than a word but non-empty — and gives one point case for each. Both covered, and each widened from the issue's example to its whole regime: the long test fuzzes any tail past the first word rather than only a `(address,address)` return, and the short test fuzzes every length `1..0x1f` rather than only a packed 20 bytes. Lengths `0` and `0x20` were already covered (`testCheckResolvedAddressesUnreadableTargetReverts` / `testCheckResolvedAddressesRevertingReadReverts`, and `testCheckResolvedAddressesMatch`), so all four length regimes are now complete.

## Other checks

- `forge fmt` — no changes.
- `reuse lint` — compliant, 72/72 files, including the new mock.
- `forge test` — 209 passed, 8 failed. All 8 failures are `vm.rollFork` / `vm.createSelectFork` errors against the local Base endpoint in `isStartBlock*` / `findDeployBlock*`, which need an archive node; the same 8 fail identically on clean `main` at `86f8d96` in this environment, so they are pre-existing and environmental, not from this change. CI binds reachable endpoints via rainix's rpc-preflight. Every `checkResolvedAddresses` test passes, including the two fork-based `OnNetworks` ones.

## Summary by CodeRabbit

* **Tests**
  * Added fuzz coverage for invalid owner-response lengths.
  * Verified that responses longer or shorter than the expected ABI format revert with the appropriate error.
  * Added test support for returning arbitrary raw response data.


thedavidmeister:
Reviewed cabd100ba9e604b393bc9c04c34a064023934ce3: ready — merging.

## Conflict resolution

This branch was `CONFLICTING` against `main`. Resolved by **merging `main` into the branch**, twice (never rebased), because `main` moved again while the first merge was being verified:

- `32050fb` — merge of `ad1e00c`. One conflict, in `test/src/lib/LibRainDeploy.t.sol`'s import block: this branch added `MockRawAnswerOwner` next to the duplicate `MockAddressRevertingFactory` import, and #123 deleted that duplicate on `main`. Resolved by taking **both** sides' intent — the duplicate stays deleted, the `MockRawAnswerOwner` import stays. No test was weakened, deleted or renamed to resolve it.
- `cabd100` — merge of `0c24e99` (28 further commits). Auto-merged with no conflict; re-verified in full rather than assumed, because the intervening commits rewrote 193 lines of the same test file and touched `src/lib/LibRainDeploy.sol`.

Semantic-conflict check on the merged tree, not just on the markers:

- `main`'s change to `checkResolvedAddresses` since the branch point is the new `NoResolvedAddressReads` guard (empty read set refused) plus a `forge-lint` disable comment on the `uint160` cast. Neither is reachable from these two tests — both hand in a one-entry `readCalls`, so the length guard they exercise is still the one they were written against.
- `main`'s additions to `LibRainDeploy.t.sol` are purely additive (`holdsName`, `testSupportedNetworksAreFullyConfigured`, two `deployToNetworks` dependency tests). No helper this PR relies on — `externalCheckResolvedAddresses`, `ownerReadCalls()`, `expected()` — changed name or signature.

## What was actually run on the merge commit

Whole suite, not just the changed tests:

- CI `rainix / test / test` on `cabd100`: **231 passed, 0 failed, 0 skipped** — https://github.com/rainlanguage/rain.deploy/actions/runs/31948995510/job/95169411926 — including `testCheckResolvedAddressesLongAnswerReverts` and `testCheckResolvedAddressesShortAnswerReverts`, 256 fuzz runs each.
- Locally, `nix develop -c forge test` on `cabd100`: 224 passed, 7 failed. All 7 failures are `vm.rollFork: ... error code 4444: pruned history unavailable` from the public Base endpoint in the local `.env` — a pruning non-archive node, not an assertion. They are the pre-existing `isStartBlock` / `findDeployBlock` fork tests, untouched by this PR, and all 7 pass in CI where the Base RPC is an archive endpoint.
- `nix develop -c forge fmt --check` on `cabd100`: clean.
- CI `rainix / legal / legal` on `cabd100`: pass.

## The one red, and why it is not this PR

`rainix / static / static` fails on `cabd100`, and it fails on `main` too, for the same reason and at the same moment: rainix now enforces a 4096-byte agent-context cap and this repo's `CLAUDE.md` is ~24KB.

- this PR: `ERROR: this repo loads 24198 bytes of agent context ... 20102 over the 4096-byte cap` — https://github.com/rainlanguage/rain.deploy/actions/runs/31948995510/job/95169411953
- `main` at `6282b56`, run started one minute later: `ERROR: this repo loads 24227 bytes ... 20131 over the 4096-byte cap` — https://github.com/rainlanguage/rain.deploy/actions/runs/31949029086/job/95169491672

`CLAUDE.md` is byte-identical between this head and `main` (24333 bytes on both), and this PR does not touch it. The red is inherited whole from `main` and needs its own work — cutting ~20KB of `CLAUDE.md` per rainlanguage/rainix#298 — not a docs rewrite smuggled into a test-coverage PR. Admin-merging past it rather than fixing it here.

## Review coverage

CodeRabbit's check is green but reports **`Review rate limited`** on this PR. The zero unresolved-thread count (`reviewThreads.totalCount` = 0 via GraphQL at `cabd100`) is therefore vacuous — there is no review behind it — not evidence that anything was reviewed and cleared. Stating that rather than citing it as coverage.

## The diff itself, against https://github.com/rainlanguage/rain.deploy/issues/62

Test-only: one new `test/concrete/MockRawAnswerOwner.sol` and two fuzz tests. No `src/` change, so nothing deployable moves and no snapshot, address or code hash is affected.

The issue asked for the `returnData.length != 0x20` guard's uncovered regimes. It is covered as the rule rather than as the issue's two examples: the long-answer test fuzzes the whole tail past the first word, so `0x21` bytes is as much an instance as `0x40`, and the short-answer test fuzzes every length in `1..0x1f` with the raw address bytes leading, so length 20 — the Zoltu factory's own answer shape — is an instance rather than a hardcoded case. The long answer's leading word is exactly the expected address, which is what makes it discriminating: a guard weakened to `>= 0x20` decodes that word and passes against a read that never answered with one address.

`MockRawAnswerOwner` returns from assembly because the length is the point — every Solidity return type ABI-encodes to a whole number of words, so no declared return type can answer with twenty bytes. `memory-safe` holds: it returns a slice of memory it read and writes nothing.

PR #123, opened 2026-08-15
test: drop duplicate MockAddressRevertingFactory import

### LINKED ISSUE #71: `MockAddressRevertingFactory` is imported twice in the same file
**Audit finding `CQ4-01`** — dimension 4, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`test/src/lib/LibRainDeploy.t.sol:9,12`

## Problem

The import block declares `import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";` at line 9 and again, verbatim, at line 12. solc tolerates the repeated import of an identical declaration, so it compiles and nothing fails — which is exactly why it survives: a redundant line in a hand-maintained import block is the shape a bad merge or a copy-paste leaves behind, and the next reader has to check whether the two lines differ before concluding they do not. No other file in the repo imports one symbol twice, so this is the single outlier against a repo-wide convention of one import per symbol.

## Proposed fix

Delete the duplicate at line 12.

Before (lines 9-15):
```solidity
import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";
import {MockResolvedOwner} from "../../concrete/MockResolvedOwner.sol";
import {MockDirtyWordOwner} from "../../concrete/MockDirtyWordOwner.sol";
import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";
import {MockDeployable} from "../../concrete/MockDeployable.sol";
import {MockDeployableV2} from "../../concrete/MockDeployableV2.sol";
import {MockReverter} from "../../concrete/MockReverter.sol";
```

After:
```solidity
import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";
import {MockResolvedOwner} from "../../concrete/MockResolvedOwner.sol";
import {MockDirtyWordOwner} from "../../concrete/MockDirtyWordOwner.sol";
import {MockDeployable} from "../../concrete/MockDeployable.sol";
import {MockDeployableV2} from "../../concrete/MockDeployableV2.sol";
import {MockReverter} from "../../concrete/MockReverter.sol";
```

---

<details>

Confirmed against source at 440e90b. test/src/lib/LibRainDeploy.t.sol lines 9 and 12 are byte-identical imports of MockAddressRevertingFactory from ../../concrete/MockAddressRevertingFactory.sol. Refutation attempts all failed: (1) the cited lines say exactly what the finding claims; (2) the symbol IS used (lines 402, 1144), so this is a redundant duplicate line, not a mischaracterized dead import — the proposed fix (delete line 12) is correct as written; (3) the repo-wide-outlier claim holds — a per-file duplicate-import scan over src/, test/, script/ returns this file as the only true hit, the three other hits being the literal string "import {" opening multi-line import blocks in RegistryDeploySuites.sol, RainDeployVerifyChain.t.sol and RainDeployVerifySnapshot.t.sol, which are false positives rather than duplicated symbols; (4) no existing guard or tooling handles it — `nix develop -c forge fmt --check` exits 0 on the tree as-is so the formatter neither flags nor removes it, slither.config.json excludes only the assembly detector, and solc silently tolerates a repeated identical declaration, so none of the three CI matrix tasks catch it; (5) CLAUDE.md documents no convention sanctioning it. Severity corrected LOW -> INFO: value-at-risk in production is zero. This is a test-only file, and the duplicate line has no runtime, compilation, or behavioural effect whatsoever — the finding's own text concedes "it compiles and nothing fails". The entire cost is a future reader briefly diffing two lines. Under severity-by-production-impact that is INFO, not LOW.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/71

`test/src/lib/LibRainDeploy.t.sol` imported `MockAddressRevertingFactory` from `../../concrete/MockAddressRevertingFactory.sol` at line 9 and again, byte-identically, at line 12. This deletes the line-12 copy. One line, nothing else in the diff.

The symbol is still used — `new MockAddressRevertingFactory()` at what are now lines 400 and 1142 — so this removes a redundant duplicate line, not a live import. solc tolerates a repeated identical declaration, so nothing about compilation, runtime or behaviour changes. The whole cost was a reader having to diff two lines to conclude they are the same, which is also why it survived: nothing fails, so nothing points at it.

## Category, not just the cited line

The issue names one file and one symbol. The rule behind it is one import per symbol per file, so I checked that property across the repo rather than only the two cited lines: over every `*.sol` tracked in git, flattening each import's braced symbol list (multi-line import blocks included, so a symbol split across lines is still counted) and looking for any symbol imported twice in the same file returns exactly one hit — this file, this symbol. There is nothing else in the category, so the one-line fix is the whole fix.

## Why nothing caught it

`nix develop -c forge fmt --check` exits 0 on the tree with the duplicate present, so the formatter neither flags nor removes it; `slither.config.json` is not configured for it; solc is silent by design. None of the three CI matrix tasks would have failed on this. This PR does not add a guard — that is scope beyond the issue — but the gap is worth stating rather than leaving implied.

## QA

- **Discriminating tests:** n/a, and deliberately. This diff deletes a redundant duplicate declaration; it has no reachable behaviour to assert, so a test asserting it would assert nothing. The compiler is the only oracle that can discriminate here, and it is exercised under "Mutations applied" below. Existing coverage is unchanged: `LibRainDeployTest` still compiles and still exercises `MockAddressRevertingFactory` at both call sites.
- **Mutations applied:** `test/src/lib/LibRainDeploy.t.sol` line 9 — delete the *surviving* import as well, so the file has zero imports of `MockAddressRevertingFactory` against two usages -> killed by `nix develop -c forge build`, which fails with `Error (7920): Identifier not found or not unique.` pointing at line 400. That is the mutation that matters for this change: it proves the line the PR keeps is load-bearing and the line the PR deletes is not, which is exactly the claim "this was a duplicate, not a dead import". The inverse direction is already established and is why the issue exists — restoring the duplicate leaves the build, the formatter and slither all green, so no mutant on the deleted line can be killed by anything.
- **Oracle:** solc symbol resolution, independent of this repo's own test assertions — the compiler decides whether `MockAddressRevertingFactory` resolves at lines 400 and 1142, and it is not consulted about which of two identical import lines provided it. The duplicate's own cited location came from the issue and was re-read against source before editing.
- **Category check:** issue asks for exactly one thing — delete the duplicate import at line 12. Covered, plus the repo-wide scan above confirming this is the only symbol imported twice in any file, so no sibling instance is left behind.

**Test runs.** `nix develop -c forge fmt --check` exits 0. Full `nix develop -c forge test` with the five documented public RPC endpoints in `.env`: 202 passed across 17 suites, 16 suites fully green. The 13 failures were confined to `LibRainDeploy.t.sol`, and every one of the 26 `[FAIL]` lines is an HTTP 429 public-RPC rate limit (`Public RPC Rate Limit Hit` / `over rate limit`) from the shared free endpoints — zero assertion failures, and no failure mentions the import, the mock, or anything this diff touches.

Re-running that one suite with `--threads 1` takes it to **53 passed / 1 failed of 54**, the single remaining failure again a 429 from `arb1.arbitrum.io`. Failures fall 13 -> 1 purely by lowering RPC concurrency, with no source change between the two runs, which is what identifies rate limiting rather than the diff as the cause. CI runs with real RPC secrets and is the verdict that counts.

### DIFF
--- a/test/src/lib/LibRainDeploy.t.sol
+++ b/test/src/lib/LibRainDeploy.t.sol
@@ -9,7 +9,6 @@ import {AddressRegistry, ADDRESS_REGISTRY_ROOT} from "../../../src/concrete/Addr
 import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";
 import {MockResolvedOwner} from "../../concrete/MockResolvedOwner.sol";
 import {MockDirtyWordOwner} from "../../concrete/MockDirtyWordOwner.sol";
-import {MockAddressRevertingFactory} from "../../concrete/MockAddressRevertingFactory.sol";
 import {MockDeployable} from "../../concrete/MockDeployable.sol";
 import {MockDeployableV2} from "../../concrete/MockDeployableV2.sol";
 import {MockReverter} from "../../concrete/MockReverter.sol";


thedavidmeister:
Reviewed 9c4eab2d59ee6a73858c8cc87dacef1cde821462: ready — Closes #71. Read the diff.

One line removed: `MockAddressRevertingFactory` was imported twice from the same path, and the second import at line 13 goes. The first, at line 10, stays, so every use in the file still resolves. Nothing else in the diff.

CI green (rainix static/legal/test), 0 unresolved threads — qualified, since the CodeRabbit check reports `Review rate limited`, which is a green check with no review behind it, so that zero is vacuous rather than clean. Base has moved under this PR since its checks ran (#92 and #125 merged into the same file's directory), and GitHub reports MERGEABLE/CLEAN against current main; a duplicate-import removal cannot conflict semantically with tests added elsewhere in the file.

PR #124, opened 2026-08-15
docs(LibRainDeploy): drop the return name from `supportedNetworks`' NatSpec

### LINKED ISSUE #50: `supportedNetworks` documents a return name its signature does not declare
Raised independently by the code-quality and documentation dimensions; the same NatSpec defect.

---

# Finding `LIBDEPLOY-09`

**Audit finding `LIBDEPLOY-09`** — dimension 4, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:235-245`

## Problem

The NatSpec is `/// @return networks The list of supported network names.` but the signature is `returns (string[] memory)` — unnamed. solc cannot check the name against anything, so `networks` is silently absorbed into the description text and the documented name is unverified. Every other documented return in the file is consistent: `isStartBlock` (`@return isStart` / `returns (bool isStart)`), `findDeployBlock`, `deployZoltu` and `deployToNetworks` all name theirs, and `zoltuAddress` documents an unnamed return with no name (`@return The address...`). This one is the only mismatch, and the body already builds a local called `networks` (238) and returns it explicitly (244), so the name exists everywhere except where it is checkable.

## Proposed fix

Name the return, which the body already assigns on the only path that reaches the end — so the assignment is locally provable and the explicit return goes away.

```solidity
    /// Returns the list of networks currently supported by Rain deployments.
    /// @return networks The list of supported network names.
    function supportedNetworks() internal pure returns (string[] memory networks) {
        networks = new string[](5);
        networks[0] = ARBITRUM_ONE;
        networks[1] = BASE;
        networks[2] = BASE_SEPOLIA;
        networks[3] = FLARE;
        networks[4] = POLYGON;
    }
```

`testSupportedNetworks` (test/src/lib/LibRainDeploy.t.sol:196-204) already pins the length and every element, so it covers the rewrite unchanged.

---

<details>

Confirmed at commit 440e90b. src/lib/LibRainDeploy.sol:236 reads `/// @return networks The list of supported network names.` while line 237 is `function supportedNetworks() internal pure returns (string[] memory)` — unnamed, so solc has no name to check the docstring against and `networks` is silently treated as part of the description. The comparison the finding draws also holds on inspection: `isStartBlock` (99 doc / 102 `bool isStart`), `findDeployBlock` (144 / 148 `uint256 deployBlock`), `deployZoltu` (209 / 210 `address deployedAddress`), and the two deploy-to-network functions (377/387, 469/479 `address deployedAddress`) all name their returns, and `zoltuAddress` (197/198) correctly documents an unnamed return with no name. `supportedNetworks` is the sole mismatch, and the body already declares a local `networks` (238) and returns it (244). No guard, test, or convention refutes it: the org convention is that named returns are fine when locally provable, and here the single straight-line body assigns `networks` on the only path to the function end, so the proposed named return is provable and the explicit `return` becomes redundant — the fix follows the convention rather than restating a deliberate deviation from it. test/src/lib/LibRainDeploy.t.sol:196-204 `testSupportedNetworks` pins length 5 and each element against the library constants, so it covers the rewrite unchanged (and does not itself refute the doc mismatch, which is not behavioural). Severity INFO stands: this is a NatSpec naming inconsistency with no effect on compiled behaviour and no production value at risk.

</details>

---

# Finding `D3-08`

**Audit finding `D3-08`** — dimension 3, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`src/lib/LibRainDeploy.sol:235-245`

## Problem

The NatSpec reads `@return networks The list of supported network names.` but the signature is `returns (string[] memory)` with no named return, so `networks` is consumed as part of the description rather than as a name and the generated docs read "networks The list of supported network names." The local variable is already called `networks` and is returned explicitly, so the documented name and the implementation disagree only in the signature.

## Proposed fix

src/lib/LibRainDeploy.sol lines 237-245, name the return so the doc matches:

```solidity
    function supportedNetworks() internal pure returns (string[] memory networks) {
        networks = new string[](5);
        networks[0] = ARBITRUM_ONE;
        networks[1] = BASE;
        networks[2] = BASE_SEPOLIA;
        networks[3] = FLARE;
        networks[4] = POLYGON;
    }
```

---

<details>

Verified against source at /home/gildlab/code/rain-deploy-audit/src/lib/LibRainDeploy.sol. Lines 235-245 read exactly as the finding describes: `/// @return networks The list of supported network names.` immediately above `function supportedNetworks() internal pure returns (string[] memory) {`, with a local `string[] memory networks` built and explicitly `return networks;`. The return parameter is unnamed, so `networks` in the tag is not a return-parameter name — solc folds it into the description text and generated docs read "networks The list of supported network names."

Could not refute it as a misread or as a deliberate convention. The same file demonstrates both correct forms, so this is an internal inconsistency, not a house style: `zoltuAddress` (lines 197-198) documents an unnamed return with no leading name (`/// @return The address the creation code deploys to.`), while `isStartBlock` (99-103), `findDeployBlock` (144-150), `deployZoltu` (209-210), and the functions at 377 and 469 all declare the named return that their `@return <name>` tag uses. `supportedNetworks` is the sole function that names a return in NatSpec without declaring it. The exception for "named returns that ARE locally provable" does not apply — that exception protects existing named returns from being flagged; here the complaint is the opposite direction (doc names a return the signature omits), and the proposed fix moves the function toward the file's dominant named-return form. Nothing in CLAUDE.md speaks to NatSpec or named returns either way, so no documented convention covers it.

Tests do not bear on it: `test/src/lib/LibRainDeploy.t.sol:194-211` covers the returned contents (exactly 5 networks, expected order, factory present on each), and callers at src/abstract/RainDeployBroadcast.sol:63, src/abstract/RainDeployVerifyChain.sol:119 and three test sites all consume the value positionally — no test can observe a NatSpec/signature name mismatch, so the absence of one neither confirms nor refutes.

Severity INFO is correct and I would not move it. Zero value at risk: the mismatch is invisible at every call site, cannot change behaviour, does not fail compilation (solc only enforces name agreement when the return parameter IS named), and the deployed bytecode is identical under the proposed fix. It is a documentation-cleanliness item only.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/50

## The defect

`supportedNetworks` carried `/// @return networks The list of supported network
names.` above `function supportedNetworks() internal pure returns (string[] memory)`
— an unnamed return. `networks` was therefore not a return-parameter name at
all: solc only checks an `@return <name>` against a declared name, so with none
declared it silently folded `networks` into the description text and the
generated docs read "networks The list of supported network names."

That is the whole finding. Issue 50 carries it twice — `LIBDEPLOY-09`
(dimension 4) and `D3-08` (dimension 3), independently raised, same defect,
severity INFO both times.

## The fix, and why this direction

There are exactly two ways to make the tag and the signature agree: add the name
to the signature, or remove it from the tag. **This repo does not name return
values, so the name comes out of the tag.**

```diff
     /// Returns the list of networks currently supported by Rain deployments.
-    /// @return networks The list of supported network names.
+    /// @return The list of supported network names.
     function supportedNetworks() internal pure returns (string[] memory) {
```

One comment line against `main`. The signature, the local
`string[] memory networks` and the explicit `return networks;` are all
untouched.

This is the form the same file already uses 39 lines up:
`zoltuAddress` (src/lib/LibRainDeploy.sol:210-211) documents its unnamed return
as `/// @return The address the creation code deploys to.` Issue 50's own text
names that as the correct handling of an unnamed return — it cites `zoltuAddress`
as the file's example of getting this right, not as the outlier.

**What an unnamed return buys, and what naming it would have cost.** solc warns
"unnamed return variable can remain unassigned" on a path that reaches the
closing brace without a `return`. A named return has no such warning: it silently
returns the type's zero value. That is not hypothetical in this repo —
https://github.com/rainlanguage/rain.deploy/pull/103 exists because
`suiteNames()`'s named return was assigned only inside a loop, so an empty input
returned `""` with nothing to flag it. Its remedy is a bare `names = "";` seed
whose own QA records the mutant as **SURVIVED**: a line no test can kill,
because the path it protects is unreachable from anywhere the tests can stand.
The org rule is that a named return must be *locally provable*; keeping the
return unnamed here means there is nothing to prove, and the compiler keeps
doing the proving on every future edit to this body.

Naming the return would also have deleted `return networks;` — the statement
that makes the function's exit value visible at the point of exit — in exchange
for nothing a caller can observe.

## Note on this branch's history

The first commit on this branch (`9fa7433`) applied the fix in the *other*
direction: it named the return and dropped the local and the explicit `return`.
That direction was rejected — we do not name return values — so `03b2a23`
reverts that code edit and fixes the docstring instead. The branch's net diff
against `main` is the single comment line shown above; the two commits are
retained rather than force-rewritten so the rejected direction stays on the
record.

## QA

- **Discriminating tests**: none, and none is possible. The change is one
  NatSpec comment. It produces byte-identical bytecode, is invisible at every
  call site (`src/abstract/RainDeployBroadcast.sol`,
  `src/abstract/RainDeployVerifyChain.sol` and the test sites all consume the
  returned array positionally), and no test can observe a comment. A test that
  could fail on base would mean this was a behaviour change, which it is not.
  Stating that plainly rather than inventing a test.
- **Mutations applied**: none, for the same reason — there is no line of code in
  this diff to mutate. The function body is unchanged from `main`, so its
  existing coverage (`testSupportedNetworks`, test/src/lib/LibRainDeploy.t.sol,
  pinning length 5 and every element against the library constants) applies to
  it unmodified and was neither weakened nor duplicated.
- **Oracle**: independent of this diff. (a) solc's own rule — an `@return <name>`
  is checked only against a *declared* return-parameter name, so an unnamed
  return makes the tag's first word description text, which is exactly what the
  issue reports. (b) The file's own correct instance of the unnamed form,
  `zoltuAddress` at 210-211, which is the shape this change copies. (c) The
  user's standing ruling that return values are not named in this org, which is
  what selects between the two otherwise-equivalent repairs.
- **Category check**: issue 50 asks for one thing — make the NatSpec and the
  signature agree — under two finding IDs with the same proposed fix. Both are
  closed. The issue's *proposed diff* is deliberately not applied, because it
  picks the named-return direction; the defect it identifies is fixed the other
  way. Every other `@return` in the file was re-read while doing this and each
  one already agrees with its signature (`isStartBlock` 112/113,
  `findDeployBlock` 157/159, `deployZoltu` 222/223, `deployToNetworks` 418/419,
  `deployAndBroadcast` 510/511 all name a declared return; `zoltuAddress`
  210/211 documents an unnamed one with no name), so no adjacent NatSpec was
  touched.

## Verification

- `main` merged in (not rebased); no conflicts.
- `nix develop -c forge fmt --check` — clean, exit 0.
- `nix develop -c forge test` with all five RPC endpoints set — **251 passed, 0
  failed, 0 skipped** across 18 suites, fork tests included. Exit 0.


thedavidmeister:
@coderabbitai review

thedavidmeister:
Reviewed 03b2a23f6f1a0d480cb6435ea51d1de83d6ad7fa: ready — the fix now points the
other way, and the net change against `main` is one comment line.

## What changed since the last head

`9fa7433` had named the return (`returns (string[] memory networks)`) and
deleted the local declaration and the explicit `return networks;`. That
direction is rejected: **this repo does not name return values.** `03b2a23`
reverts that code edit and repairs the docstring instead. The two commits are
kept rather than force-rewritten so the rejected direction stays on the record.

`git diff origin/main` at this head, in full:

```diff
--- a/src/lib/LibRainDeploy.sol
+++ b/src/lib/LibRainDeploy.sol
@@ -246,7 +246,7 @@ library LibRainDeploy {
     }

     /// Returns the list of networks currently supported by Rain deployments.
-    /// @return networks The list of supported network names.
+    /// @return The list of supported network names.
     function supportedNetworks() internal pure returns (string[] memory) {
         string[] memory networks = new string[](5);
         networks[0] = ARBITRUM_ONE;
```

1 file changed, 1 insertion(+), 1 deletion(-). The signature, the local
`string[] memory networks` and the explicit `return networks;` are byte-identical
to `main`.

## Why this closes https://github.com/rainlanguage/rain.deploy/issues/50 rather than dodging it

Both findings on that issue — `LIBDEPLOY-09` (dimension 4) and `D3-08`
(dimension 3) — report one thing: solc checks an `@return <name>` only against a
*declared* return-parameter name, so with the return unnamed, `networks` was
never a name at all and was folded into the description text. Removing the word
makes the tag agree with the signature exactly as surely as adding it to the
signature would have. The tag now reads as documentation of an unnamed return,
which is what the function has.

It is also the form the same file already uses, 39 lines up:
`src/lib/LibRainDeploy.sol:210-211`, `zoltuAddress`, documents its unnamed
return as `/// @return The address the creation code deploys to.` Issue 50's own
text cites `zoltuAddress` as the file's example of getting this right. The
issue's *proposed diff* picks the other repair; the defect it identifies is
fixed, the diff it suggests is not applied, and that is deliberate.

Checked the rest of the file's `@return` tags while here — `isStartBlock`
(112/113), `findDeployBlock` (157/159), `deployZoltu` (222/223),
`deployToNetworks` (418/419), `deployAndBroadcast` (510/511) each name a
declared return, and `zoltuAddress` (210/211) documents an unnamed one with no
name. No other mismatch, so no adjacent NatSpec touched. `grep` for
`@return networks` across `src/`, `test/`, `script/` and the markdown returns
nothing.

## The point of keeping it unnamed

solc emits *"Unnamed return variable can remain unassigned"* for a body that can
reach its closing brace without a `return`. A named return has no equivalent
diagnostic — it silently returns the type's zero value. That is not theoretical
here: https://github.com/rainlanguage/rain.deploy/pull/103 exists because
`suiteNames()`'s named return was assigned only inside a loop, so an empty input
returned `""` with nothing to flag it, and its remedy is a bare `names = "";`
seed that PR's own QA records as a **SURVIVED** mutant — a line no test can
kill, because the path it guards is unreachable from anywhere a test can stand.
Leaving `supportedNetworks` unnamed means there is nothing to prove, and the
compiler keeps proving it on every future edit to this body.

## QA

- **Discriminating tests**: none, and none is possible — stating that rather
  than inventing one. The diff is a single NatSpec comment. Bytecode is
  byte-identical, and every caller (`src/abstract/RainDeployBroadcast.sol`,
  `src/abstract/RainDeployVerifyChain.sol`, and the test sites) consumes the
  array positionally, so there is nothing a test could observe. A test that
  failed on base would mean this was a behaviour change, which it is not.
- **Mutations**: none applied, because this diff contains no line of code to
  mutate — the function body is unchanged from `main`. Its existing coverage,
  `testSupportedNetworks` (test/src/lib/LibRainDeploy.t.sol, pins length 5 and
  each element against the library constants) plus
  `testSupportedNetworksAreFullyConfigured`, applies unmodified and was neither
  weakened nor duplicated. Both are `[PASS]` in the run below.
- **Oracle**: independent of the diff. (a) solc's rule on when an `@return`
  name is checkable. (b) `zoltuAddress` at 210-211, the file's existing correct
  unnamed form. (c) the standing ruling that return values are not named, which
  is what selects between two otherwise-equivalent repairs.

## Verification at this head

- `origin/main` merged in, not rebased — no conflicts. Branch is 0 behind.
- `nix develop -c forge test` with all five RPC endpoints set — **251 passed, 0
  failed, 0 skipped** across 18 suites, fork tests included, exit 0.
  `testSupportedNetworks` (gas 7901) and `testSupportedNetworksAreFullyConfigured`
  (gas 41963) both `[PASS]`.
- `nix develop -c forge fmt --check` — clean, exit 0.
- `nix develop -c slither .` — `. analyzed (49 contracts with 100 detectors),
  0 result(s) found`, exit 0.
- `nix develop -c rainix-sol-single-contract` — exit 0.

The last three were run locally because CI could not reach them (see below).

## The one red, and why it is not this PR

`rainix / static / static` fails at
`rainlanguage/rainix/.github/actions/agent-context-cap@main`:

```
ERROR: this repo loads 24227 bytes of agent context at the start of every
session — 20131 over the 4096-byte cap.
    24227  CLAUDE.md
```

That step runs before slither, `forge fmt --check` and
`rainix-sol-single-contract`, so all three are **skipped**, not failed.

`main`'s own tip `369cfe9` fails on exactly the same step in exactly the same
job — verified against run 31949244742, where `static` is the only failing job
and `agent-context-cap` its only failing step. It is a repo-wide condition this
branch inherited through the merge, it is about `CLAUDE.md`'s size, and this
PR's entire diff is one comment line in a `.sol` file. Cutting 20kB from
`CLAUDE.md` is separate work, not this PR's scope.

All three steps that red masks were run locally instead, so nothing goes
unchecked: `forge fmt --check`, `slither .` (0 results) and
`rainix-sol-single-contract` all exit 0. `rainix / test / test` and
`rainix / legal / legal` are green on CI. No unresolved review threads
(checked via the GraphQL `reviewThreads` query, not the check status —
CodeRabbit reports `pass` here only because it was rate limited and produced no
review).

PR #125, opened 2026-08-15
test: assert every declared artifact path resolves to the contract it names

### LINKED ISSUE #82: artifactPath is a hand-written string nothing resolves, and half of it is load bearing for the shape test
**Audit finding `H6-06`** — dimension 6, severity INFO. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RegistryDeploySuites.sol:110-151`

## Problem

HAZARD: `artifactPath` is the one field of a suite that is neither derived nor checked against anything, yet its `:<Name>` half silently decides which snapshot the shape assertions attribute to which candidate — so a wrong path is both an unusable verification command and a name that is right only by accident.

SHAPE: `DeploySuite.artifactPath` is declared by hand (`"src/concrete/AddressRegistry.sol:AddressRegistry"` at line 118, `"src/concrete/MigrationRegistry.sol:MigrationRegistry"` at line 146) and RainDeploySuitesBase.sol:73-78 explains it is declared rather than derived because no naming convention recovers a grouped repo's paths — a good reason not to derive it, and not a reason to leave it unchecked. It is consumed in two places: `LibRainDeploy.deployToNetworks` prints it into the `forge verify-contract` command (src/lib/LibRainDeploy.sol:448-453), and `GeneratedSnapshotShapeTest.candidateContractName` splits it on `:` and uses the trailing token as the contract name that must match a file in `src/generated/candidate/` (test/src/lib/GeneratedSnapshotShape.t.sol:126-129). Nothing asserts the path half names a file that exists, or that the file declares that contract.

SCENARIO: `AddressRegistry.sol` is moved into a subdirectory as the repo grows a second family of concretes, and every reference the compiler can see is updated — imports, remappings, `type(X).creationCode` — while this string, which no compiler reads, is left at `src/concrete/AddressRegistry.sol:AddressRegistry`. All green: the contract name half still matches the snapshot, so the shape test is satisfied, and no other test touches the field. It next matters during a deploy dispatch, where the printed manual verification command names a file that does not exist, at the exact moment somebody is trying to verify a freshly broadcast contract on five explorers.

STRUCTURAL FIX: assert the declared path resolves to a file that declares the named contract, for every candidate, so the field is checked rather than trusted.

## Proposed fix

Add to `test/src/lib/GeneratedSnapshotShape.t.sol` (which already inherits `RegistryDeploySuites` and already derives the contract name from this field). `fs_permissions` already grants read-write on `./src`:

```solidity
/// PROPERTY: every declared artifact path resolves — `<path>:<Name>` where the
/// path is a file that exists and that file declares `<Name>`.
///
/// It is the one suite field nothing derives and nothing else checks, and it is
/// load bearing twice over: `LibRainDeploy` prints it as the explorer
/// verification command a human runs against a freshly broadcast contract, and
/// `candidateContractName` takes the contract this whole shape spec is about
/// out of it. A path left behind by a moved source file passes every other
/// assertion here, because the half that moved is the half nothing reads.
function testEveryCandidateArtifactPathResolves() external view {
    DeployCandidate[] memory candidates = checkedCandidateSuites();
        string memory declared = candidates[i].snapshot.artifactPath;
        string[] memory parts = vm.split(declared, ":");
        assertEq(parts.length, 2, string.concat("artifact path is not <path>:<Name>: ", declared));
        assertTrue(vm.exists(parts[0]), string.concat("artifact path names no such file: ", declared));
        assertTrue(
            vm.contains(vm.readFile(parts[0]), string.concat("contract ", parts[1], " ")),
            string.concat("artifact path's file declares no such contract: ", declared)
        );
    }
}
```

---

<details>

SURVIVES, downgraded LOW -> INFO.

Verified against source. The mechanism is exactly as described: artifactPath is a hand-written literal (RegistryDeploySuites.sol:118, :146), declared rather than derived per the documented reason at RainDeploySuitesBase.sol:73-78, and has exactly two consumers - RainDeployBroadcast.run() passes suite.artifactPath into deployAndBroadcast -> deployToNetworks, where LibRainDeploy.sol:447-453 prints it into a console2.log of the `forge verify-contract` command; and GeneratedSnapshotShape.t.sol:126-129 splits it on ":" and takes the trailing token as the contract name.

Grepped the full test tree for any resolution of the path half (vm.exists / vm.isFile / readFile over a declared artifact path): nothing. The only assertions on the field are equality against a hardcoded literal (RainDeployBroadcast.t.sol:136) and pass-through equality (RainDeploySuitesBase.t.sol:60) - neither resolves the path. So the path half is genuinely unchecked, and the drift scenario is clean: foundry.toml sets bytecode_hash = "none" and cbor_metadata = false, so moving a source file changes no bytecode, no address, no codehash, and every group 1-4 check plus the whole shape spec stays green.

One overreach in the finding, not fatal: "a name that is right only by accident" is false. testEveryCandidateHasASnapshot asserts set equality between the names derived from artifactPath and the names walked out of src/generated/candidate/, in BOTH directions plus a count assertion, so the :Name half is cross-checked against the generator (and independently against Build.sol's GeneratedContract.contractName, which places that file). The finding's own scenario concedes this. The surviving claim is narrower than the title: only the path half is untrusted.

Severity corrected to INFO on value-at-risk, not on depth. Three things I confirmed bound the exposure to near zero: (1) the field is inert to the deployment - it never affects creation code, the derived address, the codehash, dependency checks, or any verification assertion, so nothing on chain and nothing in the record can be wrong because of it; (2) foundry.toml's [etherscan] block records that rainix-manual-sol-artifacts passes --verify by default, so this printed string is a human fallback for an automated verification step that has already run, not the primary path; (3) the failure is loud and non-silent - `forge verify-contract --chain X <addr> <bad-path>:<Name>` errors with no matching artifact rather than verifying something else, and a mismatched path:Name cannot resolve to a different contract because two same-named contracts cannot both be the artifact. Total cost of the hazard firing is a human retyping a path during a deploy they are already supervising. LOW implies some production value at risk; there is none.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/82

`DeploySuite.artifactPath` was the one field of a suite that nothing derives and
nothing checked. Its `:<Name>` half was read — `testEveryCandidateHasASnapshot`
matches it against the generator's output — but the path half was read by no
compiler and no test, so a source file moved or renamed left the string behind
with every assertion in the repo still green. `bytecode_hash = "none"` and
`cbor_metadata = false` mean a move changes no creation code, no address and no
code hash either, so there was nothing else for it to be caught by. It next
mattered at a deploy dispatch, where `LibRainDeploy` prints it as the
`forge verify-contract` command a human runs against a freshly broadcast
contract.

## What lands

One test, `testEveryCandidateArtifactPathResolves`, in
`test/src/lib/GeneratedSnapshotShape.t.sol` — which already inherits
`RegistryDeploySuites` and already takes the contract name out of this field.
Three clauses over every checked candidate:

- the declared string splits on `:` into exactly two parts
- `vm.isFile` on the path half
- `keccak256(vm.getCode(declared))` equals `keccak256(sourceCreationCode)`

## Departures from the fix the issue proposed

The issue proposed `vm.exists(parts[0])` and a text search,
`vm.contains(vm.readFile(parts[0]), "contract <Name> ")`. Both changed.

**`vm.isFile` rather than `vm.exists`.** `exists` is also true of a directory,
and a directory named `AddressRegistry.sol` is not a file `forge verify-contract`
can read.

**`vm.getCode` rather than a text search.** A text search asserts formatting: it
is satisfied by the name occurring in a comment or inside a longer identifier,
and defeated by a declaration written with no space before its brace.
`vm.getCode` is foundry's own resolution of a `<path>:<Name>` artifact id — the
same form the printed command takes — so the test resolves the string the way
the thing it is about resolves it.

Comparing the resolved code against the candidate's own `sourceCreationCode`
also strengthens the property the issue stated. "Resolves" and "resolves to the
RIGHT contract" are different claims, and the gap between them is reachable:
two candidates with their declared paths swapped resolve perfectly, and the
swap is invisible to `testEveryCandidateHasASnapshot` because moving both halves
together leaves the SET of declared names exactly as it was. That case is
mutation M4 below, and the ablation shows this clause is the only thing in the
repo that catches it.

Hashed rather than compared as bytes, for the reason `CandidateSourceMismatch`
already gives for hashing: creation codes run to tens of kilobytes, and a
failure message carrying two of them is a message nobody reads.

## Verification

Every mutation is on the declared `artifactPath` in
`src/abstract/RegistryDeploySuites.sol`, run against the WHOLE suite, with the
failure message the run actually produced.

| # | mutation | result |
|---|---|---|
| M1 | path half stale — `src/concrete/registry/AddressRegistry.sol:AddressRegistry` (file moved deeper, name still right) | KILLED: `artifact path names no such file: src/concrete/registry/AddressRegistry.sol:AddressRegistry` |
| M2 | bare-suffix path — `AddressRegistry.sol:AddressRegistry` | KILLED: `vm.isFile: the path AddressRegistry.sol is not allowed to be accessed for read operations` |
| M3 | no colon — `src/concrete/AddressRegistry.sol` | KILLED: `artifact path is not <path>:<Name>: src/concrete/AddressRegistry.sol: 1 != 2` |
| M4 | the two candidates' paths swapped | KILLED: `artifact path resolves to another contract: src/concrete/MigrationRegistry.sol:MigrationRegistry` |
| M5 | real file, declares no such contract — `src/concrete/MigrationRegistry.sol:AddressRegistry` | KILLED: `vm.getCode: no matching artifact found` |
| M6 | unique proper-suffix path — `concrete/AddressRegistry.sol:AddressRegistry` | KILLED: `vm.isFile: the path concrete/AddressRegistry.sol is not allowed to be accessed for read operations` |

M1, M2, M4, M5 and M6 each take the count from 154 passed / 0 failed to
153 / 1 — one new failure, this test, and nothing else moves. M3 takes it to
152 / 2, because a path with no colon also makes `candidateContractName` return
the whole path, which `testEveryCandidateHasASnapshot` fails on independently.

Counts are over the 14 non-fork suites (154 tests) on the merge commit: the
three fork suites inherit no `RegistryDeploySuites`, so none of them can observe
a mutation to it, while public-RPC rate limiting makes their pass/fail
non-deterministic. Unmutated, the whole 218-test suite is green on the merge
commit, and it is 217 / 1 under M1 and under M4 on runs that came back free of
rate-limited fork tests. An earlier revision of this section cited a
169 passed / 47 failed baseline whose 47 failures were fork tests with no
`*_RPC_URL` bound — the harness declining to run them rather than results — so
that figure is not one to re-derive.

### Each clause earns its place

Ablation: disable one clause, apply the mutation it is supposed to be the one
catching, and see whether anything else notices.

| # | ablation | result |
|---|---|---|
| A1 | `isFile` disabled + M2 | still KILLED — `vm.getCode: multiple matching artifacts found`. `AddressRegistry.sol` is a suffix of BOTH `src/concrete/` and `src/generated/candidate/`, so this one is ambiguous rather than unresolvable. M2 does not isolate `isFile`; M6 does. |
| A2 | `getCode` equality disabled + M4 | SURVIVED — the whole suite goes green on two candidates naming each other's contracts. This clause is the only thing anywhere that catches it. |
| A3 | shape assert disabled + M3 | SURVIVED in this test — `getCode` resolves a colon-less `.sol` path fine. The mutation is still caught by `testEveryCandidateHasASnapshot`, but this clause is what makes THIS test name the actual defect. |
| A4 | `isFile` disabled + M6 | SURVIVED — whole suite green on a path that resolves uniquely to the right contract and names no file at the repo root. This is what `isFile` is for. |
| A5 | `isFile` disabled + M1 | still KILLED — `vm.getCode: no matching artifact found`. A file moved DEEPER leaves a stale path that is not a suffix of the new one, so `getCode` catches that one on its own; `isFile` only gives it the better message. |

A1 and A5 are the two cases where I expected `isFile` to be the sole catcher and
it was not — the ablation is in the PR because it corrected the reasoning, not
only because it confirmed it. The doc comment on the test now cites M6/A4, which
is the case that was actually verified, rather than the bare-suffix example I
first wrote down and A1 refuted.

Failure messages for M2, M6 and any path outside the `fs_permissions` roots come
out as a cheatcode revert rather than as the assertion's own string, because
`vm.isFile` refuses a path it is not permitted to read instead of returning
false. Still red, still names the offending path. Not papered over with a
try/catch that would report "no such file" for a file that exists and is
unreadable.

## QA

- Discriminating tests: `testEveryCandidateArtifactPathResolves` — PASSES on
  unmutated base, because the two declared paths are currently correct; this is
  a guard on a field that was unchecked, not a fix for a live wrong value. Its
  discrimination is the six mutations below, each of which it fails on, and the
  ablation above, which shows each of its three clauses is the sole catcher of
  at least one of them. Verified by running the suite per mutation and reading
  the per-test `[FAIL: <message>]` lines, not exit codes.
- Mutations applied: all on `src/abstract/RegistryDeploySuites.sol`
  `artifactPath` — M1 `src/concrete/AddressRegistry.sol` ->
  `src/concrete/registry/AddressRegistry.sol` -> killed by the `isFile` clause;
  M2 -> `AddressRegistry.sol:AddressRegistry` -> killed (`isFile`, and by
  `getCode` ambiguity once `isFile` is ablated); M3 -> drop the `:AddressRegistry`
  half -> killed by the `<path>:<Name>` clause (and independently by
  `testEveryCandidateHasASnapshot`); M4 -> the two candidates' paths swapped ->
  killed by the `getCode`/`sourceCreationCode` clause ONLY, whole suite green
  without it; M5 -> `src/concrete/MigrationRegistry.sol:AddressRegistry` ->
  killed by `getCode` (`no matching artifact found`); M6 ->
  `concrete/AddressRegistry.sol:AddressRegistry` -> killed by the `isFile`
  clause ONLY, whole suite green without it.
- Oracle: independent of the string under test in all three clauses — the
  filesystem (`vm.isFile`), foundry's own artifact resolution (`vm.getCode`,
  the same `<path>:<Name>` resolution `forge verify-contract` performs), and
  the candidate's `sourceCreationCode`, which is `type(X).creationCode` taken
  from the compiler rather than from the declaration. Nothing here is derived
  from `artifactPath` itself.
- Category check: the issue asks for one thing — assert the declared path
  resolves to a file that declares the named contract, for every candidate.
  Covered for every candidate via `checkedCandidateSuites()`, which reverts
  `NoDeployCandidates` rather than passing a vacuous loop. Two departures from
  the proposed implementation (`isFile` over `exists`, artifact resolution over
  a `contract <Name> ` text search) are argued above; both are strictly
  stronger, neither narrows the property.

## One thing this does not cover

The check is repo-local. `artifactPath` is a hand-written literal in every deploy
repo that inherits `RainDeploySuitesBase`, and the verification abstracts in
`src/` are what consumers inherit instead of hand-writing checks — but `test/` is
`.soldeerignore`d, so a downstream deploy repo keeps the hazard this closes here.
Moving the property into `RainDeployVerifySnapshot` would fix that for everyone
and would also impose a new `fs_permissions` requirement on every consumer, so it
is a design call rather than a detail of this issue. Happy to file it separately.


thedavidmeister:
Reviewed e2b0200d4166b85106a56156bfbb44375c7a8f89: ready — Closes #82. Read the diff independently against the issue.

Test-only. `testEveryCandidateArtifactPathResolves` asserts three clauses over every checked candidate: the declared string splits on `:` into two parts, `vm.isFile` on the path half, and `keccak256(vm.getCode(declared)) == keccak256(sourceCreationCode)`.

Both departures from the issue's proposed patch are improvements and are argued in place. `vm.isFile` over `vm.exists` — `exists` is true of a directory, which is not a file `forge verify-contract` can read. `vm.getCode` over a `vm.contains(readFile, "contract <Name> ")` text search — the search asserts formatting, passes on a mention in a comment or inside a longer identifier, and fails on a declaration with no space before its brace; `vm.getCode` resolves the id the same way the thing under test resolves it.

The third clause earns its place: resolving and resolving to the RIGHT contract are different claims, and two candidates with their paths swapped resolve perfectly while leaving the SET of declared names unchanged, so `testEveryCandidateHasASnapshot` stays green through it. That is mutation M4 and this is the only assertion in the repo that catches it. Keeping `vm.isFile` alongside `vm.getCode` is also right rather than redundant — `getCode` resolves by path SUFFIX, so `concrete/AddressRegistry.sol:AddressRegistry` resolves uniquely and correctly while naming no file anyone can point the printed command at.

CI green (rainix static/legal/test), 0 unresolved threads. One qualification on that: the CodeRabbit check reports `Review rate limited`, which is a green check with no review behind it, so the zero-threads count is vacuous rather than clean — this rests on my read and the mutation matrix, not on CodeRabbit.

PR #126, opened 2026-08-15
Generate the released-suites aggregate instead of writing it by hand

### LINKED ISSUE #83: releasedSuites() is hand-written glue over per-contract generated libs, and the doc beside it says adding a contract touches nothing else
**Audit finding `HZ6-07`** — dimension 6, severity LOW. Whole-repo audit pass 1 at `440e90b5`.

`src/abstract/RegistryDeploySuites.sol:64-85`

## Problem

HAZARD: `script/Build.sol` emits one `Lib<Contract>Released.sol` per generated contract, but the concatenation of them is written by hand (src/abstract/RegistryDeploySuites.sol:64-75). Adding a deployed contract therefore touches four places — a named candidate function, `candidateSuites()`, `Build.generatedContracts()` and this concat — while the `@dev` immediately below says three: "a third named candidate below, a third entry here, and a third entry in `script/Build.sol`'s generated-contract list — nothing else" (src/abstract/RegistryDeploySuites.sol:78-80). CLAUDE.md repeats the same claim.

SHAPE: three of the four places are self-enforcing — a candidate with no snapshot fails `GeneratedSnapshotShapeTest.testEveryCandidateHasASnapshot`, and a snapshot with no candidate fails it too, and a missing generated file fails to compile. The fourth is not: a `Lib<Contract>Released` that exists but is never concatenated compiles cleanly, is read by nothing, and leaves the whole suite green.

PRODUCTION SCENARIO: a third registry is added following the doc's list of three. Every test passes for as long as nothing is released. The `sol-v*` tag then runs `cutRelease()`, which freezes all three contracts into `src/generated/<tag>/` from `generatedContracts()`, after which `testEveryFrozenSnapshotIsReleased` reverts `FrozenSnapshotNotReleased` for the third file — the release job fails at the one moment (tag pushed, deploy already broadcast) when the operator least wants to be editing a declaration, and the diagnosis points at the frozen record rather than at the concat that omitted it.

STRUCTURAL FIX: generate the concatenation from the same one list everything else reads, so there is no fourth place.

## Proposed fix

1) Emit the aggregate from `src/lib/LibRainDeploySnapshot.sol`, beside `writeReleasedSuitesLib`:

```solidity
    /// Generate the ONE released-suites lib a repo's declaration reads: the
    /// per-contract libs concatenated, emitted from the same list that wrote
    /// them. The concatenation is the last place a contract could be generated,
    /// aliased and frozen and still be absent from the declaration, and a
    /// declaration that misses a release is a release every check quietly stops
    /// asking about.
    /// @param vm The Vm instance for file operations.
    /// @param contractNames The contracts whose released libs to aggregate.
    /// @return The path written.
    function writeReleasedSuitesAggregate(Vm vm, string[] memory contractNames) internal returns (string memory) {
        // Emits `src/lib/LibReleasedSuites.sol` — imports every
        // `Lib<Contract>Released`, sums their lengths and copies them in
        // declaration order.
    }
```

2) Call it from `script/Build.sol`'s `regenerateLibs()`, which already holds `generatedContracts()`, so both entry points get it.

3) Reduce `src/abstract/RegistryDeploySuites.sol:64-75` to the generated call:

```solidity
    function releasedSuites() internal pure override returns (DeploySuite[] memory) {
        return LibReleasedSuites.releasedSuites();
    }
```

4) Correct the `@dev` at src/abstract/RegistryDeploySuites.sol:78-80 and the matching CLAUDE.md sentence to state what is actually true after the change: a third contract is a third named candidate, a third entry in `candidateSuites()` and a third entry in `script/Build.sol` — the released declaration then follows from the generator.

Coverage: `test/src/lib/LibRainDeploySnapshot.t.sol` already builds a record tree of its own under `test/generated`; add cases there for the aggregate writer — zero contracts emits a lib returning an empty array (the pre-first-release state, which must compile and be readable); one contract with no releases emits an empty array; two contracts with one release each emits both entries in declaration order; one contract with two releases emits both in tag order; and the emitted source, once compiled, returns entries whose keys are unique, since `allSuites()` reverts `DuplicateDeploySuite` on a collision and the aggregate is the one place two contracts' entries meet.

---

<details>

Tried to refute; the code and the citations hold.

VERIFIED AS DESCRIBED:
- `/home/gildlab/code/rain-deploy-audit/src/abstract/RegistryDeploySuites.sol:64-75` is a hand-written concatenation of `LibAddressRegistryReleased.releasedSuites()` and `LibMigrationRegistryReleased.releasedSuites()`. Lines 78-80 do literally read "A third deployed contract is a third named candidate below, a third entry here, and a third entry in `script/Build.sol`'s generated-contract list — nothing else." Adding a contract also requires two imports and a third branch in the concat above, so "nothing else" is false as written. Line citations are exact.
- `script/Build.sol:127-137` (`regenerateLibs`) emits one `Lib<Contract>Released.sol` per entry of `generatedContracts()`; nothing emits or checks the aggregate. No `LibReleasedSuites.sol` exists.
- Grepped the whole test tree for the claimed missing enforcement. `GeneratedSnapshotShapeTest.testEveryCandidateHasASnapshot` walks `src/generated/candidate/` and matches it both ways against `checkedCandidateSuites()` — that covers snapshots, not the released libs. `LibRainDeploySnapshot.t.sol:testWriteReleasedSuitesLibWritesTheLibAtItsPath` re-runs the real writer against the real record but only for `AddressRegistry` (`EMITTED_CONTRACT`), asserting the committed file's CONTENT; nothing asserts the SET of released libs, and nothing reads the concat. So a generated-but-unconcatenated `Lib<X>Released` is read by nothing and the suite stays green — confirmed, not merely asserted.

WHY IT STILL ONLY RATES LOW (severity confirmed, not raised):
The material failure is caught, just late and loudly. `RainDeployVerifySnapshot.testEveryFrozenSnapshotIsReleased` (src/abstract/RainDeployVerifySnapshot.sol:263) matches every file under `src/generated/<tag>/` by derived address against `releasedSuites()` and reverts `FrozenSnapshotNotReleased(path)` for the omitted contract. Per `.github/workflows/package-release.yaml` the on-chain broadcast is a separate manual job run BEFORE the tag, so the failure lands after a deploy and a pushed tag — friction and a re-tag, but no wrong bytecode on chain and nothing incorrect published to Soldeer, since the gate blocks the release rather than passing it. Value at risk is a failed release job, not a bad deployment.

ONE OVERSTATEMENT IN THE DESCRIPTION (does not sink the finding): "CLAUDE.md repeats the same claim" is not supported. CLAUDE.md:206-210 says only that `generatedContracts()` is declared once and that a contract added to it is "generated, aliased, released and frozen together" — true of the generator; it never states the three-places-and-nothing-else list. The false claim lives only in the Solidity `@dev`.

Not a documented org convention, not a misread, not already guarded structurally. Survives at LOW.

</details>

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/83

## The finding

`RegistryDeploySuites.releasedSuites()` was a hand-written concatenation of the per-contract `Lib<Contract>Released` libs. Adding a deployed contract therefore touched a fourth place, while the `@dev` immediately below it said three.

Three of those four places are self-enforcing — a candidate with no snapshot fails the shape assertions, a snapshot with no candidate fails them too, and a missing generated file fails to compile. The concat is not. A `Lib<X>Released` that exists and is concatenated nowhere compiles cleanly, is imported by nothing, is read by nothing, and leaves the whole suite green for exactly as long as that contract has never been released. It lands at the release: `cutRelease()` freezes every contract in the generated list, `testEveryFrozenSnapshotIsReleased` then reverts on the frozen file no released suite declares, and that happens during the release job with the tag pushed and the deploy already broadcast.

## The fix

`LibRainDeploySnapshot.writeReleasedSuitesAggregate` emits `src/lib/LibReleasedSuites.sol` from the same `generatedContracts()` list that writes the per-contract libs and drives the freeze, so there is no fourth place to be missing from. `releasedSuites()` is now one call to it, and `RegistryDeploySuites` spells nothing about a release by hand.

`script/Build.sol` gains `generatedContractNames()`, built once so the freeze and the aggregate cannot be handed different lists, and `regenerateLibs()` writes the aggregate — so both entry points get it, not just `cutRelease()`.

Two shared constants come out of it: `LIB_DIR` is one spelling of `src/lib` rather than one per writer, because the aggregate imports the per-contract libs by sibling path and that is only a sibling path while both writers agree on the directory; `releasedLibraryName` is one spelling of `Lib<Contract>Released`, because the writer that emits it and the aggregate that imports it have to agree or the aggregate imports a file nothing wrote.

`writeReleasedSuitesAggregate` takes `libDir` deliberately. Forge runs suites in parallel and `GeneratedSnapshotShapeTest` reads the live `src/lib/LibReleasedSuites.sol`, so a writer test aimed at that file would be rewriting it underneath a suite reading it. Mutation probe M05 below demonstrates that rather than asserting it.

## Coverage

Six new tests, in two halves that are worth nothing apart:

- `testTheCommittedAggregateIsWhatTheGeneratorEmits` — the committed file is what the generator makes of the contracts it names. A hand edit, or a generator nobody re-ran, is a failure.
- `GeneratedSnapshotShapeTest.testEverySnapshotIsInTheReleasedAggregate` — those are the contracts this repo generates, matched both directions and on count against the rolling snapshots. This is the half that closes the issue's actual hole; a file consistent with a list that left a contract out is exactly the defect being removed, and probe M06 is that file.

Matched against the snapshots rather than a list restated in the test, because a list in the test is one more place a contract has to be added to — which is the finding.

Plus the emitter units: the import block imports every released lib and nothing else; the library block for no contracts (the pre-first-release state, which must compile because the aggregate is imported by ordinary source); the library block's offsets spelled literally for one, two and three contracts rather than derived the way the emitter derives them; and the writer landing the file at `<libDir>/LibReleasedSuites.sol`, handed a list that is not this repo's so a writer that emitted the declaration it found rather than the list it was given says so.

## QA

Probed with `mutation-probe-rs`, which refuses a target that does not match exactly once and scores only against forge's own tally, so a mutant that failed to apply or failed to compile cannot be reported as a survivor. Baseline re-verified green at **221 passed / 0 failed** before every pass.

- **Discriminating tests:** `testAggregateImportBlockImportsEveryReleasedLib`, `testAggregateLibraryBlockDeclaresNothingForNoContracts`, `testAggregateLibraryBlockConcatenatesEveryReleasedLib`, `testWriteReleasedSuitesAggregateWritesTheLibAtItsPath`, `testTheCommittedAggregateIsWhatTheGeneratorEmits`, `GeneratedSnapshotShapeTest.testEverySnapshotIsInTheReleasedAggregate`. These cannot "fail on base": every symbol they exercise (`aggregateImportBlock`, `aggregateLibraryBlock`, `writeReleasedSuitesAggregate`, `pathForLib`, `LibReleasedSuites`) is added by this PR, so on base they do not compile rather than fail. Discrimination is therefore shown by mutation, below — each of the six kills at least one mutant.

- **Mutations applied** (7 killed / 1 survived):

  | # | Mutation | Verdict | Killed by |
  |---|---|---|---|
  | M01 | `aggregateImportBlock`: emit released-lib imports as `"../"` not `"./"` | KILLED | `testAggregateImportBlockImportsEveryReleasedLib`, `testTheCommittedAggregateIsWhatTheGeneratorEmits` |
  | M02 | `aggregateImportBlock`: loop `i < len` to `i + 1 < len`, silently dropping the last contract | KILLED | `testAggregateImportBlockImportsEveryReleasedLib`, `testTheCommittedAggregateIsWhatTheGeneratorEmits` |
  | M03 | `aggregateLibraryBlock`: offset expression to `""`, so every copy loop writes at `i` and overwrites earlier contracts | KILLED | `testAggregateLibraryBlockConcatenatesEveryReleasedLib`, `testTheCommittedAggregateIsWhatTheGeneratorEmits` |
  | M04 | `aggregateLibraryBlock`: empty-repo branch emits `new DeploySuite[](1)` not `(0)` | KILLED | `testAggregateLibraryBlockDeclaresNothingForNoContracts` |
  | M05 | `writeReleasedSuitesAggregate`: ignore `libDir`, write to the real `LIB_DIR` | KILLED | `testWriteReleasedSuitesAggregateWritesTheLibAtItsPath` |
  | M06 | committed `LibReleasedSuites.sol` hand-edited to drop `MigrationRegistry`, in exactly the shape the generator emits for a one-contract list | KILLED | `testEverySnapshotIsInTheReleasedAggregate` |
  | M07 | `Build.generatedContractNames()`: drop the last contract, then regenerate | KILLED | `testEverySnapshotIsInTheReleasedAggregate` |
  | M08 | `Build.regenerateLibs()`: remove the `writeReleasedSuitesAggregate` call, then regenerate | **SURVIVED** | none |

  **M06 is the one that justifies keeping both coverage halves.** The omitting file is self-consistent — it is byte-for-byte what the generator emits for the list it names — so `testTheCommittedAggregateIsWhatTheGeneratorEmits` passes on it. Only the snapshot-matched half sees it. That is the issue's actual hole, and only one of the two tests closes it.

  **M05 reproduced the race `libDir` exists to prevent, rather than arguing it.** With `libDir` ignored, the writer test overwrote the live `src/lib/LibReleasedSuites.sol` with an import of `LibThirdRegistryReleased.sol`, a file that does not exist — leaving a tree that would not compile. The probe's own pristine-tree check caught it mid-pass.

  **M08 survived, and I read it as equivalent under the current declaration rather than as a coverage gap.** The aggregate's content is a function of the contract list alone, so with that list unchanged there is nothing for the write to change and removing it is unobservable. It stops being equivalent the moment the list changes — M07 is that same code path with the list changed, and it is killed. Recorded here rather than papered over; a reviewer who wants the call itself pinned should say so. (Scored in isolation from a pristine tree: batched behind M07 it reports a false KILL, because M07's regeneration rewrites the aggregate as collateral the probe does not track or restore.)

- **Oracle:** the emitted text is asserted as literal expected strings — the offsets in `testAggregateLibraryBlockConcatenatesEveryReleasedLib` are spelled out for one, two and three contracts rather than computed, so a test that derived them the way the emitter does cannot agree with it about a wrong answer. `testEverySnapshotIsInTheReleasedAggregate` takes its oracle from the filesystem (the rolling `src/generated/candidate/` snapshots), independent of both the emitter and any list written in a test. Separately, `forge script script/Build.sol` was re-run on the committed tree and `git status --porcelain` came back empty, so the committed `LibReleasedSuites.sol` is byte-identical to what the generator produces.

- **Category check:** the issue asks for (1) the emitter beside `writeReleasedSuitesLib`, (2) the call from `regenerateLibs()`, (3) `releasedSuites()` reduced to the generated call, (4) the false `@dev` corrected — all four done. Its coverage list is covered where it is the aggregate's to cover: zero contracts emitting an empty array, and declaration order across multiple contracts. The remaining suggestions ("one contract with two releases in tag order", "one contract with no releases") are properties of the per-contract `writeReleasedSuitesLib`, already covered by the existing emitter tests; the aggregate cannot see release counts at all, since a released lib's entries are only known when the emitted source runs. On "entries whose keys are unique": the committed aggregate is compiled and exercised through `allSuites()`, which reverts `DuplicateDeploySuite` on a collision, but with no release yet cut there are currently no keys to collide. Note also that the issue's proposed signature omits `libDir`; it is added deliberately, for the reason M05 demonstrates.

## Note on the issue text

The issue's own verification block already retracts one line of its description: "CLAUDE.md repeats the same claim" is not supported — CLAUDE.md never stated the three-places-and-nothing-else list. The false claim lived only in the Solidity `@dev`, and that is what is corrected here. CLAUDE.md is not touched: the generated declaration is already stated in `RegistryDeploySuites`'s `@dev`, in `Build.sol`'s header and in the generated file's DO-NOT-EDIT banner, and CLAUDE.md's own bar is "Only what an agent working here would get WRONG".

## Summary by CodeRabbit

* **New Features**
  * Added a generated aggregate of released deployment suites, combining suites in declaration order.
  * Updated registry deployment flows to use the consolidated released-suite library.
  * Added support for generating an empty aggregate when no contracts are configured.

* **Bug Fixes**
  * Ensured generated released-suite data remains synchronized with rolling snapshots.

* **Tests**
  * Added coverage for aggregate generation, ordering, paths, contents, and snapshot consistency.


thedavidmeister:
## Brought up to date with `main` (merge `31925dc`)

`origin/main` (`cafbd08`, PR #134) merged in. Merge, not rebase; no force-push.

### The three conflicts

**`script/Build.sol`** — #134 cut the file-level exposition from 111 comment
lines to 34; this branch had edited the paragraphs that were deleted. Took
`main`'s cut version and extended it only where the aggregate makes a statement
false: `run()`'s bullet and `regenerateLibs()`'s `@notice` now name the
aggregate, and "both lib writers" became "all three lib writers".
`generatedContractNames()` is kept, with its six-line comment cut to two.
34 -> 39 comment lines.

**`src/lib/LibRainDeploySnapshot.sol`** — `main` added a defaulting
`writeReleasedSuitesLib` overload immediately after the parameterised one, in
the same place this branch added `aggregateImportBlock` /
`aggregateLibraryBlock` / `writeReleasedSuitesAggregate`. Both sides kept, in
that order.

**`test/src/lib/LibRainDeploySnapshot.t.sol`** — this branch factored the
expected generated header into `generatedFilePrefix()`; #134 changed the same
two assertions to spell it from `RAIN_SPDX_LICENSE_IDENTIFIER` /
`RAIN_COPYRIGHT_TEXT` instead of restating the text. Kept the helper and moved
its body onto the two constants, so the helper no longer restates them either
and both assertions go through it. The reuse-lint string split is preserved.

### The semantic conflict, outside the markers

`writeReleasedSuitesAggregate` took the licence and copyright as required
parameters — the shape `main` had just moved its three sibling writers away
from. Left as-is it would have been the one writer of four that a repo in this
org has to hand the org's own header to, and `script/Build.sol` would have had
to name a licence constant the repo no longer declares.

- The parameterised arity is now
  `(vm, libDir, spdxLicenseIdentifier, copyrightText, contractNames)` —
  licence after the pathing argument and before the payload, which is where
  `writeSnapshot` and `writeReleasedSuitesLib` both put it.
- A defaulting overload `(vm, libDir, contractNames)` applies
  `RAIN_SPDX_LICENSE_IDENTIFIER` / `RAIN_COPYRIGHT_TEXT`. It keeps `libDir`
  (M05).
- `script/Build.sol`'s call is unchanged and now resolves to the defaulting
  arity; it names no licence.
- `testWriteReleasedSuitesAggregateDefaultsToTheOrgHeader` added, modelled on
  the three `...DefaultsToTheOrgHeader` tests #134 added: the defaulting arity
  writes byte-for-byte what the parameterised one writes when handed the two
  constants. Pointed at `AGGREGATE_FIXTURE_DIR`, not the real `LIB_DIR`.

Also `test/lib/LibReleasedSuitesAggregate.sol` imported `forge-std-1.16.1`,
which #134 moved to `1.16.2`. Bumped — it is a branch-new file, so the version
bump did not reach it.

### Suite, on the merge commit `31925dc`

`nix develop -c forge test`: **216 passed / 51 failed / 0 skipped, 267 total.**

Every one of the 51 is `vm.createSelectFork: environment variable
`*_RPC_URL` not found` — missing RPC secrets on the machine this ran on, not
defects. Classification is mechanical:
`grep '\[FAIL' <log> | grep -vc '_RPC_URL. not found'` is **0**.

`main` after #134 on the same machine is 209 passed / 51 RPC-env failed. The
+7 is this PR's six new tests plus the defaults test above.

`nix develop -c forge fmt --check` exits 0. `forge build` is clean; the only
lint warning (`boolean-cst` at `LibRainDeploySnapshot.t.sol:1000`) is on `main`
already.

Probe: swapping the two constants in the new defaulting overload is killed by
`testWriteReleasedSuitesAggregateDefaultsToTheOrgHeader` and by
`testWriteReleasedSuitesAggregateWritesTheLibAtItsPath`.

thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens source@31925dc5ef9df8cd19fbe5e318d87acc18f754a6 + audit skill invoked at pr:126
Reviewed 31925dc5ef9df8cd19fbe5e318d87acc18f754a6: needs-work — Closes #83 — the four asks are covered, but in-diff defects: (1) aggregateImportBlock (LibRainDeploySnapshot.sol:1028) hardcodes the DeploySuite import as ../abstract/RainDeploySuitesBase.sol, so writeReleasedSuitesAggregate emits non-compiling output for any libDir except this repo's src/lib while its NatSpec (:1136-1138) claims sibling imports make any directory correct — parameterize the import or correct the claim; (2) testWriteReleasedSuitesAggregateWritesTheLibAtItsPath and testWriteReleasedSuitesAggregateDefaultsToTheOrgHeader share AGGREGATE_FIXTURE_DIR and can race under forge's concurrent in-contract execution — give the defaults test its own fixture dir; (3) RegistryDeploySuites.sol:69 "adding a contract does not touch this file at all" is false (a third candidate still edits this file — only releasedSuites() is untouched).
cost 672 — release-freeze aggregate emitter, multi-file

thedavidmeister:
Adversarial review fixes, at `fbd9610`. Six items, one of them a blocker that
made this branch's green CI meaningless.

## 1. The fixture race — blocker, was 20/20 red

`testWriteReleasedSuitesAggregateWritesTheLibAtItsPath` and
`testWriteReleasedSuitesAggregateDefaultsToTheOrgHeader` both `createDir` /
write / read / `removeDir` on one `AGGREGATE_FIXTURE_DIR =
"test/generated-aggregate"`, and forge runs the tests in a contract
concurrently — which this file already says, at `LibRainDeploySnapshot.t.sol:563`,
and which is why `RELEASED_FIXTURE_ROOT`, `SELECTED_FIXTURE_ROOT` and the two
`FREEZE_*` roots each have their own directory.

On `31925dc`, `nix develop -c forge test --mt testWriteReleasedSuitesAggregate`
failed **20 of 20 runs**:

```
36x [FAIL: vm.writeFile: failed to open file "test/generated-aggregate/LibReleasedSuites.sol": No such file or directory]
 2x [FAIL: vm.readFile:  failed to open file "test/generated-aggregate/LibReleasedSuites.sol": No such file or directory]
 2x [FAIL: vm.removeDir: failed to remove dir "test/generated-aggregate": No such file or directory]
```

The full suite passed only by scheduling luck. A release job would have hit it
with the tag pushed and the deploy broadcast, on a test with nothing to do with
the release.

**One fixture directory per test**, following the roots this contract already
separates: `fixture-lib/aggregate-path` and `fixture-lib/aggregate-defaults`.

**And out of `test/`.** The writer names the file it emits
`LibReleasedSuites.sol`, and that file imports `./Lib<Contract>Released.sol`
and `../abstract/RainDeploySuitesBase.sol` — neither resolves from
`test/generated-aggregate/`. A copy left behind there by a failure, which is
the deliberate behaviour (the `removeDir` is before the assertions), fails the
WHOLE build:

```
Error (6275): Source "test/generated-aggregate/LibAddressRegistryReleased.sol" not found
```

reproduced by hand. `fs_permissions` granted only `./src` and `./test`, both
compiled, so `foundry.toml` gains `{ access = "read-write", path = "./fixture-lib" }`.
Nothing compiles that root, and `forge build` with both fixture directories
left populated on disk now exits 0 — confirmed.

After: `--mt testWriteReleasedSuitesAggregate` **20 consecutive green runs**
(40 `[PASS]` lines, 0 `[FAIL]`), and `--mc LibRainDeploySnapshotTest` **20
consecutive green runs**, 52 tests each.

## 2. Declaration order was claimed and unpinned

The emitted library documents its entries as "in declaration order" and
`generatedContractNames()` documents itself as giving "the order the aggregate
emits its entries in" — and nothing pinned the committed file to
`generatedContracts()`. `testTheCommittedAggregateIsWhatTheGeneratorEmits`
takes the contract list from the file itself, so it holds for any permutation;
`testEverySnapshotIsInTheReleasedAggregate` is a set match.

`BuildTest.testTheCommittedAggregateIsInDeclarationOrder` pins it positionally
against `generatedContracts()`.

Verified by mutation, not asserted: swapping the two imports and the two
`released<N>` locals in the committed `src/lib/LibReleasedSuites.sol` —
byte-exactly what the generator emits for the reversed list — gives

| test | verdict on the mutant |
| --- | --- |
| `testTheCommittedAggregateIsWhatTheGeneratorEmits` | PASS |
| `testEverySnapshotIsInTheReleasedAggregate` | PASS |
| `testTheCommittedAggregateIsInDeclarationOrder` | **FAIL** — `the committed aggregate is not in the generator's declaration order: MigrationRegistry != AddressRegistry` |

Runtime impact today is confined to `suiteNames()` error text, but it is the
same shape as M06, which this PR's coverage argument is built on.

## 3. The `libDir` doc stated something false

It said "Sibling imports are what make any directory correct, so the caller's
only obligation is to name the one the released libs went to." The emitted file
also carries `import {DeploySuite} from "../abstract/RainDeploySuitesBase.sol";`,
which is parent-relative and not a sibling; and `writeReleasedSuitesLib` /
`writeAliasLib` both write to `pathForLib(libraryName)`, which takes no
directory and is always `LIB_DIR` — so the released libs can only ever be in
`src/lib` and no other argument is correct. Item 1's build break is that defect
being exercised.

The doc now states that constraint, on both arities, and
`aggregateImportBlock`'s matching claim about a repo that moves the directory
is corrected the same way.

## 4. Not fixed here — filed as https://github.com/rainlanguage/rain.deploy/issues/135

`testWriteAliasLib*` and `testWriteReleasedSuitesLib*` on `main` read-then-
restore live committed `src/lib` files, in the same shape, four tests over two
paths. Pre-existing from #134, not this PR's, deliberately not widened into it.

## 5. A false claim struck from the PR body

The body said "CLAUDE.md is extended rather than fixed, to describe the
generated declaration." The branch does not touch CLAUDE.md, verified against
`main...83-generate-released-suites-aggregate`. Struck rather than made true:
CLAUDE.md's own bar is "Only what an agent working here would get WRONG", and
the generated declaration is already stated in `RegistryDeploySuites`'s `@dev`,
in `Build.sol`'s header and in the generated file's DO-NOT-EDIT banner.

## 6. A doc contradiction inside one file

`RegistryDeploySuites.releasedSuites()`'s `@dev` ended "adding a contract does
not touch this file at all", twelve lines above `candidateSuites()`'s `@dev`
saying a third contract is "a third named candidate below, a third entry here".
It means the function; it now says so. Issue #83 exists because a comment in
that file overclaimed what adding a contract touches.

## Suite

`nix develop -c forge test` on `fbd9610`: **217 passed / 51 failed**, and

```
grep '\[FAIL' log | grep -vc '_RPC_URL. not found'   ->   0
```

All 51 are `vm.createSelectFork: environment variable *_RPC_URL not found` on
this machine — missing RPC secrets, not defects. 217 is baseline 216 plus the
one test item 2 adds. `nix develop -c forge fmt --check` exits 0.

PR #131, opened 2026-08-16
Record a migration at the moment it ran

### LINKED ISSUE #127: MigrationRegistry cannot record a migration that already ran: appliedAt is always block.timestamp
`MigrationRegistry.applyMigration` stamps `block.timestamp` and offers no way to say otherwise:

```solidity
sApplied[msg.sender][migration] = block.timestamp;
```

So a writer can only record a migration as having been applied *now*. A deploy script that already ran — before this registry existed on the chain, or before the writer started recording at all — has no way to register itself with the time it actually ran. Its options are to record a time that is wrong, or not to appear in the record at all.

Both of those defeat what the registry is for. `applied(writer, migration)` is read to answer "when did this writer apply this migration", and a backfilled record that says "today" answers it wrongly for every historical migration. Leaving it out instead makes the namespace's head wrong too, because `applyMigration` refuses anything not applied onto the current head — so a writer that skipped its past migrations cannot record the next one either, and the whole namespace is stuck at genesis.

## Ask

Let the caller supply the applied-at timestamp when registering a migration that already ran.

## What this has to settle, and does not yet

- **Which guarantee an author-supplied timestamp actually removes.** The record is namespaced by `msg.sender` and there is no authority, so every entry is already only as trustworthy as the writer that wrote it — a reader asking about a namespace it trusts is reading something only that authority could have written. What a reader can rely on today, and would stop being able to, is that `appliedAt` is a block time and therefore monotonic and bounded by the block it landed in.
- **Whether a supplied timestamp may be in the future**, and whether it must be greater than the namespace's previous record. Append-only per writer is currently a property of the head, not of the times; a settable timestamp makes "the record goes forwards" a separate thing to decide.
- **Whether this is the same entry point or a second one.** `applyMigration(expectedHead, migration)` is the live signature; a `appliedAt` parameter changes it, and an overload leaves two ways to write one record.
- **`ZeroTimestamp` becomes a caller-input check, not an environment one.** Zero is currently refused because `applied` reads it as "never applied" while the head has moved — that reasoning is unchanged, but the check moves from guarding against a zero-genesis chain to guarding against a caller passing zero, which is the ordinary case rather than the exotic one.
## Cost of the change

None of the kind first assumed here. This section originally claimed the
registry was deployed and pinned by consumers, so a signature change meant
`IMigrationRegistryV2` and a consumer migration. That is false: no release has
ever been cut. `src/generated/` holds only `candidate/`, and
`LibMigrationRegistryReleased.releasedSuites()` returns an empty array. The
`sol-v*` tags are soldeer package releases and pin no deployed address.

With an empty release record there is nothing to stay compatible with, so the
change lands in place on `MigrationRegistry` / `IMigrationRegistryV1` /
`LibMigrationRegistry`. See the correction comment below for the full evidence
and the settled answers to the four questions above.

thedavidmeister:
Closes https://github.com/rainlanguage/rain.deploy/issues/127

## The ruling: no V2

The first cut of this branch added `IMigrationRegistryV2` / `MigrationRegistryV2`
beside the existing pair, on the premise that `MigrationRegistry` was released
and pinned by consumers. That premise is false. No release has ever been cut:

- `src/generated/` holds only `candidate/` — `candidate/MigrationRegistry.sol`
  and `candidate/AddressRegistry.sol`. There is no frozen `src/generated/<tag>/`
  record, for any tag.
- `LibMigrationRegistryReleased.releasedSuites()` returns `new DeploySuite[](0)`.
- The `sol-v*` tags (`sol-v0.1.4`, `sol-v0.1.5`) are soldeer package releases.
  They pin no deployed address.

With an empty release record there is nothing to stay compatible with, so the
change lands in place on `MigrationRegistry` / `IMigrationRegistryV1` /
`LibMigrationRegistry`. Issue 127's "Cost of the change" claimed otherwise and
has already been corrected.

## The five ruled behaviours

1. **No future timestamps.** `FutureTimestamp(appliedAt, blockTimestamp)`. A
   record says a migration HAS run, and `applied() <= block.timestamp` stays true
   forever, so `block.timestamp - applied(...)` cannot underflow.
2. **The chain is the total order.** Each record keeps the head it was applied
   onto, so a namespace's records are a chain from `head` back to
   `MIGRATION_HEAD_GENESIS` whatever moments they carry. Order is never the
   caller's, whichever entry point supplied the moment.
3. **Moments are monotonic along that chain, and equal is allowed.**
   `TimestampBeforeHead(appliedAt, headAppliedAt)` refuses only `<`. Two
   migrations in one transaction share a block; two backfilled to the same day
   share a moment. The floor is read unconditionally from the record at the head
   — genesis can never be applied, so its record is zero in every namespace
   forever and the genesis case needs no branch.
4. **`ZeroTimestamp` is kept, and changes character.** It moves from guarding a
   zero-timestamp block to guarding a caller's uninitialised `uint256`, and
   guards both. Zero is still the one moment a record cannot carry, because
   `applied` reads it as "never applied" while the head has moved and the
   migration can never be applied again.
5. **Two entry points.** `applyMigration(expectedHead, migration)` stamps
   `block.timestamp` atomically; `applyMigrationHistory(expectedHead, migration,
   appliedAt)` takes the moment from the caller. Both reach one internal writer,
   so there is one record shape and one set of refusals.

## Two calls made by implementers — cheap to overrule

Neither was asked for by issue 127. Both are small to reverse now and awkward
later, so they are flagged rather than buried.

- **The external `appliedOnto` reader, and `appliedAt` on the `Migrated`
  event.** The chain in ruling 2 is stored per record; without a reader no
  consumer can traverse it, which makes it dead storage. The event's old NatSpec
  said the timestamp was the block's — once the moment can differ from the block
  the log no longer carries it, so it is added as a parameter.
- **A changed refusal in a zero-timestamp block.** `ZeroTimestamp` now precedes
  the namespace checks, so `applyMigration` on an already-applied migration in a
  block whose timestamp is zero reverts `ZeroTimestamp` where it previously
  reverted `MigrationAlreadyApplied`. Encoded by
  `testApplyMigrationZeroBlockCheckedAfterIdsAndBeforeTheNamespace`.

## One open judgement call

This repo's test tree is one file per external function —
`MigrationRegistryApplied.t.sol`, `MigrationRegistryAppliedOnto.t.sol`,
`MigrationRegistryHead.t.sol`, `MigrationRegistryApplyMigration.t.sol`. The 18
`testApplyMigrationHistory*` tests are in `MigrationRegistryApplyMigration.t.sol`
rather than in a `MigrationRegistryApplyMigrationHistory.t.sol` of their own.
Left as-is for the reviewer to call.

## QA

- `nix develop -c forge test` — **259 passed / 51 failed** (310 total). Every one
  of the 51 is a `vm.createSelectFork` failure naming a missing
  `<NETWORK>_RPC_URL` environment variable; zero non-RPC failures. CI, which
  binds the RPC vars, is green on all four checks.
- `nix develop -c forge fmt --check` — exit 0.
- `nix develop -c slither . --config-file slither.config.json` — 49 contracts,
  100 detectors, **0 findings**.

### Mutation pass

21 exact-string mutants of `src/concrete/MigrationRegistry.sol`, each probed
against `test/src/concrete/MigrationRegistry*.t.sol` — 4 suites, 79 tests, green
baseline, the file restored from git between mutants. **19 killed, 2 expected
survivors, 0 unexplained survivors.**

**The filter is narrowed deliberately, and the wide one would have made every
verdict meaningless.** `test/src/*/*Migration*.t.sol` also matches
`test/src/lib/LibMigrationRegistry.t.sol`, which deploys
`type(MigrationRegistry).creationCode` through the Zoltu factory — `CREATE2`
over its calldata at a zero salt, so the address is a pure function of the
creation code — and asserts the code hash pinned in
`src/generated/candidate/MigrationRegistry.sol`. *Every* mutant changes the
creation code, so the contract lands at a different address and the pin misses.
Measured, not assumed: mutant 20 below is semantically a no-op, and under the
wide filter it fails 24 of that suite's 39 tests with
`UnexpectedMigrationRegistryCodeHash` while all four concrete suites stay green.
`test/src/concrete/MigrationRegistry*.t.sol` uses `new MigrationRegistry()` and
pins nothing.

Each mutant run asserts it ran the same 4 suites and the same 79 tests as the
baseline, so a filter that matched nothing and a mutant that failed to compile
are reported as such rather than as survivals. Verdicts are read from
`forge test --json`, because a fuzz failure line carries `args=[...]` and a
regex over the human-readable summary silently yields no killing test at all.

| # | mutation | verdict | killing test |
| --- | --- | --- | --- |
| 1 | `ZeroTimestamp` refusal deleted | killed (7) | `testApplyMigrationHistoryZeroTimestampReverts` |
| 2 | `ZeroTimestamp` refusal inverted (`== 0` → `!= 0`) | killed (60) | `testApplyMigrationAnyCallerAppliesUnderItself` |
| 3 | `TimestampBeforeHead` refusal deleted | killed (5) | `testApplyMigrationHistoryMomentBeforeHeadReverts` |
| 4 | `TimestampBeforeHead` refuses equal too (`<` → `<=`) | killed (15) | `testApplyMigrationHistoryMomentsMayBeEqual` |
| 5 | `TimestampBeforeHead` comparison reversed (`<` → `>`) | killed (58) | `testApplyMigrationHistoryMomentAfterHeadIsAccepted` |
| 6 | the floor is read from the migration's own record, not the head's | killed (5) | `testApplyMigrationHistoryMomentBeforeHeadReverts` |
| 7 | `FutureTimestamp` refusal deleted | killed (4) | `testApplyMigrationHistoryFutureTimestampReverts` |
| 8 | `FutureTimestamp` refuses the current block (`>` → `>=`) | killed (53) | `testApplyMigrationHistoryCurrentBlockIsAccepted` |
| 9 | the record stores `block.timestamp`, not the supplied moment | killed (7) | `testApplyMigrationHistoryRecordsTheSuppliedMoment` |
| 10 | the record stores itself as its own predecessor | killed (10) | `testApplyMigrationRecordsTheHeadItWasAppliedOnto` |
| 11 | `Migrated` carries the block, not the recorded moment | killed (1, sole) | `testApplyMigrationHistoryEvent` |
| 12 | `applyMigration` stamps a constant, not the block | killed (8) | `testApplyMigrationStoresTheBlockTimestamp` |
| 13 | `appliedOnto` drops its three key refusals | killed (5) | `testAppliedOntoZeroMigrationReverts` |
| 14 | `appliedOnto` always answers zero | killed (11) | `testAppliedOntoFirstRecordIsGenesis` |
| 15 | `MigrationAlreadyApplied` refusal deleted | killed (6) | `testApplyMigrationTwiceReverts` |
| 16 | `UnexpectedMigrationHead` refusal deleted | killed (10) | `testApplyMigrationSkippedPredecessorReverts` |
| 17 | `head` returns the stored zero instead of genesis | killed (61) | `testHeadEmptyNamespaceIsGenesis` |
| 18 | genesis gets a moment floor of its own | killed (10) | `testApplyMigrationHistoryNoMomentBoundAtGenesis` |
| 19 | `applyMigration` clamps a zero block instead of refusing it | killed (2) | `testApplyMigrationZeroBlockReverts` |
| 20 | the two id refusals swap order | **survived, expected** | `bytes32(0) != MIGRATION_HEAD_GENESIS`, so no input reaches both and the order is unobservable |
| 21 | `TimestampBeforeHead` moved past `FutureTimestamp` | **survived, expected** | `headAppliedAt <= block.timestamp` always holds, so no input reaches both; see below |

The count in the verdict column is how many of the 79 tests failed. A row marked
"sole" had exactly one killer; the named test is one of the actual failing set in
every row.

## Check order, and what it does and does not pin

`ZeroMigration` → `GenesisMigration` → `ZeroTimestamp` →
`MigrationAlreadyApplied` → `UnexpectedMigrationHead` → `TimestampBeforeHead` →
`FutureTimestamp`.

Refusals about the call alone, then about the namespace it arrives at, then
about the block it lands in — the order in which a caller can do something about
them. `FutureTimestamp` is last because it is the only one time itself resolves.

The last two are **mutually exclusive on a real chain**: `headAppliedAt <=
block.timestamp` always holds, so no input reaches both `TimestampBeforeHead` and
`FutureTimestamp`, and their relative order is unobservable outside a test that
warps backwards. No test pins it, and the mutation pass records the swap as an
expected survivor rather than hiding it.


thedavidmeister:
🤖 ai:producer

Producer note: this PR is now GREEN, and two claims in the body above are stale — correcting the record here since the body itself is not mine to edit.

- **"Pre-existing red: `rainix / static / static` ... fails on `main` too" is no longer true, and the diagnosis was wrong on its own terms.** It was accurate when written: at this branch's merge base (`ad1e00c`), `CLAUDE.md` was 24016 bytes, over the 4096-byte `agent-context-cap`. But `main` has since cut that file to 2344 bytes, under the cap. The red was this branch's alone — the branch was 91 commits behind, and the PR *added* 62 lines of V2 prose to the fat file.
- **The "307 tests" figure is stale** — 342 after merging `main`.

Fixed by merging `origin/main`. `CLAUDE.md` resolved to main's cut-down version verbatim (the PR's additions were exactly the architecture prose main's new bar deletes; the V2 rationale already lives in `src/interface/IMigrationRegistryV2.sol` natspec). A second commit carries across main's correction to the `checkCodeHash` natspec, which `LibMigrationRegistryV2.sol` had inherited from V1 in its factually-wrong form (it claimed a call into an empty account "decodes as zero"; solc actually reverts when returndata is too short to decode). Comments only — creation code and every generated snapshot are unchanged, and `forge script script/Build.sol` regenerates byte-identical.

This is **not** the split-release-lifecycle migration case: the failing check was `agent-context-cap`, not a prod-pin / `testProdDeploy*` / deploy-pinned-constant check. Nothing was deployed.

Worth a reviewer's attention: because the static job fails fast at step 4, steps 5–11 had **never run** on this branch until now. Verified locally in the checkout's own toolchain: slither 0 findings over 54 contracts, `forge fmt --check` clean, `rainix-sol-single-contract` OK, `snapshots-append-only` clean, `agent-context-cap` 2209/4096. `forge test` 291 passed / 51 failed, every failure an `ARBITRUM_RPC_URL`/`BASE_RPC_URL` fork-env failure with no local RPC secrets — zero code failures.

thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens source@168196f3e9247ae52cfa90388edece9b2319c5f6 + audit skill invoked at pr:131
Reviewed 168196f3e9247ae52cfa90388edece9b2319c5f6: needs-work — Closes #127: V2 settles all four issue questions correctly (single entry point, future-timestamp refusal, floor-vs-head with equal allowed, ZeroTimestamp as caller-input guard) with boundary and check-order tests; send-back is convention-level: fuzz params `now_` break the org no-underscore param rule (MigrationRegistryV2ApplyMigration.t.sol:70,86,104,152,257; LibMigrationRegistryV2.t.sol:257) — rename them; and LibMigrationRegistryV2.sol:88's "empty account's code hash is zero" should note a touched codeless account presents keccak256("").
cost 642 — new registry concrete with timestamp guards

thedavidmeister:
**Superseded: there is no V2.**

Ruling: no release of `MigrationRegistry` was ever cut, so nothing pins it and there is no compatibility surface to preserve. Verified against main — `src/generated/` holds only `candidate/` with no frozen `<tag>/` directory, and `LibMigrationRegistryReleased.releasedSuites()` returns `new DeploySuite[](0)`. The `sol-v0.1.4`/`sol-v0.1.5` tags are soldeer *package* releases and pin no deployed address.

Issue #127's "Cost of the change" section asserts the opposite and is what produced this PR's shape. It is wrong and will be corrected.

This PR is being rewritten as an in-place change to `MigrationRegistry` / `IMigrationRegistryV1` / `LibMigrationRegistry`, with all 11 V2 files deleted. The head commit here is still the old V2 work; the rewrite is in flight and unpushed. The body below still describes V2 and will be replaced when it lands.

The behaviour, all ruled:

1. **No future timestamps** — a supplied `appliedAt` above `block.timestamp` is refused.
2. **Total ordering is structural** — each record stores the head it was applied onto, so entries form a linked list and order is a property of the chain, not the clock.
3. **Timestamps are still monotonic, allowing equal** — a new record's `appliedAt` must be `>=` that of the record at the current head. Equal accepted, earlier refused, no comparison at genesis. This is a separate rule from (2), not replaced by it.
4. **`ZeroTimestamp` stays**, now guarding a caller passing zero rather than a zero-genesis chain.
5. **Two entry points, one record** — one takes `appliedAt` for a migration that already ran, one stamps `block.timestamp` for the atomic case inline with the script doing the applying.

PR #134, opened 2026-08-17
Bump rain-sol-codegen 0.1.6 -> 0.1.36 and forge-std 1.16.1 -> 1.16.2

thedavidmeister:
Bumps `rain-sol-codegen` from `0.1.6` to `0.1.35`, the latest published release.

`forge-std` goes `1.16.1` -> `1.16.2` in the same change, and that is forced rather than scope creep: codegen 0.1.35's own `LibFs.sol` and `LibCodeGen.sol` import `forge-std-1.16.2/src/Vm.sol`, and a `Vm` from 1.16.1 is a different type — nothing compiles otherwise.

## Why this is more than a prefix rewrite

- **`filePrefix()` now takes required `(spdxLicenseIdentifier, copyrightText)`** with no defaulting overload — rain.sol.codegen#135, closing its issue #75, so the library stops stamping one org's licence into another repo's generated files. `buildFileForContract` gained the same parameters.
- **`LibRainDeploySnapshot` was smuggling a path separator through the contract-name argument** (`pathForContract(snapshotName(dir, contractName))`). Current codegen refuses that. `pathForTaggedContract` / `buildFileForTaggedContract` were added for exactly this layout (rain.sol.codegen#137, its issue #78), and this repo is the consumer they were built for.

## Decisions

**Licence and copyright are parameters of `LibRainDeploySnapshot`, not constants inside it.** Declared in `script/Build.sol` and threaded down. `rain-deploy` is itself published and `rain.factory.deploy` consumes it, so hardcoding Rain's header in the library would recreate one layer up the precise defect codegen #135 fixed.

**`snapshotName` is deleted** — it existed only to build the smuggled name.

**`LIB_FS_ROOT = GENERATED_DIR`** imported from `LibFs` rather than a second `"src/generated"` literal.

**The root-aware `dirForSnapshot`/`pathForSnapshot` overloads carry the writer's guards** (`requireTag` then `requireIdentifier`, in `pathForTaggedContract`'s order), so both path spellings share one validity rule.

Fixture dirs renamed to the tag alphabet (`write-snapshot-not-a-tag` -> `writeSnapshotNotATag`); they must stay non-`isTag` while being alphabet-legal. Record roots and raw `string.concat` fixtures are untouched.

## The regeneration is byte-identical, and that is now measured

`forge script script/Build.sol` on this head, then `git status`: **empty**. Not just `src/generated/candidate/` — nothing anywhere in the tree moved. Both rolling snapshots, both alias libs and both released-suites libs regenerate to the bytes already committed under codegen 0.1.6.

That claim was previously reasoning about the two versions' emitters. It is now the measurement, which matters because rainix runs `frozen-snapshots-append-only` and because the deploy pins consumers hold are exactly these bytes: a bump that moved them would move `DEPLOYED_ADDRESS`.

## QA

**Discriminating tests.** `testRootAwareSnapshotPathIsTheWritersAtTheRealRoot`, rewritten as a staticcall outcome-parity fuzz — both path spellings accept/refuse the same names and return the same bytes, through new `externalPathForSnapshotAt` / `externalPathForSnapshot` wrappers. Writer-test expected headers are built from `script/Build.sol`'s licence/copyright constants, so those constants are anchored to the committed artifacts rather than restated. Each fails on base by construction: `snapshotName` no longer exists on this head and the writers take two more arguments, so the base spelling does not compile.

Added here: **`testWriteSnapshotDeclaresTheLicenceItWasHanded`**. Making the header a parameter is only worth doing if what is handed in is what comes out, and the snapshot writer had no assertion on that — the alias and released emitters each have a byte-for-byte one, and every other test that reads a snapshot header reads the *committed* file, which is evidence about a generation that already happened. The two values are strings of the same type in adjacent positions, so a call that swapped them compiles and writes a file carrying both, in the wrong tags, into a consumer's append-only record. The test asserts the header *entire* — everything before `pragma` — against text spelled out rather than taken from `LibCodeGen.filePrefix`, so the oracle is not the code that wrote the file.

**Mutations.** The pass over the changed lines recorded **11 mutants, 11/11 killed, 0 survived** — the deleted `snapshotName` path, the guard placement in the root-aware overloads, and the licence/copyright threading. That pass is not re-run here.

Run here, for the test added here: swapping `spdxLicenseIdentifier` and `copyrightText` at the `buildFileForTaggedContract` call in `writeSnapshot`. **Killed** — `LibRainDeploySnapshot.t.sol` goes 43 passed to 42 passed / 1 failed, and the single failure is the new test, on `// SPDX-License-Identifier: Copyright (c) 2020 Rain Open Source Software Ltd`. Nothing else in the file notices, which is the gap it was written for.

**Oracle.** The committed artifacts under `src/generated/candidate/`, and now the empty regeneration diff against them rather than an argument about emitters.

**Category check.** The ask is one thing — bump to the latest codegen release. Covered: the pin, `soldeer.lock`, `remappings.txt`, all 19 `.sol` import prefixes (zero `rain-sol-codegen-0.1.6` or `forge-std-1.16.1` survive), both API changes above, and the README versions.

## Checks run locally

- `forge script script/Build.sol` — successful, zero diff.
- `reuse lint` — compliant, 82/82 files.
- `forge fmt --check` — clean.
- `slither .` — 49 contracts, 100 detectors, **0 results**.
- `forge test` — **206 passed, 51 failed, 0 skipped, 257 total** across 18 suites. All 51 failures are `vm.createSelectFork: environment variable ARBITRUM_RPC_URL / BASE_RPC_URL / POLYGON_RPC_URL not found` — fork tests wanting RPC secrets CI has and a local shell does not. Counted, not sampled: zero of the 51 is anything else, and none is related to this change.

## The symlink/ffi risk is closed

codegen 0.1.35 reaches `removeSymlink`, and its `vm.tryFfi`, only inside `if (vm.exists(path) && isSymlinkIn(vm, dir, path))` — an actual symlink sitting at a generated path. This repo does not set `ffi = true`, so that branch would revert.

It is unreachable here: `find src test script -type l` is empty, there is no `ln -s` or symlink-creating call anywhere outside `dependencies/`, and both the regeneration and the full suite write generated files with `ffi` unset without hitting it.

## About the earlier red on `db4e887`

All three checks on `db4e887` failed at step 1, **"Set up job"**, 48-88 seconds after starting at `14:51:18Z` — before checkout, before dependencies, before anything in this diff ran. That is inside the critical-impact GitHub incident opened at `13:40:03Z` with Actions in major outage. The runner never started, so those reds are not a verdict on this code and nothing about them was fixed, because there was nothing there to fix.

Actions has since recovered. `rainix / static`, `rainix / legal` and `rainix / test` are all **completed/success** on `dc3204f`, started `15:42:03Z`.

## Summary by CodeRabbit

* **Documentation**
  * Updated documented dependency versions and configuration examples.

* **Refactor**
  * Improved snapshot path handling and generation reliability.
  * Added configurable SPDX license and copyright metadata to generated files.

* **Tests**
  * Expanded snapshot path and metadata validation.
  * Updated test coverage for generated headers and snapshot writers.

* **Chores**
  * Updated Forge and code-generation tooling versions across the project.


thedavidmeister:
@coderabbitai review

The earlier review failed during the GitHub outage, and the head has moved to `dc3204f`.

thedavidmeister:
Re-pointed 0.1.35 -> **0.1.36**. 0.1.35 predates the licence defaults; 0.1.36 is the release cut from https://github.com/rainlanguage/rain.sol.codegen/pull/144, which restores `RAIN_SPDX_LICENSE_IDENTIFIER` and `RAIN_COPYRIGHT_TEXT` as constants and adds the arities of `filePrefix` and the three `LibFs` writers that resolve to them.

Also deleted the `rain-sol-codegen-0.1.35/` remapping. `forge soldeer update` appends the new remapping and leaves the superseded one behind, pointing at a `dependencies/` directory that no longer exists.

Local: `forge fmt --check` exit 0, 206 passed. The 51 failures are all `vm.createSelectFork: environment variable \`*_RPC_URL\` not found` — 102 of 102 `[FAIL` lines match that, none of any other kind — which is this machine having no RPC secrets, not the bump.

thedavidmeister:
Removed rain.deploy's own licence constants.

`script/Build.sol` declared `GENERATED_SPDX_LICENSE_IDENTIFIER` and `GENERATED_COPYRIGHT_TEXT` as its own string literals, holding the same two values 0.1.36 exports as `RAIN_SPDX_LICENSE_IDENTIFIER` and `RAIN_COPYRIGHT_TEXT`. Two definitions of one invariant with nothing comparing them. Both deleted; the 3 call sites in `Build.sol` and 12 in `LibRainDeploySnapshot.t.sol` name codegen's, which the exported constants' own NatSpec anticipates — `Exported rather than inlined into the defaulting overload, so a consumer that threads the header through its own build can name the value instead of restating the string.`

`LibRainDeploySnapshot` and `LibFs` still take the header as parameters. Both are dependencies of deploy repos in other orgs whose `src/generated/<tag>/` is append-only, so neither may choose one; what changed is only which repo the values are written down in.

**The emitted header is unchanged.** `testWriteAliasLibWritesTheLibAtItsPath` emits with these constants and asserts the result is the committed generated file byte for byte — it passes, and nothing under `src/generated/` moved.

Local, before and after, identical: 206 passed; 51 failures, all `vm.createSelectFork: environment variable \`*_RPC_URL\` not found` (102 of 102 `[FAIL` lines), confined to `RainDeployVerifyChain`, `RainDeployVerifyChainCandidate` and `LibRainDeploy`. This machine has no RPC secrets.

thedavidmeister:
Pushed the defaults one layer up, which is where the question actually was.

Codegen 0.1.36 defaults the header at its own writers, but nothing in this repo calls those directly — every caller goes through `LibRainDeploySnapshot`, whose `writeSnapshot`, `writeAliasLib` and `writeReleasedSuitesLib` each had exactly one arity, all three requiring the licence and copyright. So a repo in this org still had to name the values and `script/Build.sol` did. That is the same either/or the codegen change already removed one layer down: taking the values as parameters is what a repo *outside* this org needs, and it never ruled out an arity that defaults.

Each of the three writers now has the defaulting overload. `Build.sol` drops the arguments and the import — **nothing in this repo names a licence in the build path any more**.

Covered three ways:

- The two tests that emit over the committed generated files and assert byte equality now go through the defaulting arity, so the defaults are pinned to the headers the committed files actually carry.
- One test per writer asserts the defaulting arity writes exactly what the parameterised one writes when handed the org values.

Mutations, applied to the branch and reverted:

| Mutation | Result |
| --- | --- |
| `writeAliasLib` defaulting overload passes the two constants swapped | `[FAIL: assertion failed: // SPDX-License-Identifier: Copyright (c) 2020 Rain Open Source Software Ltd]` — `testWriteAliasLibDefaultsToTheOrgHeader`, `testWriteAliasLibWritesTheLibAtItsPath` |
| `writeSnapshot` defaulting overload passes `"MIT"` | `[FAIL: assertion failed: // SPDX-License-Identifier: MIT]` — `testWriteSnapshotDefaultsToTheOrgHeader` |

`testWriteSnapshotDefaultsToTheOrgHeader` reverts EVM state between its two writes: the Zoltu deploy is CREATE2, so writing one creation code twice lands on an address that already has code. Caught by the first run of it, which failed `DeployFailed(false, 0x0)`.

209 passed, up from 206. Same 51 RPC-env failures. `forge fmt --check` exit 0.

PR #175, opened 2026-09-14
Test LibStringSet.holds membership directly

thedavidmeister:
## What

Adds `test/src/lib/LibStringSet.t.sol`, a unit test for `LibStringSet.holds`.
No source change.

## Why

`holds` had no test of its own. Its only exercise was indirect, through
`GeneratedSnapshotShapeTest` and `BuildTest`, which pass it real contract and
release names. Those names never collide in length without also colliding in
content, and the lists are never empty, so a large part of what the NatSpec
claims was asserted nowhere: that equality is over the whole content rather
than the length or a prefix of it, that an empty list holds nothing, and that
the empty string is an ordinary member, held only when an element is itself
empty.

The NatSpec is the oracle. It says membership over a `string[]` for a set whose
order it does not fix, compared over the whole string. Each test states one case
of that definition, with the answer derived from the definition rather than from
the loop that implements it.

## QA

- Discriminating tests: testHoldsIsFalseOnAnEmptyHaystack, testHoldsComparesTheWholeContent, testHoldsComparesBeyondTheFirstWord, testHoldsEmptyNeedleMatchesOnlyAnEmptyElement - each was verified by re-running the whole suite against the mutant it targets: the mutant survived the suite before this file and is killed by it after, with the probe naming the test.
- Mutations applied: see the table below - each row is one line of src/lib/LibStringSet.sol replaced, the verdict of the whole suite before this file, the verdict after, and the test the probe named as the killer.
- Oracle: the LibStringSet NatSpec - membership over a string[] for a set whose order it does not fix, compared over the whole string. Every expected value in the tests is a concrete list and needle answered from that definition, never from the keccak loop that implements it.
- Category check: this is not filed against an issue. It is one unit of a mutation sweep of the repo; the categories are the behaviour lines of holds, and all eleven are in the table with a verdict.

Mutation matrix over `src/lib/LibStringSet.sol`. Suite is `forge test` minus the
61 fork tests that need RPC env vars absent on the machine this ran on.
`LibStringSet` is not compiled into `AddressRegistry` or `MigrationRegistry`, so
the pinned address and code hash tests carry real behaviour for it and were kept
in rather than excluded. Baseline before: 397 passed, 0 failed. Baseline after:
409 passed, 0 failed. Every mutant below was applied to the unmutated source one
at a time and the whole suite run against it.

Before the new file, 6 of 11 mutants were killed by pre-existing indirect tests
and 5 survived. After it, 11 of 11 are killed.

| Behaviour | Mutation | Before | After | Killing test |
|---|---|---|---|---|
| A match returns true | `return true` inside the loop becomes `return false` | KILLED | KILLED | testSnapshotContractNamesAreTheGeneratedContracts |
| No match returns false | the trailing `return false` becomes `return true` | KILLED | KILLED | testFrozenSnapshotPathsExcludesTheRollingSnapshot |
| An empty list holds nothing | the trailing `return false` becomes `return haystack.length == 0` | SURVIVED | KILLED | testHoldsIsFalseOnAnEmptyHaystack |
| Index 0 is searched | the loop starts at `i = 1` | KILLED | KILLED | testSnapshotContractNamesAreTheGeneratedContracts |
| The last index is searched | the loop bound becomes `i + 1 < haystack.length` | KILLED | KILLED | testSnapshotContractNamesAreTheGeneratedContracts |
| Equality is content, not length | the keccak comparison becomes a length comparison | SURVIVED | KILLED | testHoldsComparesTheWholeContent |
| Same length, different bytes are different | the comparison becomes length plus first byte only | SURVIVED | KILLED | testHoldsComparesTheWholeContent |
| The empty needle is an ordinary member | an early `return true` when the needle is empty | SURVIVED | KILLED | testHoldsEmptyNeedleMatchesOnlyAnEmptyElement |
| Equality is by value, not by memory identity | the comparison becomes pointer equality in assembly | KILLED | KILLED | testSnapshotContractNamesAreTheGeneratedContracts |
| A match is the answer whatever follows it | the loop keeps going and returns the last comparison | KILLED | KILLED | testSnapshotContractNamesAreTheGeneratedContracts |
| Content past the first word is compared | the comparison becomes length plus the first 32 bytes only | SURVIVED | KILLED | testHoldsComparesBeyondTheFirstWord |

No bug was found in `holds`. Every surviving mutant was a test gap: the current
output is the one the NatSpec calls for in each case.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for string membership checks, including empty values, varied element positions, ordering, whole-content comparisons, and longer strings.
  * Added fuzz testing for membership behavior and strings exceeding the length of stored elements.
  * Added mutation-testing configuration to evaluate the robustness of the test suite.


thedavidmeister:
Closing unmerged: its subject is gone. LibStringSet was deleted from this repo in #210, which replaced all sixteen call sites with LibMemoryKV. The membership behaviours these tests pinned are now covered by that library's own suite, and the five gaps they closed are moot because the loop they covered no longer exists here.

PR #184, opened 2026-09-14
test: pin every generated contract's committed alias lib to its declaration

thedavidmeister:
## What

One read-only test in `test/script/Build.t.sol`:
`testTheCommittedAliasLibsAreWhatTheDeclarationEmits`. For EVERY entry of
`Build.generatedContracts()` it asserts the committed
`src/lib/Lib<contractName>Deploy.sol` is byte-exactly what the alias emitter
produces from THAT ENTRY — its `contractName`, its `constantPrefix`, and
`LibRainDeploySnapshot.CANDIDATE`.

Nothing is written, so it cannot race the suites that compile those files.

## Why

`constantPrefix` was asserted only to be non-empty and distinct from the other
entry's. Neither says what it must be, and it names every constant the alias lib
exports: `ADDRESS_REG` in place of `ADDRESS_REGISTRY` is unique, non-empty, and
renames `ADDRESS_REGISTRY_DEPLOYED_ADDRESS` and
`ADDRESS_REGISTRY_DEPLOYED_CODEHASH` — the two names `LibAddressRegistry` and
every downstream consumer pin. The declaration stays green and the next
regeneration is what performs the rename.

The existing emitter checks in `LibRainDeploySnapshot.t.sol` hard-code their own
contract and their own prefix, so they agree with a declaration they never read.
They also cover `AddressRegistry` alone, which left the second generated
contract's alias lib — `LibMigrationRegistryDeploy.sol` — pinned to the
generator by nothing at all.

## QA

- Discriminating tests: testTheCommittedAliasLibsAreWhatTheDeclarationEmits - fails on base under each of the two prefix mutations below, verified by re-probing them (SURVIVED before the test, KILLED after).
- Mutations applied: script/Build.sol generatedContracts()[0].constantPrefix ADDRESS_REGISTRY -> ADDRESS_REG -> testTheCommittedAliasLibsAreWhatTheDeclarationEmits; script/Build.sol generatedContracts()[1].constantPrefix MIGRATION_REGISTRY -> MIGRATION_REG -> testTheCommittedAliasLibsAreWhatTheDeclarationEmits. Full 32-mutant matrix over both files below.
- Oracle: the committed src/lib/Lib<contract>Deploy.sol files and the GeneratedContract NatSpec, which states constantPrefix is the prefix for the constants the alias lib exports. The expected text is built from the alias emitters, not restated from the declaration under test.
- Category check: no issue; this is an adversarial mutation pass over script/Build.sol and src/abstract/BuildScript.sol. Both files behaviour lists covered, with the not-executed rows named explicitly below.

Adversarial mutation probe over `script/Build.sol` and
`src/abstract/BuildScript.sol`. Suite: `check.sh` — `forge test` minus the 61
fork tests that need RPC env vars absent on this machine. Not `check-nopin.sh`:
neither file under probe is compiled into `AddressRegistry` or
`MigrationRegistry`, so the bytecode pin tests carry real behaviour here and are
kept. Baseline before: 397 passed, 0 failed. Baseline after: 398 passed,
0 failed.

Gaps this PR closes:

| behaviour | mutation | before | after | killed by |
| --- | --- | --- | --- | --- |
| `generatedContracts()[0].constantPrefix` is `ADDRESS_REGISTRY` | `constantPrefix: ADDRESS_REG` | SURVIVED | KILLED | testTheCommittedAliasLibsAreWhatTheDeclarationEmits |
| `generatedContracts()[1].constantPrefix` is `MIGRATION_REGISTRY` | `constantPrefix: MIGRATION_REG` | SURVIVED | KILLED | testTheCommittedAliasLibsAreWhatTheDeclarationEmits |

Behaviours already covered, left as they are:

| behaviour | mutation | verdict | killed by |
| --- | --- | --- | --- |
| exactly one entry per declared candidate | array truncated to 1 entry | KILLED | testGeneratedContractsAreExactlyTheDeclaredCandidates |
| `[0].contractName` is the artifact's contract | `AddressRegistryX` | KILLED | testGeneratedContractNameMatchesTheArtifactPath |
| `[0].constantPrefix` non-empty | `constantPrefix` emptied | KILLED | testConstantPrefixesAreNonEmptyAndUnique |
| `[0].constantPrefix` unique | set to `MIGRATION_REGISTRY` | KILLED | testConstantPrefixesAreNonEmptyAndUnique |
| `[0].candidate` is `addressRegistryCandidate()` | `migrationRegistryCandidate()` | KILLED | testGeneratedContractNameMatchesTheArtifactPath |
| `[1].contractName` is `MigrationRegistry` | duplicated as `AddressRegistry` | KILLED | testGeneratedContractNamesAreUnique |
| `[1].constantPrefix` unique | set to `ADDRESS_REGISTRY` | KILLED | testConstantPrefixesAreNonEmptyAndUnique |
| `[1].candidate` is `migrationRegistryCandidate()` | `addressRegistryCandidate()` | KILLED | testGeneratedContractsAreExactlyTheDeclaredCandidates |
| `snapshotContractNames()` length is the declaration's | truncated to 1 | KILLED | testSnapshotContractNamesAreTheDeclarationInOrder |
| `snapshotContractNames()` is declaration order | reversed | KILLED | testSnapshotContractNamesAreTheDeclarationInOrder |
| `recordRoot()` defaults to `LIB_FS_ROOT` | returns `CANDIDATE` | KILLED | testRecordRootDefaultsToTheRepoRecord |
| `cutRelease()` freezes into `recordRoot()` | `LIB_FS_ROOT` literal | KILLED | testCutReleaseFreezesTheRegeneratedSnapshot |
| `run()` regenerates snapshots | call deleted | KILLED | testRunRegeneratesAndFreezesNothing |
| `run()` regenerates libs | call deleted | KILLED | testRunRegeneratesAndFreezesNothing |
| `run()` does snapshots before libs | calls swapped | KILLED | testRunRegeneratesAndFreezesNothing |
| `run()` freezes nothing | `freeze` inserted | KILLED | testRunRegeneratesAndFreezesNothing |
| `freeze` gets `regenerateSnapshots` as its hook | `regenerateLibs` passed | KILLED | testCutReleaseFreezesTheRegeneratedSnapshot |
| `freeze` gets `snapshotContractNames()` | empty name list | KILLED | testCutReleaseFreezesTheRegeneratedSnapshot |
| the frozen copy is the regenerated bytes | no-op regenerate hook | KILLED | testCutReleaseFreezesTheRegeneratedSnapshot |
| `cutRelease()` regenerates libs after the freeze | libs moved before it | KILLED | testCutReleaseRegeneratesLibsFromTheRecordJustCut |
| `cutRelease()` regenerates libs at all | call deleted | KILLED | testCutReleaseRegeneratesLibsFromTheRecordJustCut |

Not executed by any test, and not fixable from the test side: every behaviour of
`Build.regenerateLibs` and `Build.regenerateSnapshots`. A bare `revert` inserted
as the first statement of either one survives the whole suite. Both hard-code
`LibRainDeploySnapshot.LIB_DIR` and `CANDIDATE`, so any call rewrites committed
source other suites compile — which is why `BuildTest` documents that it refuses
to call `run()` or `cutRelease()`. Six mutations inside them survive: each
loop truncated to one contract, the alias pointed at a tag dir rather than the
candidate, `recordRoot()` replaced by the `LIB_FS_ROOT` literal at the
released-lib call, the aggregate write deleted, and the snapshot written from
`candidate.snapshot.creationCode` rather than `candidate.sourceCreationCode`.
Reported separately rather than papered over here.

`run()` and `cutRelease()` being non-virtual is a structural property Solidity
gives a test no way to observe: adding `virtual` changes no behaviour and
survives by construction.


thedavidmeister:
Closing unmerged as a duplicate. Its single test, testTheCommittedAliasLibsAreWhatTheDeclarationEmits, asserts the same property as testEveryCommittedAliasLibIsWhatTheGeneratorEmits, which landed on main in #176 and is driven from the same generatedContracts() list. #176 also covers the released-suites lib, which this does not, so merging this would add a second name for a covered behaviour and a conflict to resolve for nothing. Two campaign batches reached the same gap from different units.

Issue #202, opened 2026-09-14
An unset DEPLOYMENT_SUITE can select and broadcast a suite, because the empty string is a declarable suite key


thedavidmeister:
## Unit and file

`suiteByName` / `allSuites` in `src/abstract/RainDeploySuitesBase.sol` (lines 237-288), reached from `run()` in `src/abstract/RainDeployBroadcast.sol` line 93.

Found twice, independently, from both ends: once from the suites base (the empty string is a declarable key) and once from the broadcast (an unset env var is exactly that key). One defect, one place to fix it, so one issue.

## The intent oracle

`test/src/abstract/RainDeploySuitesBase.t.sol` on `testEmptySuiteIsUnknown`:

> An empty key is just another unknown key, so an unset `DEPLOYMENT_SUITE` reports the valid set rather than silently taking a default.

`test/src/abstract/RainDeployBroadcast.t.sol` lines 67-78, on why the unset leg must use a genuinely absent variable rather than a present-and-empty one:

> ... it would pass just as happily if that default became a real suite key, and a deploy that picks something when told nothing is how the wrong contract reaches a chain.

and immediately after:

> That the reported key is empty rather than a suite is what says there is no default; `RainDeploySuitesBaseTest.testEmptySuiteIsUnknown` is what says an empty key is unknown.

Sibling rule for comparison: `NoDeployCandidates` (same file, lines 17-35) is already a rule about declaration SHAPE living in this abstract, adopted because a declaration that is wrong in shape has quietly opted out of a check while every other assertion stays green.

## The property violated

Never-defaults. `run()` uses the empty string as its absent-sentinel, and the empty string is simultaneously a legal key in the space that sentinel is matched against. The sentinel is reserved nowhere: `checkedCandidateSuites` refuses only an EMPTY LIST, and `allSuites` refuses only DUPLICATE keys. Nothing anywhere requires a key to be non-empty, so the never-defaults property holds only because no declaration happens to use that key today. A downstream declaration with an omitted or empty key string turns an operator dispatch with the suite input left blank into a silent selection, and CREATE2 at a zero salt is permanent on every chain the dispatch reached.

`testEmptySuiteIsUnknown` asserts that the FIXTURE registry has no empty key. It asserts nothing about a registry that does.

## Verified repro, and what it does now

Re-verified against current `main` (e9a922a) in a fresh clone.

[...]

PR #219, opened 2026-09-15
Hold a deploy suite key to an alphabet the key list reads back as

### LINKED ISSUE #203: The valid-key list in UnknownDeploymentSuite cannot be read back as the set of keys it reports
## Unit and file

`suiteNames()` in `src/abstract/RainDeploySuitesBase.sol` lines 262-270, and the `UnknownDeploymentSuite` error it feeds (same file, lines 10-15).

## The intent oracle

The error's own NatSpec:

> Thrown when `DEPLOYMENT_SUITE` names no declared suite. Carries the valid keys, because the whole point of a registry is that the answer is not one hardcoded string the caller has to already know.

with `@param validSuites The declared keys, comma separated.`

`testSuiteNamesIsTheRegistry` pins the rendering of the real fixture registry, and `testUnknownSuiteNamesEveryValidSuite` pins that an unknown key fails naming every valid key.

## The property violated

The list is meant to tell a caller who does not already know the answer what the valid keys ARE. `suiteNames` joins with a comma and a space and the key alphabet is unconstrained, so the rendering is not injective: a two-suite registry keyed `a, b` and `c` renders identically to a three-suite registry keyed `a`, `b`, `c`. A reader of the failure is sent after a key that does not exist, and is told a different number of suites exist than do.

Re-verified against current `main` (e9a922a) in a fresh clone.

Fixture: an `ExternalDeploySuites` with two candidates keyed `a, b` and `c`, both anchored to `type(AddressRegistry).creationCode`.

- `externalAllSuites().length` is 2.
- `externalSuiteNames()` is `a, b, c`.
- `externalSuiteByName` with an unknown key reverts `UnknownDeploymentSuite(<key>, a, b, c)` — a two-suite registry rendering as three apparent keys.

`nix develop -c forge test --match-contract TmpReproTest`: 2 passed, 0 failed (this leg and the empty-key leg together). Fixture deleted afterwards; tree clean.

## Triage

For intended: this string is revert data a human reads, not a parsed protocol value. A grep for consumers finds exactly one use — line 287 of the same file — and nothing under `.github/`, `script/` or `tools/` touches `validSuites`. The property the docs and `testSuiteNamesIsTheRegistry` actually assert is that the list is BUILT from `allSuites()` rather than restated, so it cannot drift from the registry, and that holds: every declared key appears verbatim. The blast radius is one wasted retry against a still-failing lookup, with no irreversible action behind it. Fixing it means constraining key characters, which is a new rule about key contents the declaration deliberately does not carry.

For bug: the error exists precisely so a caller does not have to already know the answer, and a list whose parse is ambiguous partly defeats that. Keys are unconstrained strings, so nothing prevents the rendering.

Note: the independent verification pass returned INTENDED on this one, having grepped for consumers and found none, and recommended closing it. Filed for the record and for the triage decision rather than as an asserted defect. The empty-key rendering mentioned by the original candidate is not independent evidence here — that is a consequence of the empty-key issue filed separately, not a defect in `suiteNames`.

thedavidmeister:
Closes #203.

## The rule

A deploy suite key is a NAME, or a name and a release TAG joined by an at sign:
the name lowercase letters and hyphens, the tag those plus digits and
underscores, neither half empty and at most one at sign.

Digits and underscores are legal ONLY after the at sign. That is what the keys
this repo generates need and all they need: `LibRainDeploySnapshot` emits a
released entry as the candidate key it was cut from, the at sign, and the record
directory's tag, which `isTag` already holds to `X_Y_Z` — so
`address-registry@0_1_10` passes as it stands and the hand written
`address-registry` it was cut from passes as it stands.

The rule is held against the WHOLE key rather than against a candidate key with
the released form derived from it, because `releasedSuites()` declares finished
keys: `allSuites()` is the one place every key a repo declares is read, and a
rule that only knew about the candidate half could not be asked about a released
entry at all.

The test fixtures spelled their released key `address-registry-0-0-1`, a shape
`LibRainDeploySnapshot` has never emitted and the rule now refuses, so they move
to `address-registry@0_0_1` — the generated shape. That is most of the diff's
line count and none of its behaviour.

## What was wrong

`UnknownDeploymentSuite` carries the declared keys "because the whole point of a
registry is that the answer is not one hardcoded string the caller has to
already know". `suiteNames()` builds that list by joining the registry on `", "`
and nothing constrained what a key may contain, so the rendering was not
injective: the TWO suite registry keyed `a,b` and `c` renders `a,b, c`, which is
exactly what the THREE suite registry keyed `a`, `b` and `c` reports. A reader is
told a different number of suites exist than do, and `b` — declared nowhere — is
handed to them as a valid key.

## Code or docs

The code. The NatSpec is the intent oracle and it is right: a list a caller
cannot read back as the set of keys it names IS the hardcoded string the registry
exists to replace, just spelled differently.

The refusal lives in `allSuites()`, in the pairwise pass that already refuses a
duplicate: both the deploy side and the verify side read the registry through
it, so every reader refuses at once and `suiteByName` can never answer a failed
lookup with an ambiguous list.

## Reconciled with the empty-key rule, not stacked on it

#231 (issue #202) landed `EmptyDeploySuiteKey(uint256 index)` in this same pass
while this PR was open. An alphabet that admits no zero-byte key already refuses
the empty key, so this does not add a second rule beside it: the two are one
`InvalidDeploySuiteKey(uint256 index, string suite)`, and the empty-key
reasoning moves into its NatSpec unchanged — the empty string is the value
`RainDeployBroadcast.run()` substitutes for an absent `DEPLOYMENT_SUITE`, so it
is the one refusal here with a broadcast behind it.

The error keeps #231's index, for #231's reason: an empty key names nothing, so
the position in `allSuites()` is the only thing that can. It gains the key
itself, which is what a reader needs for every refusal that is not the empty
one. `testEmptySuiteKeyReverts` is kept, asserting the same index at the same
reader set, under the error that now refuses it.

## Migration

Every key a consumer declares is now held to the alphabet, on every reader —
the deploy script, the verification tests and the chain matrix alike — with
`InvalidDeploySuiteKey(<index>, <key>)` at declaration read time, which cannot
reach a broadcast.

Measured rather than assumed: the 40 distinct keys declared across every repo in
the org that declares one — `rain.extrospection.deploy`, `rain.factory.deploy`,
`rain.math.float.deploy`, `rain.metadata.deploy`, `rain.tofu.erc20-decimals.deploy`,
`rainlang.deploy` and `raindex` — were extracted from a fresh clone of each and
held against the rule. Zero are rejected. So are this repo's own: the six
`<name>@<tag>` released keys under `src/lib/Lib*Released.sol` and the two
candidate keys in `RegistryDeploySuites`.

## Suite

`nix develop -c forge test -j 2` on the merged branch: 535 passed, 2 failed
(537 total). Both failures are `testFindDeployBlockZoltuFactory` and
`testIsStartBlockAtDeployBlock` in `test/src/lib/LibRainDeploy.t.sol`, and both
are the environmental Base archive read — `HTTP 403 ... Archive requests require
a personal token` from the allowed provider, before any assertion. They fail the
same way on unmodified `main`.

## QA

- Discriminating tests: `testSuiteKeyAlphabetRefuses`, `testSeparatorKeyIsRefused`,
  `testEmptySuiteKeyReverts` - each fails on base, verified two ways. (1) The base
  VALUES were pinned against unmodified `main` (`feefc95`) in a separate fresh
  clone: with the `a,b` / `c` fixture, `externalAllSuites().length == 2`,
  `externalSuiteNames() == "a,b, c"` and `externalSuiteByName("b")` reverting
  `UnknownDeploymentSuite("b", "a,b, c")` — a phantom key answered as valid — all
  green there; throwaway test deleted afterwards. (2) Base behaviour was then
  reproduced in place as mutant M12 below, which makes `checkSuiteKey` accept
  every key: all three tests fail under it. `testSuiteKeyAlphabetAccepts` is
  deliberately NOT in this list — it is the half that pins which keys the rule
  must not refuse, so it passes on base by construction and is here to stop the
  rule being narrowed into refusing the repo's own generated release keys.
- Mutations applied: 12 mutants over `src/abstract/RainDeploySuitesBase.sol`, all
  KILLED (`mutation-probe`, baseline green at 16 passed).
  `if (i == 0 || tagStart != 0)` -> drop the `i == 0` half (empty name admitted)
  -> `testSuiteKeyAlphabetRefuses`; same line -> drop the `tagStart != 0` half
  (two at signs admitted) -> `testSuiteKeyAlphabetRefuses`;
  `nameAlphabet` -> hyphen dropped -> `testEverySuiteIsSelectableByKey`,
  `testEmptySuiteKeyReverts`, `testSameLengthKeysSelectApart` and two more;
  `!(nameAlphabet || (tagStart != 0 && tagAlphabet))` -> `!(nameAlphabet || tagAlphabet)`
  (tag alphabet read over the whole key) -> `testSuiteKeyAlphabetRefuses`;
  `if (tagStart == key.length)` -> `if (false)` (empty-half guard removed) ->
  `testEmptySuiteKeyReverts`, `testSuiteKeyAlphabetRefuses`; same line ->
  `if (key.length == 0)` (narrowed to #202's check, trailing at sign admitted) ->
  `testSuiteKeyAlphabetRefuses`; `revert InvalidDeploySuiteKey(index, suite)` ->
  `(0, suite)` on the alphabet path -> `testSuiteKeyAlphabetRefuses`, and on the
  empty-half path -> `testEmptySuiteKeyReverts`, `testSuiteKeyAlphabetRefuses`;
  `if (char == "@")` -> `if (false)` (at sign not recognised as the separator) ->
  `testSuiteKeyAlphabetAccepts`, `testEverySuiteIsSelectableByKey` and three more;
  `checkSuiteKey(i, suites[i].suite)` in `allSuites()` -> deleted ->
  `testSeparatorKeyIsRefused`, `testEmptySuiteKeyReverts`; same line -> guarded to
  `i < released.length` (released half only) -> `testEmptySuiteKeyReverts`; M12,
  `checkSuiteKey` body -> `return;` (base behaviour, every key accepted) ->
  `testSeparatorKeyIsRefused`, `testEmptySuiteKeyReverts`,
  `testSuiteKeyAlphabetRefuses`.
- Oracle: `UnknownDeploymentSuite`'s own NatSpec — the list exists so a caller who
  does NOT already know the valid keys is told them — which fixes the expected
  behaviour as "the rendering must be injective", independent of any
  implementation. The accepting half of the table is oracled by what the
  declarations in this repo and its consumers actually spell: the emission shape
  in `LibRainDeploySnapshot` (`<candidate key>` + at sign + `tagForRecordPath`,
  already held to `X_Y_Z` by `isTag`) and the 40 consumer keys in Migration above,
  neither derived from the check under test.
- Category check: the issue asks for (A) the non-injective rendering, (B) the
  reader being told a different number of suites than exist, and (C) a phantom key
  handed back as valid. All three are one observable of one declaration and are
  covered together by `testSeparatorKeyIsRefused`, which asserts the refusal at
  `externalAllSuites`, `externalSuiteNames`, `externalSuiteByName("c")` (the key
  that IS declared) and `externalSuiteByName("b")` (the phantom the rendering
  invented) — the base values for all of which were pinned on `main` as above. The
  issue's own triage question, code or docs, is answered under "Code or docs".


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

PR #226, opened 2026-09-15
Generate the rolling snapshots under the record root they are frozen from

### LINKED ISSUE #206: cutRelease under an overridden recordRoot freezes from a record the regeneration never wrote, after writing the real one
## Unit and file

`regenerateSnapshots()` in `script/Build.sol` lines 86-97, against `cutRelease()` and `recordRoot()` in `src/abstract/BuildScript.sol` lines 34-56.

## The intent oracle

`BuildScript.recordRoot` NatSpec:

> The record root a release is frozen into, and the one the rolling snapshots are read from.
>
> Overridable so a `cutRelease()` can be exercised against a record other than the repo's own, which is append-only and cannot hold a test's release.

`testRecordRootDefaultsToTheRepoRecord` narrows it: the root is overridable only so a release can be cut somewhere a test may leave one.

Against that, `LibRainDeploySnapshot.writeSnapshot` NatSpec says the capability `Build` would need does not exist and is not meant to:

> There is no output root to choose ... `freeze` does take a root and that is not the same freedom: it COPIES, within one record tree ... GENERATING this repo's record anywhere but under `LIB_FS_ROOT` remains something nothing here can express.

## The property violated

`cutRelease()` regenerates and then freezes from `recordRoot()`. `Build.regenerateSnapshots` calls the five-argument `writeSnapshot`, whose path helper is the two-argument `pathForSnapshot` and therefore always `LIB_FS_ROOT`, while `freeze` reads each rolling snapshot at `pathForSnapshot(recordRoot(), CANDIDATE, name)`. So under an overridden root the regeneration writes to one tree and the freeze looks in another, and the seam `recordRoot` advertises cannot work for any `Build` written the way the README tells a downstream repo to write one. The asymmetry is visible inside `Build` itself: `regenerateLibs` DOES pass `recordRoot()` through to `writeReleasedSuitesLib`, and `regenerateSnapshots` has no way to.

Sharper form, and the part that is a hazard rather than a doc conflict: a subclass that points `recordRoot` at a fixture PRECISELY to keep its hands off the append-only real record still rewrites `src/generated/candidate/` before the revert. The side effect lands in the tree the override existed to avoid.

Re-verified against current `main` (e9a922a) in a fresh clone.

A `contract BuildRootHarness is Build` overriding `recordRoot()` to a fixture path under `test/`, with a test calling `harness.cutRelease()`:

`[FAIL: NothingToFreeze(test/generated-build-root/candidate/AddressRegistry.sol)]`

and the write did land in the real record: `src/generated/candidate/AddressRegistry.sol` and `MigrationRegistry.sol` mtimes advanced from 20:10:11 to 20:19:01 across the run (md5 unchanged, the candidate being current), while `test/generated-build-root/` was never created. Harness deleted afterwards; tree clean.

## Triage

For intended: `Build` is this repo's own concrete script and never overrides `recordRoot`, so the real lifecycle is unaffected. `recordRoot`'s NatSpec speaks about where a release is FROZEN into and READ from, never about where regeneration WRITES, so an implementor of `regenerateSnapshots` arguably owns honouring it — and `BuildScriptHarness` does exactly that, deriving its rolling path from `recordRoot()` and writing there itself. The library boundary is deliberate and argued in `writeSnapshot`: a copier may be pointed at a tree of its own, a generator may not. Failure is loud, names the path, and corrupts nothing.

For bug: the advertised seam is unreachable for the shape the README prescribes, which is a documented capability that the documented way of building does not have; and the fixture-root subclass, the one case the override exists for, still writes into the real record on its way to the revert.

Note: the independent verification pass reproduced this exactly, confirmed the real-record write by mtime, and returned AMBIGUOUS — it reads as a conflict between two NatSpecs rather than a behavioural defect. Its framing of the decision, which is the one to make here: either narrow `recordRoot`'s NatSpec to say the override is exercisable only by a `BuildScript`-level harness supplying its own regeneration, or give `writeSnapshot` a root and pass `recordRoot()` through `Build.regenerateSnapshots` the way `regenerateLibs` already passes it to `writeReleasedSuitesLib`.

thedavidmeister:
Closes #206.

## What was wrong

`cutRelease()` regenerates and then freezes from `recordRoot()`, but
`Build.regenerateSnapshots` called the `writeSnapshot` whose only output root
was `LIB_FS_ROOT` — its path helper was the two-argument `pathForSnapshot`, by
way of `LibFs.buildFileForTaggedContract`. Under an overridden root the
regeneration wrote one tree and `freeze` read another, so the seam
`BuildScript.recordRoot` advertises could not work for any `Build` written the
way this repo's own is: `NothingToFreeze(<fixture root>/candidate/<Name>.sol)`,
after `src/generated/candidate/` — the append-only tree the override exists to
keep a caller's hands off — had already been rewritten.

The code was wrong, not the documentation. `writeSnapshot`'s NatSpec argued the
capability did not exist ("GENERATING this repo's record anywhere but under
`LIB_FS_ROOT` remains something nothing here can express"), but `LibFs` exposes
`buildFileForContract`, which takes the directory it writes into, so the root
was always expressible; and the asymmetry was already visible inside `Build`,
where `regenerateLibs` passes `recordRoot()` through to
`writeReleasedSuitesLib` and `regenerateSnapshots` had no way to.

## What changed

`writeSnapshot` takes the record root, on both arities, required rather than
defaulted for the reason `freeze` and `writeReleasedSuitesLib` require theirs.
It writes through `LibFs.buildFileForContract` at `dirForSnapshot(root, dir)`
and returns `pathForSnapshot(root, dir, contractName)`;
`Build.regenerateSnapshots` passes `recordRoot()`, the way `regenerateLibs`
already does. The NatSpec paragraph that described the missing capability is
replaced by why the root is there.

The root is a CHECKED parameter. `dirForSnapshot(root, dir)` is the one place
the root becomes a path, and it already held `dir` to `LibFs`'s tag alphabet, so
before this change it checked the half of the path that names a directory inside
the tree and concatenated the half that decides where the tree IS. A `dir` held
to an alphabet beside a root taken verbatim is the shorter half of the path
confined: `../`, a leading `/`, a trailing one and the empty string all reached
the writer, and `fs_permissions` — a consuming repo's config rather than an
argument this library gets to see, and one that grants `./src` — was the only
thing left between a caller's string and a write outside the tree it named.

`requireRecordRoot` is that check, with `InvalidRecordRoot(root)` of its own. A
root is a PATH, so `LibFs.requireTag` cannot be asked it — the separators that
make a path a path are exactly what `requireTag` refuses — but the alphabet
BETWEEN the separators is `requireTag`'s, widened by exactly one character,
because that is how roots are spelled in this repo: `src/generated` is tag
segments already, and every fixture root the tests build (`test/generated-x`,
`test/generated-write-snapshot-root`, `test/fixture-record`, and the rest) is
tag segments plus `-`. So a root is one or more `/`-separated segments, each
non-empty and drawn from that alphabet. The empty segment is what carries the
rule from a segment to a path: refusing it refuses the leading separator of an
absolute path, the trailing one, the doubled one and the empty root. Refusing
`.` is what refuses `.` and `..`, for the reason `requireTag`'s own NatSpec
gives for a tag.

Two call sites reach it, and between them they are every way a root enters this
library: `dirForSnapshot(root, dir)`, which every writer and every rooted path
goes through (`pathForSnapshot`, `writeSnapshot` on both arities, `freeze`), and
`frozenSnapshotPaths(vm, root)`, the record WALK, which is the one root-taking
entry point that does not. The walk needs its own because it answers a missing
root with an empty record — a real state for a repo that has released nothing,
and silence for a root nothing could ever have been written under. The
read-side functions that take a `recordRoot` (`recordPathsForContract`,
`newestFrozenTag`, `checkReleaseFollowsRecord`, `writeReleasedSuitesLib`) all
reach the walk, so they are covered by it rather than by a check restated at
each.

At `LIB_FS_ROOT` the directory is byte-for-byte the one `LibFs.dirForTag` names,
so this repo's own `run()`/`cutRelease()` write exactly where they did: the
committed record is unchanged by this PR, and the suite's own green run leaves
`src/generated/` untouched.

## Migration

Breaking for a downstream `Build`: a call of the form

```solidity
LibRainDeploySnapshot.writeSnapshot(vm, CANDIDATE, name, creationCode, dependencies);
```

becomes

```solidity
LibRainDeploySnapshot.writeSnapshot(vm, recordRoot(), CANDIDATE, name, creationCode, dependencies);
```

Passing `LibRainDeploySnapshot.LIB_FS_ROOT` instead of `recordRoot()` keeps the
old behaviour, and is what the tests that build fixtures in the real record pass.

A downstream `recordRoot()` override, or any other root handed to this library,
must now be a path of those segments. `LIB_FS_ROOT` is one, so a repo that never
overrode the root is unaffected. A repo that overrode it to something with a
`.`, a leading or trailing `/`, or a character outside the alphabet now gets
`InvalidRecordRoot` where it previously got a path built out of it.

## Relation to the other open branch on this file

#232 (`fix-190`) adds a `setUp` that clears the fixture roots
`LibRainDeploySnapshotTest` writes, from a single `fixtureRoots()` list, and is
being reworked to one fixture root. It conflicts with this PR in
`test/src/lib/LibRainDeploySnapshot.t.sol` — both add constants and tests to
that contract. The resolution is additive in both directions: whichever lands
second takes the other's constants, and the fixture roots this PR adds
(`ROOTED_FIXTURE_ROOT`, and the escape fixture's `ESCAPE_FIXTURE_CLIMB_DIR` and
`ESCAPE_FIXTURE_DIR`) belong in that PR's clear list. There is no source
conflict: #232 changes no source.

## QA

- Discriminating tests: `testWriteSnapshotRefusesARootThatClimbsOutOfTheTreeItNames`,
  `testRecordRootRefusesEveryRootThatIsNotOne`,
  `testRecordRootSegmentIsTheTagAlphabetPlusHyphen`,
  `testFrozenSnapshotPathsRefusesARootThatIsNotOne`, all in
  `test/src/lib/LibRainDeploySnapshot.t.sol`, alongside
  `testRegenerateSnapshotsWritesUnderTheRecordRoot`
  (`test/script/BuildRecordRoot.t.sol`, over `test/concrete/BuildRecordRootHarness.sol`)
  and `testWriteSnapshotWritesUnderTheRootItIsHanded` for the root parameter
  itself. `testWriteSnapshotRefusesARootThatClimbsOutOfTheTreeItNames` names no
  new symbol, so it RUNS on the base commit: applied alone to `fix-206` at
  `324c57b` it is
  `[FAIL: the write landed in the real record, outside the root it was handed]`
  — the root `test/generated-escape-root/../../src/generated` walked the write
  out of the tree it named and into the append-only record. The other three name
  `InvalidRecordRoot` or `requireRecordRoot` and cannot compile on base, so they
  are verified by mutation below.
  `testRegenerateSnapshotsWritesUnderTheRecordRoot` fails on `main` at `baa1a9c`
  with only its test files applied:
  `[FAIL: regeneration wrote nothing under the record root: AddressRegistry]`,
  and that base run advanced `src/generated/candidate/{AddressRegistry,MigrationRegistry}.sol`
  from 14:44:25 to 15:36:07 with their md5s unchanged, which is the issue's
  "writes the real record on the way to the revert" reproduced.
- Mutations applied: six, all over `src/lib/LibRainDeploySnapshot.sol`, each run
  as the whole suite minus `LibRainDeployTest` — the fork-dependent contract
  whose two endpoint failures are about no code in this file — against a
  baseline of 439 passed, 0 failed. 6/6 killed, 0 survived, 0 no-run.

  | # | Mutation | Verdict |
  | --- | --- | --- |
  | M01 | `dirForSnapshot` drops `requireRecordRoot(root)`, keeping `LibFs.requireTag(dir)` — the pre-PR state, a checked `dir` beside a verbatim root | KILLED |
  | M02 | `frozenSnapshotPaths` drops `requireRecordRoot(root)` — the walk left as the one root-taking entry point with no check | KILLED |
  | M03 | `requireRecordRoot` stops refusing an empty segment before a `/`, admitting the leading separator and the doubled one | KILLED |
  | M04 | `requireRecordRoot` stops refusing an empty final segment, admitting the trailing separator and the empty root | KILLED |
  | M05 | `-` is dropped from the root alphabet, narrowing it to `LibFs.requireTag`'s exactly | KILLED |
  | M06 | `.` is admitted to the root alphabet, so `.` and `..` become legal segments | KILLED |
- Oracle: for the root parameter, `BuildScript.recordRoot`'s NatSpec and
  `freeze`, which reads each rolling snapshot at
  `pathForSnapshot(root, CANDIDATE, name)` — the expected path is the freeze's
  own spelling rather than the writer's, so the two are compared rather than one
  restated; the expected BYTES are the committed `src/generated/candidate/<Name>.sol`
  for the `Build` test and the same snapshot written at `LIB_FS_ROOT` for the
  library test, both independent of the new writer. For the root RULE, the
  alphabet's oracle is `LibFs.requireTag` itself: the widened copy is compared
  against it over all 256 byte values by
  `testRecordRootSegmentIsTheTagAlphabetPlusHyphen`, so what is asserted is the
  original's answer and not a restatement of the copy's, and `-` is asserted as
  the one byte they must DISAGREE about. The shape rule's oracle is
  `requireTag`'s own stated reason for its alphabet (no separator, no `.`, so no
  segment is `.` or `..`), applied to a path instead of a segment; the escape
  test's expected-absent path is `pathForSnapshot(dir, contractName)` — the real
  record's own spelling — rather than a literal.
- Category check: #206 asks for one of two remedies — narrow `recordRoot`'s
  NatSpec, or give `writeSnapshot` a root and pass `recordRoot()` through
  `Build.regenerateSnapshots`. The second is taken, because the capability the
  NatSpec called inexpressible is expressible (`LibFs.buildFileForContract`), so
  narrowing the doc would have documented a defect. Both halves of the issue are
  covered: the unreachable seam (a release now freezes from the record its own
  regeneration wrote, asserted by the library test's path and by the `Build`
  test's bytes) and the side effect (nothing lands in the real record — asserted
  directly by the library test's `assertFalse`, and by the suite leaving
  `src/generated/` clean on the branch where the base run did not). The root
  check is not a third remedy: it is what makes the new input an argument rather
  than a hole, and it is asserted over the same real record the issue is about.

## Verification

`git clean -fdq test src && rm -rf fixture-lib cache/fuzz` before every run,
`.env.example` with drpc/publicnode endpoints substituted for arbitrum, base,
bsc and hyperevm.

- Base `fix-206` at `324c57b`: 512 passed, 2 failed of 514.
- This branch: 516 passed, 2 failed of 518 — the four new tests, and the same
  two failures.
- The two are `testFindDeployBlockZoltuFactory` and `testIsStartBlockAtDeployBlock`
  in `LibRainDeployTest`, both
  `403 ... Archive requests require a personal token` from the public base
  mainnet endpoint. Per `CLAUDE.md` a fork instantiation failure is an endpoint
  fault, not a statement about a deployment; #232 reports the same pair from the
  same endpoint. drpc's base endpoint serves archive but rate-limits: it answers
  429 and fails 16 of the same suite at `-j 1`, so it is not the better of the
  two. Every other suite is green on both sides.


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

**This branch.** merge-base `baa1a9ca5821c6b41bd148fecc4fd87bd0f3666c`. This PR does touch `LibRainDeploySnapshot.t.sol` (+247/-3), but both functions are byte-identical to the merge-base; only their line numbers move, 599/611 to 825/837. Both functions came in with

PR #227, opened 2026-09-15
Move the foundry.toml config check onto RainDeployVerifySnapshotBase

### LINKED ISSUE #194: Nothing in the suite can make the foundry.toml config assertions fail
## Unit and file

`testSupportedNetworksAreFullyConfigured` in `src/abstract/RainDeployVerifySnapshot.sol` -- the whole config group, which is inline in a test body that takes no arguments and reads the real `foundry.toml`. Its only subject anywhere in the repo is `test/src/abstract/RegistryDeployVerify.t.sol`, a one-line binder over this repo's real declaration.

Verified against a fresh read-only clone of `main` at `e9a922a845056c95a9fd38bdf8fbe674db1ece6e` with the oracle `dependencies/` copied in.

## The intent oracle

The NatSpec on the test makes an enforcement claim:

> This is what makes the `[etherscan]` half enforced at all. The RPC half is enforced only incidentally, by the fork tests, and only forwards.

and states the group's job as failing on the PR that drifts:

> so this needs no RPC and fails on the PR that drifts rather than at dispatch time.

The sibling record group is deliberately shaped the other way: `checkFrozenSnapshotsReleased(string[] memory paths, DeploySuite[] memory released)` in `src/abstract/RainDeployVerifySnapshotBase.sol` takes its subject as arguments, and `RainDeployVerifySnapshotBaseTest` drives it at every position and every shape of declaration. The repo already ships the pattern that would close this.

## The property violated

An enforcement claim with no demonstration behind it. Seven behaviours in this unit are assertions that nothing in the suite has ever driven to a failure, so none of them is known to be capable of failing. The two reverse loops (`LibStringSet.holds` over `vm.parseTomlKeys`) in particular have never been driven to false by anything: if `vm.parseTomlKeys` began returning an empty array for a missing section instead of reverting, both loops would go inert, every run would stay green with no subject, and nothing would notice. That is the exact inert-check failure shape the record group's own NatSpec spends two paragraphs guarding against.

## The verified repro, and what it does now

Neuter the `[etherscan]` key-existence assertion -- replace its `vm.keyExistsToml(config, string.concat(.etherscan., networks[i]))` argument with the literal `true` -- and run the suite:

```
nix develop -c forge test
```

Result now: byte-identical to the unmutated baseline in this environment (`Encountered a total of 61 failing tests, 398 tests succeeded`, the 61 being fork tests that need RPC env vars absent locally). The mutant SURVIVES. The original campaign reports the same against its own green 397-test probe baseline, along with M07 through M13 all SURVIVED.

The structural claim was independently confirmed: `grep` over `test/` finds no occurrence of `etherscan` in any `.sol` test file, and the only contract that reaches this test is the real binder, so no assertion in this test has ever been observed to fail.

Also unasserted, and folded in here rather than filed separately:

- A position-versus-membership mutation on the `[rpc_endpoints]` reverse loop cannot be discriminated today at all, because `[rpc_endpoints]` happens to list the nine aliases in `supportedNetworks()` order, so the two agree element for element.
- A supported network whose name contained a dot would make `string.concat(.rpc_endpoints., name)` query a nested key rather than the alias. No current network is affected and nothing asserts it stays that way.

The file was restored; the clone tree is clean and nothing was committed or pushed.

## Triage

For it being a defect: the group's whole value is that it fails on the PR that drifts, and nothing demonstrates that it fails on anything. Nothing in the NatSpec or README defends the group's undrivability. What the NatSpec defends is that the FILE READ must be the binder's own (`vm.readFile` resolves against the project root of whatever runs it), which is true of the test but says nothing about an internal `check(config, networks)` that the test would call with its own read -- precisely the sibling shape. The record group's anti-seam argument is aimed narrowly at the ROOT, not at parameterisation as such, so it does not transfer here.

For it being intended: the check must read the binder's own `foundry.toml`, which is why it is bound where the binder is, and factoring the body into an internal function adds surface to a contract whose NatSpec argues hard that every optional-looking seam is a way to make a check inert. The one thing it buys is testability rather than a new guarantee.

Note -- independent verdict: **bug**, in the coverage sense this campaign is for, at moderate rather than high severity: on the real binder the assertions do execute against the real file every run, so they are unproven rather than dead. The sharpest version of the complaint is the NatSpec's own enforcement claim, which has no demonstration behind it anywhere in the suite. The proposed fix is the sibling's shape -- an internal checker over (config, networks) plus synthetic TOML fixtures -- which is the pattern the repo already ships.

thedavidmeister:
Closes #194.

`testSupportedNetworksAreFullyConfigured` claimed, in its own NatSpec, to be
"what makes the `[etherscan]` half enforced at all" and to "fail on the PR that
drifts". Nothing in the suite had ever seen any of its seven assertions fail:
the whole check was inline in a test body that takes no arguments and reads the
real `foundry.toml`, so its only subject anywhere was this repo's own config,
which is correct. An enforcement claim with no demonstration behind it.

Fixed the way the repo already fixes this: the sibling record group keeps its
CHECK on `RainDeployVerifySnapshotBase`, taking its subject as arguments, and
its BINDING as a one-line test on `RainDeployVerifySnapshot`. The config group
now has the same shape.

## What changed

`checkConfigSectionNamesNetworks(config, section, networks)` and
`checkNetworksFullyConfigured(config, networks)` move the check onto the base,
where `RainDeployVerifySnapshotBaseTest` drives it over synthetic TOML.
`testSupportedNetworksAreFullyConfigured` is now the binding and nothing else —
it still spells `vm.readFile("foundry.toml")` and
`LibRainDeploy.supportedNetworks()` itself, so there is still nothing a consumer
can point somewhere quieter.

Two behaviour changes fall out of that, both of them fixes:

**The forward direction now reads the section's KEYS rather than probing a key
path.** `vm.keyExistsToml(config, ".rpc_endpoints." + network)` asks a PATH
question, and TOML reads an unquoted dot as nesting. So for a network named
`sub.chain`:

| config | main | this PR |
| --- | --- | --- |
| `"sub.chain" = "..."` — the alias, correctly spelled | FAILS | passes |
| `sub.chain = "..."` — a nested table, no such alias | passes | FAILS |

Both answers on main are backwards. No supported network carries a dot today,
which is exactly why nothing was holding that.

**A section that is absent entirely now fails as `ConfigSectionMissing`.** On
main that case depended on `vm.parseTomlKeys` reverting — the cheatcode
behaviour the issue names as the inert-check hazard: were it ever to return an
empty array instead, both reverse loops would walk nothing and the suite would
stay green with no subject. Reading the keys once and driving both directions
off them means an empty section fails the forward direction per network, so
there is now no shape of section this passes without a subject.

The three `assertTrue` messages become three errors —
`ConfigSectionMissing(section)`, `NetworkNotConfigured(section, network)`,
`ConfigEntryNotSupported(section, entry)` — matching
`FrozenSnapshotNotReleased`'s shape so the negative cases assert exact values
rather than a bare failure.

## Migration

A consumer inheriting `RainDeployVerify` needs no change: same test, same name,
same subject, and this repo's own `foundry.toml` passes unchanged. What moves is
the failure OUTPUT — a drifted consumer now sees a typed revert naming the
section and the key instead of an assertion message. The three error symbols are
new exports of `RainDeployVerifySnapshotBase.sol`.

## QA

- Discriminating tests: `testNetworkConfigNetworkNameContainingADot`,
  `testNetworkConfigIsMembershipNotPosition`, `testNetworkConfigEmptySectionReverts`,
  `testNetworkConfigMissingSectionReverts`, `testNetworkConfigMissingRpcAliasReverts`,
  `testNetworkConfigMissingEtherscanKeyReverts`, `testNetworkConfigExtraRpcAliasReverts`,
  `testNetworkConfigExtraEtherscanKeyReverts`, `testNetworkConfigExactSectionsPass`
  - `testNetworkConfigNetworkNameContainingADot` fails on base's own expression:
    mutant M08 below restores main's `vm.keyExistsToml(config, string.concat(path, ".", networks[i]))`
    verbatim and that test, alone of the 38, kills it. The other eight drive
    `checkNetworksFullyConfigured`, which base does not have to call — base's
    check is unreachable from any test by construction, which is the defect —
    so their discrimination is the M01-M12 matrix below rather than a base run.
- Mutations applied: `mutation-probe` over `src/abstract/RainDeployVerifySnapshotBase.sol`,
  12/12 KILLED, 0 survived, against a green 38-test baseline
  (`RainDeployVerifySnapshotBaseTest` + `RegistryDeployVerifyTest`, fork test excluded):
  - M01 `if (!vm.keyExistsToml(config, path))` -> `if (false)` -> `testNetworkConfigMissingSectionReverts`
  - M02 forward `!entrySet.has(...)` -> `false` -> `testNetworkConfigMissingRpcAliasReverts`, `testNetworkConfigMissingEtherscanKeyReverts`, `testNetworkConfigEmptySectionReverts`, `testNetworkConfigNetworkNameContainingADot`
  - M03 reverse `!networkSet.has(...)` -> `false` -> `testNetworkConfigExtraRpcAliasReverts`, `testNetworkConfigExtraEtherscanKeyReverts`
  - M04 forward loop `i < networks.length` -> `i < 1` -> `testNetworkConfigMissingRpcAliasReverts`, `testNetworkConfigMissingEtherscanKeyReverts`
  - M05 reverse loop `i < entries.length` -> `i < 1` -> `testNetworkConfigExtraRpcAliasReverts`, `testNetworkConfigExtraEtherscanKeyReverts`
  - M06 `"etherscan"` call -> second `"rpc_endpoints"` call -> `testNetworkConfigMissingEtherscanKeyReverts`, `testNetworkConfigExtraEtherscanKeyReverts`, `testNetworkConfigMissingSectionReverts`, `testNetworkConfigEmptySectionReverts`
  - M07 `"rpc_endpoints"` call -> second `"etherscan"` call -> `testNetworkConfigMissingRpcAliasReverts`, `testNetworkConfigExtraRpcAliasReverts`, `testNetworkConfigMissingSectionReverts`, `testNetworkConfigEmptySectionReverts`, `testNetworkConfigNetworkNameContainingADot`
  - M08 forward membership -> main's `vm.keyExistsToml` key path -> `testNetworkConfigNetworkNameContainingADot` ONLY
  - M09 `NetworkNotConfigured(section, networks[i])` -> `(section, "")` -> `testNetworkConfigMissingRpcAliasReverts`, `testNetworkConfigMissingEtherscanKeyReverts`, `testNetworkConfigEmptySectionReverts`, `testNetworkConfigNetworkNameContainingADot`
  - M10 `ConfigEntryNotSupported(section, entries[i])` -> `(section, "")` -> `testNetworkConfigExtraRpcAliasReverts`, `testNetworkConfigExtraEtherscanKeyReverts`
  - M11 forward membership -> position compare -> `testNetworkConfigIsMembershipNotPosition`, `testNetworkConfigExtraRpcAliasReverts`, `testNetworkConfigExtraEtherscanKeyReverts`
  - M12 reverse membership -> position compare -> `testNetworkConfigIsMembershipNotPosition` ONLY
- Oracle: the TOML specification for what a dot in an unquoted key means, and
  foundry's own cheatcodes read directly rather than through the check —
  `testNetworkConfigNetworkNameContainingADot` asserts
  `vm.keyExistsToml(nested, ".rpc_endpoints.sub.chain")` is true and
  `vm.parseTomlKeys(nested, ".rpc_endpoints")[0]` is `"sub"` before it asks the
  check anything, so the expected answers are derived from the file format and
  not from the implementation. Every expected revert is an exact
  `abi.encodeWithSelector(...)` with section and key spelled as string literals,
  never as the constants the source uses. Fixture networks are `alpha`/`beta`/
  `gamma`, which no real config names, so no case can pass on this repo's own
  `foundry.toml` by accident.
- Category check: the issue asks (A) the seven config assertions are undrivable
  — an enforcement claim with no demonstration; (B) position-versus-membership
  on the reverse loops cannot be discriminated today because `[rpc_endpoints]`
  happens to be written in `supportedNetworks()` order; (C) a network name
  containing a dot would query a nested key rather than the alias. Covered A by
  moving the check to an argument-taking internal on the base and driving all of
  it from synthetic sections (M01-M10 all killed, including the two reverse
  loops the issue singles out and the empty-section inert shape); B by
  `testNetworkConfigIsMembershipNotPosition`, which passes two derangements of
  the network order and is the only killer of M12 and a killer of M11; C by
  reading the section's keys instead of probing a key path, with
  `testNetworkConfigNetworkNameContainingADot` pinning both directions and
  killing M08 alone.


thedavidmeister:
Superseded by #233. Generating `[rpc_endpoints]` and `[etherscan]` from `supportedNetworks()` leaves these membership loops nothing to compare — both sides come from one list, so an alias that is not a supported network cannot exist to be found.

Two findings from this branch are carried forward rather than dropped: the forward direction must read the section's parsed keys rather than probing a key path, because TOML reads an unquoted dot as nesting and `vm.keyExistsToml` answers backwards for a network name containing one; and an absent section must fail on its own rather than resting on `vm.parseTomlKeys` reverting.

Closing unmerged. #194 stays open until #233 lands.

PR #228, opened 2026-09-15
test: run regenerateLibs into a fixture dir and pin what it emits

### LINKED ISSUE #207: No argument, loop bound or call ordering inside Build's generator hooks is asserted anywhere
## Unit and file

`regenerateLibs()` and `regenerateSnapshots()` in `script/Build.sol` lines 63-97 — the two hooks `run()` and `cutRelease()` exist to call.

## The intent oracle

`BuildTest` NatSpec states the choice outright:

> Deliberately nothing here calls `run()` or `cutRelease()`. Both rewrite the committed `src/generated/` snapshots and `src/lib/` libs that other test contracts read, and forge runs test contracts in parallel — a contract rewriting what another one is reading is a race, not a check.

`LibRainDeploySnapshot.t.sol` names the consequence and the substitute: the alias emitter is otherwise the one generator path in this repo that nothing anywhere executes, because only `script/Build.sol` calls it and nothing may call that; and a check that regenerated the file first would pass on a stale one by overwriting it, which is why the committed OUTPUT is pinned instead.

## The property violated

Nothing inside either hook is asserted: not an argument, not a loop bound, not a call ordering. The substitute pins (`testTheCommittedAliasLibIsWhatTheGeneratorEmits`, `testTheCommittedReleasedLibIsWhatTheGeneratorEmits`, `testTheCommittedAggregateIsWhatTheGeneratorEmits`) are output-anchored, so they catch drift only AFTER someone re-runs the generator and commits the result — a wrong argument is invisible until then — and the alias and released-lib pins cover `AddressRegistry` alone; there is no text pin for `LibMigrationRegistryDeploy.sol` or `LibMigrationRegistryReleased.sol` against the emitters. Both hooks hard-code `LibRainDeploySnapshot.LIB_DIR` and `CANDIDATE`, so no test can drive them into a fixture directory: this is a seam that is missing, not a test that was skipped.

Re-verified against current `main` (e9a922a) in a fresh clone. Baseline `nix develop -c forge test`: 398 passed, 61 failed — every failure a missing `<NETWORK>_RPC_URL`, this sandbox having no endpoints.

- `revert();` as the first statement of `Build.regenerateLibs`: 398 passed, 61 failed.
- `revert();` as the first statement of `Build.regenerateSnapshots`: 398 passed, 61 failed.

Identical summary line in all three runs; delta zero. (The candidate reported 397 passed / 0 failed under `sh check.sh`, which supplies the RPC URLs. The delta is the claim and it holds either way.) `grep` over `test/` confirms nothing calls `Build`'s `run()` or `cutRelease()`; the only callers of those names in `test/` are `BuildScriptHarness` and unrelated deploy suites.

## Triage

For intended: the repo states the reason in two places rather than leaving it implicit, and the reason is a real race it has already been bitten by elsewhere. Widening `Build`'s hooks to take a lib dir or a root would add parameters to a script whose whole stated surface is `run()` and `cutRelease()`. `audit/mutation-test-scans.json` already carries `notExecutedBySuite` as a logged category, so this class is tracked rather than novel.

For bug: these two functions are the entire generator, and the substitute pins cover one of the two generated contracts, after the fact.

Note: the independent verification pass reproduced the zero delta exactly, checked each supporting claim by reading, and returned INTENDED. The residual it recorded, and the only part worth acting on, needs no new parameters on the script: extend the emitter pins to `MigrationRegistry`'s two committed libs, which closes the per-contract half of the gap without touching `Build`'s surface.

thedavidmeister:
Closes #207.

## What was wrong

`Build`'s two generator hooks were executed by nothing. A `revert()` as the
first statement of `regenerateLibs` left the suite at 512 passed / 0 failed on
`baa1a9c`; the same in `regenerateSnapshots` left it at 508 passed / 4 failed,
the four being public-RPC rate limits in `LibRainDeployTest` that are there
without any mutation. Zero delta both ways, re-verified on current `main` rather
than on the `e9a922a` the issue was filed against.

Every check on the hooks was output-anchored: it compared a committed lib
against the emitters, so it saw drift only AFTER somebody re-ran the generator
and committed what came out. A loop bound that stopped one contract short, or a
`constantPrefix` taken from `contracts[0]` on every pass, was invisible until
the next release cut it into the append-only record.

The issue's own recommended residual — extend the emitter pins to
`MigrationRegistry`'s two committed libs — has already landed, in `4eb5363`
(PR #176), which postdates the commit the issue was verified against. That half
is closed; this PR closes the half the issue's title names.

## What changed

`regenerateLibs` takes its directory from a new `internal virtual libDir()`
instead of hard-coding `LibRainDeploySnapshot.LIB_DIR`, for exactly the reason
`BuildScript.recordRoot` is already overridable: a hook that can only be pointed
at the committed tree can only be RUN by overwriting files the rest of the suite
compiles and reads, and forge runs test contracts in parallel. No parameter is
added to `run()` or `cutRelease()`, and nothing outside `Build` can see the
hook.

`BuildHarness` gains the fixture directory, an `externalRegenerateLibs()` that
refuses to run when it is left pointing at `src/lib`, and an `externalLibDir()`
so the production default is still assertable.

`regenerateSnapshots` gets no such seam and can have none: `LibFs.dirForTag`
confines every snapshot write to `src/generated/<tag>/`, which is the record
`frozenSnapshotPaths` walks, so there is no directory to drive it into that the
suites running beside it do not read.

## Migration

None. `libDir()` is `internal` with the previous constant as its default, so a
downstream `Build` that overrides nothing behaves exactly as before. A consumer
repo with its own `BuildHarness` copy must pass the constructor's new argument;
`""` reproduces the old behaviour.

## QA

- Discriminating tests: `BuildTest.testRegenerateLibsEmitsExactlyTheCommittedLibs`,
  `BuildTest.testTheDefaultLibDirIsWhereTheCommittedLibsAre`,
  `BuildTest.testRegenerateLibsRefusesToWriteTheCommittedLibs` - each fails on
  base because on base the hook cannot be run at all: the two mutations the
  issue names were applied to `main` itself and the whole suite stayed green
  (`revert()` in `regenerateLibs`: 512 passed, 0 failed; `i + 1 <
  contracts.length` in the same loop: 505 passed, 7 failed, every one of the 7
  a `vm.createFork` rate limit, which is the same kind of failure the unmutated
  baseline has and never an assertion). The same two mutations on this branch
  are killed, named below.
- Mutations applied (each reverted before the next; killing test named by its
  own failure message):
  - `script/Build.sol:78` `regenerateLibs` -> `revert();` as its first statement
    -> `testRegenerateLibsEmitsExactlyTheCommittedLibs` (`EvmError: Revert`)
  - `script/Build.sol:81` `i < contracts.length` -> `i + 1 < contracts.length`
    -> `testRegenerateLibsEmitsExactlyTheCommittedLibs` (`emitted a different
    number of files than the repo commits: 3 != 5`)
  - `script/Build.sol:83` `contracts[i].constantPrefix` ->
    `contracts[0].constantPrefix` ->
    `testRegenerateLibsEmitsExactlyTheCommittedLibs` (`emitted something other
    than the committed LibMigrationRegistryDeploy.sol`)
  - `script/Build.sol:83` `LibRainDeploySnapshot.CANDIDATE` -> `"0_1_10"` ->
    `testRegenerateLibsEmitsExactlyTheCommittedLibs` (`emitted something other
    than the committed LibAddressRegistryDeploy.sol`)
  - `script/Build.sol:72` `return LibRainDeploySnapshot.LIB_DIR;` -> `return
    "fixture-lib";` -> `testTheDefaultLibDirIsWhereTheCommittedLibsAre`
    (`fixture-lib != src/lib`)
  - `test/concrete/BuildHarness.sol:48` the empty-dir refusal deleted ->
    `testRegenerateLibsRefusesToWriteTheCommittedLibs` (`next call did not
    revert as expected`). That run also took
    `testEveryCommittedAliasLibIsWhatTheGeneratorEmits` down with it, reading
    `src/lib/LibAddressRegistryDeploy.sol` as empty while the unguarded write
    was in it - the race the refusal exists for, observed. `git status src/lib`
    was clean afterwards, which is the no-op oracle above holding in the one
    place it can be checked directly.
- Oracle: the committed tree, not the emitters. A regeneration of a clean
  checkout is a no-op, so every file `regenerateLibs()` writes must be
  byte-identical to the file already in `src/lib/`, and it must write no others.
  That is independent of the emitters in the way the existing pins are not: they
  say the committed files are what the emitters produce, and this says the hook
  asks the emitters for those files. `testTheDefaultLibDirIsWhereTheCommittedLibsAre`
  asserts the literal `src/lib` rather than the constant the default returns,
  so it is not the implementation compared against itself.
- Category check: the issue asks for an argument, a loop bound and a call
  ordering. Arguments: every argument of all three calls in `regenerateLibs` is
  covered, the three mutations above being `constantPrefix`, `CANDIDATE` and the
  directory; `contractName` and `candidate.snapshot` reach the file name and the
  emitted entries, and `recordRoot()` reaches the released lib's import block, so
  a wrong value in any of them changes the bytes compared. Loop bound: the file
  COUNT is asserted as well as the contents, because a short loop writes nothing
  wrong - it writes nothing at all - and an extra write is a generated file
  nothing regenerates. Call ordering: there is no ordering property inside
  `regenerateLibs` to assert, and asserting one would be theatre - `writeAliasLib`
  and `writeReleasedSuitesAggregate` read nothing, `writeReleasedSuitesLib` reads
  only the record, so no permutation of the three changes a byte anywhere. The
  ordering that IS observable, `regenerateSnapshots` before `regenerateLibs`, is
  already asserted by `BuildScriptTest.testRunRegeneratesAndFreezesNothing`
  through `BuildScriptHarness`'s markers. `regenerateSnapshots` itself stays
  undrivable for the `LibFs` reason above, which is a boundary in a pinned
  upstream package rather than a seam this repo declined to open.

`nix develop -c forge test --threads 1` on `baa1a9c`: 507 passed / 5 failed, all
5 being public-RPC rate limits in `LibRainDeployTest` and
`RainDeployBroadcastTest` (that suite passes 79/79 on its own when the endpoint
is not rate-limiting). On this branch: 515 passed / 0 failed, 512 + the three
new tests.


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

**This branch.** merge-base `de11ac01efc51e8e86a2910a07ce5525a358f6d6`. `LibRainDeploySnapshot.t.sol` is not in this PR's diff at all; the two functions sit at 599 and 611 on both sides, byte-identical. Both functions came in with

PR #229, opened 2026-09-15
Assert [etherscan] entries resolve and bind each chain id to its network

### LINKED ISSUE #192: The config check passes an [etherscan] entry that takes verification down for every network
## Unit and file

`testSupportedNetworksAreFullyConfigured` in `src/abstract/RainDeployVerifySnapshot.sol` (the `.etherscan.<network>` assertion, currently a `vm.keyExistsToml` existence check), bound for this repo by `test/src/abstract/RegistryDeployVerify.t.sol`. The config it reads is the `[etherscan]` section of `foundry.toml`.

Verified against a fresh read-only clone of `main` at `e9a922a845056c95a9fd38bdf8fbe674db1ece6e` with the oracle `dependencies/` copied in.

## The intent oracle

`foundry.toml`, in the comment that introduces `[etherscan]`, names this exact state and why it is worse than a missing key:

> `chain` is stated on the entries whose alias foundry does not itself resolve to a chain. An entry with neither `chain` nor `url` under such an alias is not a missing key, it is a config error -- At least one of `url` or `chain` must be present for Etherscan config with unknown alias -- raised while resolving the section, so it takes down verification for every network in it and not only its own.

The same comment states what the check is for:

> `rainix-manual-sol-artifacts` passes `--verify` by default and exports exactly these variable names, so a deploy without this section broadcasts and then fails with no API key configured for the chain -- after spending the gas.

The test NatSpec states the same purpose for the group: the drift must fail on the PR, and

> This is what makes the `[etherscan]` half enforced at all.

## The property violated

The group exists so that an `[etherscan]` section which cannot verify fails on the PR rather than after a broadcast. An entry carrying only a `key` under an alias foundry cannot itself resolve to a chain cannot verify -- and takes the whole section down with it -- yet it satisfies every assertion the check makes, because the assertion looks only at key existence and never at the entry contents.

## The verified repro, and what it does now

In the clone, drop `chain = 999` from the `hyperevm` line of `[etherscan]` in `foundry.toml`, leaving the entry with its `key` alone, then:

```
nix develop -c forge test --match-test testSupportedNetworksAreFullyConfigured
```

Result now:

```
[PASS] testSupportedNetworksAreFullyConfigured() (gas: 82962)
Suite result: ok. 1 passed; 0 failed; 0 skipped
```

Gas is identical to the unmutated baseline, which confirms the assertion never reads the entry contents. The config left behind is one that cannot verify on any of the nine networks. The independent check also confirmed the hazard is live rather than hypothetical: `forge verify-contract --chain hyperevm` in this toolchain rejects the alias outright, and foundry's enumerated alias list carries `hyperliquid` but no `hyperevm`, and `robinhood-testnet` but no `robinhood` -- so the entries carrying `chain` and `url` today carry them because foundry genuinely cannot resolve those aliases, exactly as the comment says.

`foundry.toml` was restored; the clone tree is clean and nothing was committed or pushed.

## Triage

For it being a defect: the failure mode is named in the repo's own words in two places, the config group is the only thing standing between it and a dispatch, and the entries that carry `chain` or `url` today prove the repo already knows which aliases foundry cannot resolve while nothing asserts it stays that way. The closing assertion is cheap -- every `[etherscan]` entry carries at least one of `chain` or `url` -- and it encodes no third-party data, is monotonic (adding `chain` to an alias foundry does resolve is harmless and correct), and cannot red-line when foundry adds an alias.

For it being intended: both places that name the hazard also, in the same breath, place it outside what is checked. The `foundry.toml` sentence exists precisely to distinguish the failure the key check catches from the one it does not, and the test NatSpec is explicit that the KEYS are the whole contract here. A declared scope is not a broken promise, and a shape assertion is new surface on a contract whose NatSpec argues hard against optional-looking seams.

Note -- independent verdict: **ambiguous**. The repro reproduces exactly as written. The verdict reads slightly against the original case-for: the docs place entry shape outside the declared scope, so this is not a contract violation. What keeps it off intended is that the original case-against does not survive contact -- the assertion that closes the hazard needs no knowledge of foundry's alias table at all, so the stated reason for leaving it unasserted does not hold and the real reason is a scope choice. This is a human call on whether the scope should widen, with the cheap closing assertion already justified in the docs twice.

thedavidmeister:
Closes #192.

## What was wrong

`testSupportedNetworksAreFullyConfigured` asserted that a `[etherscan]` key
EXISTS for every supported network and nothing about what the entry contains.
An entry carrying only a `key`, under an alias foundry does not itself resolve
to a chain, satisfied every assertion in the group and is config that cannot
verify anything — foundry resolves the SECTION rather than the one entry the
network being verified needs, so it raises

```
At least one of `url` or `chain` must be present for Etherscan config with unknown alias `hyperevm`
```

for whichever network `--verify` was pointed at. `foundry.toml`, the test
NatSpec and the README each named that failure mode and each placed it outside
what was checked.

The hazard is live, not hypothetical. With `chain = 999` dropped from the
`hyperevm` entry on `main`, `forge verify-contract --chain 137` (polygon, a
different network entirely) dies on that error, while with the entry intact the
same command resolves and reaches the explorer. Both runs on this toolchain
(foundry `43923a4`).

Stating `chain` on all nine entries then creates a second, quieter gap. Those
ids were read off `cast chain-id` once, by hand, and nothing re-reads them. A
wrong id resolves, reads as well-formed and passes every assertion in the
group, because every assertion in the group is about the TEXT — and `chain` is
what `--verify` submits, so the deployment is verified against another chain's
explorer, after the gas is spent. The same gap swallows the mirror case: an
`[rpc_endpoints]` alias bound to a different network than it names, which is
worse, because every chain-anchored assertion ever made through that alias was
made somewhere nobody named.

## What changed

- `[etherscan]` entries now state `chain` on all nine networks, not only on the
  four whose alias foundry cannot resolve. Chain ids were read off the repo's
  own `[rpc_endpoints]` with `cast chain-id`, and the four already stated (1,
  56, 999, 4663) came back matching, which is what validates the other five.
- The config group now asserts that every `[etherscan]` entry carries at least
  one of `chain` or `url` — foundry's own condition, spelled strictly so that it
  needs to know nothing about foundry's alias table. Stating the chain an alias
  already resolves to resolves it to the same chain, so the strict form is
  monotonic and cannot red-line when foundry adds or renames an alias.
- The group's assertions moved into `RainDeployVerifySnapshotBase` as
  `checkNetworksConfigured(config, networks)`, leaving
  `testSupportedNetworksAreFullyConfigured` as the file read that binds them.
  This is the split the base already documents for
  `checkFrozenSnapshotsReleased`: a check that takes its subject as an argument
  can be handed one a test builds. Without it, deleting the new assertion's call
  site was a mutation nothing killed — the check would have existed while
  nothing proved it was wired.
- **`testSupportedNetworkChainIdsAreBound`** in `RainDeployVerifyChain` forks
  every supported network that states a `chain` and compares the declared id
  against `block.chainid`, failing with `NetworkChainIdMismatch(network,
  declared, reported)`. Both ids are in the error because which of the two is
  wrong — the declaration, or the alias the endpoint is bound to — is not
  something the check can know, and they are opposite fixes.
- It is in the chain group rather than the config group because the subject is
  the endpoint, not the text. This is the contract that already forks, and
  keeping the comparison out of the snapshot half is what leaves that half
  bindable by a job with no RPC credentials.
- An entry resolving through a `url` alone states no id and is skipped, since
  the config group requires only one of the two of it. Every entry being that
  way is `NoDeclaredChainIds` rather than a check that passes having forked
  nothing.
- `foundry.toml`, the test NatSpec and the README passages updated: entry shape
  and the id binding are both inside the declared scope now, so the places that
  said the keys were the whole contract no longer say it, and the group table
  says what the chain group catches.

Two consumer-visible notes. A consumer repo whose `[etherscan]` section has an
entry with neither `chain` nor `url` goes red on bumping, which is the point;
the fix is to add `chain = <id>` to that entry. And `RainDeployVerifyChain` now
reads the binder's own `foundry.toml`, so the `{ access = "read", path =
"./foundry.toml" }` entry in `fs_permissions` that the snapshot half already
needed is needed by the chain half too — a repo binding `RainDeployVerify` has
it already.

## Fork budget

Measured rather than assumed, because the new test is inherited rather than
declared once. `testSupportedNetworkChainIdsAreBound` creates nine forks per
concrete contract that inherits `RainDeployVerifyChain`, and four do
(`RainDeployVerifyChainTest`, `RainDeployVerifyChainCandidateTest`,
`RainDeployVerifyChainEmptyTest`, `RegistryDeployVerifyTest`): 36 fork
creations, plus 3 in the negative cases, for 39 more than base.

That is not visible in the wall clock: whole-suite runs against the same
endpoints came in at 66s, 66s and 72s on base and 64s, 65s and 112s here, which
is endpoint variance rather than a trend. It does not move the flake floor
either, and the floor is not zero. `rpc.hyperliquid.xyz` intermittently answers
`invalid block height` for a block it has just reported as latest, which
`vm.createFork` surfaces as a fork that cannot be instantiated — twice in three
base runs (`testChainNotDeployedRevertsForALaterSuite`,
`testChainMatrixCoversEverySupportedNetwork`) and once in four here
(`testSuitesLiveOnEverySupportedNetwork`), all three of them tests this PR does
not touch. One run here came back with seven fork failures while an immediate
re-run of the identical tree was 528/528. That is the public endpoints at this
fork volume, before and after; CI holds its own, and `rainix / test` is green
on this commit.

## QA

- Discriminating tests: `testSupportedNetworkChainIdsAreBound` is the binding
  itself; `testChainIdMismatchReverts`, `testChainIdMatchPasses`,
  `testChainIdChecksEveryDeclaration`,
  `testChainIdIsReadFromTheForkedEndpoint`, `testChainIdNoDeclarationsReverts`,
  `testDeclaredChainIdsReadsTheConfigText`,
  `testDeclaredChainIdsSkipsEntriesWithNoChain`,
  `testDeclaredChainIdsSkipsNetworksWithNoEntry` and
  `testDeclaredChainIdsOfUrlOnlyEntriesIsRefused` discriminate the pieces. For
  the earlier half, `testConfigWithResolvableEtherscanEntriesPasses`,
  `testConfigWithUnresolvableLastEtherscanEntryReverts` and
  `testConfigWithUnresolvableMiddleEtherscanEntryReverts`. Verified on base by
  breaking what they assert rather than by running tests that do not exist
  there: with `src/` and `test/` checked out at `a741d90` and `flare` restated
  as `chain = 15`, the whole suite is **515 passed, 0 failed** — a config that
  would submit Flare deployments to chain 15's explorer is invisible to every
  check on base. The same mutation on this commit is `M08` below and fails with
  `NetworkChainIdMismatch("flare", 15, 14)`. For the entry-shape half the
  equivalent is dropping `chain = 999` from `hyperevm`, which on `main` leaves
  `[PASS] testSupportedNetworksAreFullyConfigured() (gas: 74114)` against an
  unmutated `74198` — the gas delta is the shorter config string and confirms
  the entry contents were never read.
- Mutations applied: `mutation-probe` 0.1.0 (`rainlanguage-skills`), 9/9 KILLED,
  0 survived, 0 no-run, 0 harness errors, over two scopes, plus the earlier
  pass on `RainDeployVerifySnapshotBase` (9/9 KILLED) which still holds.
  - `src/abstract/RainDeployVerifyChain.sol`, scope `--match-contract
    RainDeployVerifyChainTest` (baseline green, 21 passed):
    - `M01` `if (declared != reported)` -> `if (false)` ->
      `testChainIdMismatchReverts`
    - `M02` select/compare loop `i < declared.length` -> `i < 1` ->
      `testChainIdChecksEveryDeclaration`
    - `M03` `revert NoDeclaredChainIds()` -> `return` ->
      `testChainIdNoDeclarationsReverts`
    - `M04` `vm.selectFork(forkIds[i])` deleted ->
      `testChainIdIsReadFromTheForkedEndpoint`
    - `M05` count loop's `vm.keyExistsToml(...)` -> `true` ->
      `testDeclaredChainIdsSkipsEntriesWithNoChain`
    - `M06` `network: networks[i]` -> `network: ""` ->
      `testDeclaredChainIdsReadsTheConfigText`
    - `M07` `next++` deleted -> `testDeclaredChainIdsReadsTheConfigText`
  - `foundry.toml`, scope `--match-contract RainDeployVerifyChainEmptyTest
    --match-test testSupportedNetworkChainIdsAreBound` (baseline green, 1
    passed):
    - `M08` `flare` `chain = 14` -> `chain = 15` ->
      `testSupportedNetworkChainIdsAreBound`
    - `M09` `flare = "${FLARE_RPC_URL}"` -> `"${ARBITRUM_RPC_URL}"` (the alias
      bound to another network) -> `testSupportedNetworkChainIdsAreBound`, with
      `NetworkChainIdMismatch("flare", 14, 42161)`
- Oracle: foundry itself for the entry shape — the condition asserted is the
  text of foundry's own error, reproduced live on this toolchain by breaking the
  `hyperevm` entry and watching `--verify` fail on polygon. For the ids, `cast
  chain-id` against each `[rpc_endpoints]` URL, which is neither the config nor
  the check: arbitrum 42161, base 8453, base_sepolia 84532, bsc 56, ethereum 1,
  flare 14, hyperevm 999, polygon 137, robinhood 4663. All nine agree with what
  `[etherscan]` states, which is why the new test is green rather than red — and
  agreeing is what it is FOR, since nothing now lets them drift apart unnoticed.
- Category check: #192 asks for one thing — that an `[etherscan]` entry which
  cannot verify fail on the PR rather than after a broadcast — and names the
  closing assertion as "every `[etherscan]` entry carries at least one of
  `chain` or `url`". Covered exactly that, at exactly that strength. The chain
  id binding is added on top of it by direct ruling, because stating `chain` on
  nine entries is what made the ids worth re-checking and this is the PR that
  states them; it covers the declared id, the url-only entry that declares none,
  the config where nothing declares one, and the mirror case of an alias bound
  elsewhere. Nothing else in either group's contract changed — membership is
  still asserted in both directions, by membership, on the same raw text.

Suite: 515 passed / 0 failed at `a741d90`, 528 passed / 0 failed here (13 more:
nine new cases, and the inherited binding in each of the four contracts).
`forge fmt --check` clean.


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

PR #232, opened 2026-09-15
Clear one fixture tree at the start of every run

### LINKED ISSUE #190: A failing freeze-guard test leaves its fixture cut behind and then fails permanently with a different error
## Unit and file

`testFreezeChecksTheRecordItIsAppendingTo` at test/src/lib/LibRainDeploySnapshot.t.sol:1827, against `freeze` in src/lib/LibRainDeploySnapshot.sol (the `SnapshotAlreadyFrozen` guard at :1353 and `checkReleaseFollowsRecord` at :1364). The same shape applies to every test in the file that writes under `test/generated-*` or the fixture lib root.

## The intent oracle

The file documents its own discipline repeatedly: read before the fixture is removed and assert after, because forge-std assertions revert -- cleaning up afterwards cleans up in every case except a failure, which is the one case that leaves a directory behind (test/src/lib/LibRainDeploySnapshot.t.sol:447, :629, :1190). The intent is that a failing run leaves no fixture residue in the tree.

`freeze` checks `vm.exists(frozenDir)` and reverts `SnapshotAlreadyFrozen` at :1353, BEFORE `checkReleaseFollowsRecord` at :1364, so a leftover cut necessarily shadows the ordering error this test asserts.

## The property violated

A test must be repeatable: running it must not change the outcome of running it again. This one is not. The discipline the file documents protects against an ASSERTION failure only -- an `expectRevert` mismatch aborts at the guarded call, which is upstream of `vm.removeDir` in every test in the file that follows the pattern. So the pattern is a class-wide gap, and this test is the instance where the residue is self-poisoning: the residue it leaves is a real `0_1_10/` cut that the unrefused freeze wrote, and on the next run `freeze` hits `SnapshotAlreadyFrozen` before it can reach the ordering guard the test exists to observe.

One failing run therefore converts a repeatable test into a permanently failing one that reports a different and misleading cause, recoverable only by hand.

In a mutation pass this is worse than misleading: the first mutant to break this test silently kills every mutant after it. That happened in this campaign -- an inherited clone arrived with `test/generated-freeze-guard/0_1_10/` on disk from a previous worker, and a full pass reported 54/54 killed when seven of those were not killed at all.

## Verified repro

Re-verified today against current `main` at e9a922a in a fresh clone with no inherited residue (`test/generated-freeze-guard` absent at the start).

1. Apply the mutation in `isStrictTriple`: change `char >= 0 && char <= 9` to `char >= 0 && char < 9` (digit range upper bound).
2. `nix develop -c forge test --match-test testFreezeChecksTheRecordItIsAppendingTo` gives `[FAIL: next call did not revert as expected]`, and the tree is left holding `test/generated-freeze-guard/{9_9_9,candidate,0_1_10}/MockDeployable.sol` -- including the real `0_1_10` cut the unrefused freeze wrote.
3. `git checkout src/lib/LibRainDeploySnapshot.sol` to revert the mutation, then re-run UNMUTATED:

```
[FAIL: Error != expected error: SnapshotAlreadyFrozen(0_1_10, test/generated-freeze-guard/0_1_10) != NonMonotonicRelease(0_1_10, 9_9_9)]
```

4. It stays red on further re-runs. `rm -rf test/generated-freeze-guard` restores green, and the clean pass leaves no residue.

## Triage framing

For it being intended: the file documents the read-before-cleanup, assert-after ordering deliberately, and most of its tests follow it precisely so that a failure does not leave a directory under the real `src/generated/`. This one cannot be ordered the other way: the observation it needs (that no cut exists) is about the same tree it has to tear down, and it does read both facts before removing. The residue only appears on the `expectRevert` path, which is not an assertion whose ordering the test controls. The residue is also visible -- `git status` shows `?? test/generated-freeze-guard/`, since `test/generated-*` is not gitignored -- so a human who runs `git status` would see it.

Against: a `setUp()` that removes the fixture roots fixes it without changing anything the test observes, so the it cannot be ordered any other way defence does not hold. The harm falls on automated passes and on anyone re-running the suite without inspecting the tree, and the second failure names a cause that has nothing to do with the property under test.

Note -- independent verification verdict: bug. The verifier reproduced the full cycle in a fresh clone with no inherited residue and confirmed both the residue and the permanent, misleadingly-named second failure, and derived the shadowing mechanism from the source ordering (`SnapshotAlreadyFrozen` at :1352 before `checkReleaseFollowsRecord` at :1364) without running anything. It notes the residue is visible to `git status`, so severity rests on automated passes rather than on invisibility, with this campaign's own 54/54 report as the concrete harm.

thedavidmeister:
Closes #190.

`LibRainDeploySnapshotTest` had no `setUp`, so every run started on whatever the
run before it left in the tree. Each test reads before it removes and asserts
after — the discipline the file documents — but a mismatched `vm.expectRevert`
fires at the guarded call, which is upstream of every `vm.removeDir` in the
contract, so the one run that leaves a fixture behind is a run that already
failed.

For `testFreezeChecksTheRecordItIsAppendingTo` that residue is self-poisoning:
the unrefused freeze writes a real `<tag>/` cut, and `freeze` checks
`vm.exists(frozenDir)` and reverts `SnapshotAlreadyFrozen` BEFORE it reaches
`checkReleaseFollowsRecord`. One failing run therefore turns a repeatable test
into a permanently red one reporting a cause that has nothing to do with the
property under test, recoverable only by deleting the directory by hand. In a
mutation pass the first mutant to break it silently kills every mutant after it.

## What changed

Every fixture root this contract writes under `test/` moves into ONE tree,
`test/generated-snapshot/`, named once by `FIXTURE_ROOT`. `setUp` removes that
tree, so a root added later is covered by being under it — nothing enumerates
the roots, so nothing can be left out of the clear. `setUp` is the only safe
point to do it: forge runs a contract's tests concurrently, and runs `setUp`
once, before any of them.

Two `clearFixtureTree` calls rather than one, because `FIXTURE_LIB_ROOT`
(`fixture-lib/`) cannot join the tree. A generated lib imports `../generated/`,
`../abstract/` and `./Lib<Contract>Released.sol`, which resolve from `src/lib`
and nowhere else, so a copy under any COMPILED root fails the build for every
suite — and the build runs before `setUp`, which puts it upstream of anything
`setUp` could do about it. That is why `fixture-lib/` is outside `src/` and
`test/` to begin with.

Deliberately NOT cleared:

- The seven `src/generated/<dir>` fixtures. That root is the writer under test,
  and those directory names are deliberately not tag shaped, so what a failure
  leaves there is passed over by every record walk and overwritten by the test
  that wrote it.
- `FROZEN_FIXTURE_ROOT` (`test/fixture-record/`). Being outside the tree is what
  keeps it: it is committed and read only, so a clear that reached it would
  delete it from the repo.

`testClearingAFixtureTreeRemovesAStaleCut` drives the clear over exactly the
residue a failed freeze-guard run leaves.

No source change: `freeze` writing that cut is `freeze` doing its job once the
ordering guard has passed. The defect is that the run inherits it.

## What the diff actually is

`+170/-58`, of which the mechanism is 106 added lines: the `FIXTURE_ROOT`
constant, `setUp`, `clearFixtureTree` and the new test — comments included, and
the comments are most of it.

The remaining `+64/-58` has no behaviour in it. 24 existing fixture-root
constants are re-rooted under the tree, one line each (`+24/-24`,
`test/generated-freeze` -> `test/generated-snapshot/freeze`); the walk test's
`FIXTURE_ROOT` becomes `WALK_FIXTURE_ROOT` at all 21 of its sites, now that the
bare name belongs to the tree; and `forge fmt` re-wraps the comments and
assertions the longer name no longer fits on one line. `foundry.toml` is
`+2/-2`, the comment naming the root.

The hand-maintained surface is what the extra lines buy. The first version of
this branch cleared the same residue from a written-out `fixtureRoots()` array
of 33 entries — every fixture directory named twice, once at its constant and
once in the list, and a root silently uncleared if the list was not updated with
it. That was `+135/-2`: a smaller diff carrying a standing invariant no compiler
checks. One tree costs more lines once and nothing after.

## QA

- Discriminating tests:
  - `testFreezeChecksTheRecordItIsAppendingTo` — pre-existing, and it
    discriminates from the defect's own precondition, which is the whole of the
    defect. With `<root>/0_1_10/MockDeployable.sol` planted before the run: on
    base (`baa1a9c`) it is
    `[FAIL: Error != expected error: SnapshotAlreadyFrozen("0_1_10", "test/generated-freeze-guard/0_1_10") != NonMonotonicRelease("0_1_10", "9_9_9")]`,
    still red on a second run, and leaves `?? test/generated-freeze-guard/`
    behind; on this branch, from the identical plant at
    `test/generated-snapshot/freeze-guard/0_1_10/`, it passes and `git status` is
    clean afterwards. Both run.
  - `testClearingAFixtureTreeRemovesAStaleCut` — new, and cannot run on base:
    `clearFixtureTree` and the guarantee it pins do not exist there. It is
    discriminated by mutation instead, below.
- Mutations applied — all three run on this branch, each from a clean tree:
  - `FIXTURE_ROOT`: `"test/generated-snapshot"` ->
    `"test/generated-snapshot/walk"` — the clear still runs, but over one test's
    subtree instead of over the tree — with
    `test/generated-snapshot/freeze-guard/0_1_10/MockDeployable.sol` planted ->
    killed by `testFreezeChecksTheRecordItIsAppendingTo`, as
    `SnapshotAlreadyFrozen("0_1_10", "test/generated-snapshot/freeze-guard/0_1_10") != NonMonotonicRelease("0_1_10", "9_9_9")`.
    Unmutated, from the same plant, it passes and leaves nothing. This is the
    mutation that pins the one-tree design itself: the umbrella has to be the
    umbrella and not one of the things under it.
  - `clearFixtureTree`: `if (vm.exists(root))` ->
    `if (false && vm.exists(root))` (the clear becomes a no-op) -> killed by
    `testClearingAFixtureTreeRemovesAStaleCut`, as
    `SnapshotAlreadyFrozen("0_1_10", "test/generated-snapshot/freeze-stale/0_1_10") != NonMonotonicRelease("0_1_10", "9_9_9")`
    — the exact shadowing #190 describes.
  - `clearFixtureTree`: `vm.removeDir(root, true)` ->
    `vm.removeDir(root, false)` (non-recursive) -> killed by
    `testClearingAFixtureTreeRemovesAStaleCut`, as
    `vm.removeDir: failed to remove dir ".../test/generated-snapshot/freeze-stale": Directory not empty (os error 39)`.
- Oracle: the expected refusal is `NonMonotonicRelease(tag, "9_9_9")`, the
  ordering guard's own refusal in `checkReleaseFollowsRecord`, reached only when
  the `SnapshotAlreadyFrozen` guard above it does not fire — read from
  `src/lib/LibRainDeploySnapshot.sol`, not from anything the test writes. The
  intent — that a run does not inherit the fixtures of the run before it — is
  the file's own documented read-before-remove, assert-after discipline, stated
  at three of its cleanup sites.
- Category check: #190 asks for (a) a failing run not leaving its fixture cut
  behind and (b) the next run not failing permanently with a different error.
  Both are covered by the same change and verified together: from the planted
  `0_1_10/` cut the run passes AND `git status` is clean when it ends, so the
  residue a failure leaves is gone by the start of the next run rather than
  deciding it. The issue also names the shape as class-wide — every test in the
  file writing under `test/generated-*` or the fixture lib root — which is why
  the clear is a `setUp` over a whole tree rather than one root, and why the
  roots moved into a tree rather than into a list.

## Verification

`git clean -fdq test src && rm -rf fixture-lib cache/fuzz` before every run,
`.env.example` with the hyperliquid, arbitrum and bsc endpoints swapped for
reachable ones.

- Whole suite, one run each side: base `baa1a9c` 512 passed / 0 failed; this
  branch 513 passed / 0 failed. `git status` clean after both.
- `forge fmt --check` clean. `forge lint` reports one `boolean-cst` warning at
  `LibRainDeploySnapshot.t.sol:1596`, which is the same code on base and is not
  touched here.

### The pre-existing race, measured

Unrelated, pre-existing, and NOT touched here: `freezeConsensusFixture()` is
shared by `testWriteSnapshotRecordsTheZoltuAddress` and
`testWriteSnapshotHashesTheCodeItRecords`, which run concurrently over the one
`src/generated/writeConsensusNotATag` directory.

Measured rather than assumed: 40 runs of
`forge test --match-contract LibRainDeploySnapshotTest` per side, alternating
base and branch so both sides see the same machine load. Base `baa1a9c` 24/40
red; this branch 26/40 red. Every one of those 50 red runs failed on
`src/generated/writeConsensusNotATag` and on nothing else, in one of those two
tests, as an ENOENT out of `vm.writeFile`, `vm.readFile`, `vm.removeFile` or
`vm.removeDir` — the one shared-directory race in four of its spellings. The
rate swings hard with load (4/10, 4/10 and 16/20 on base against 5/10, 7/10 and
14/20 on the branch, in that order) and the two sides swing together, which is
the observation worth having: the race is there on base, this change neither
introduces it nor removes it, and 40 runs a side do not separate the two rates.

It is a different defect — two tests sharing one directory — from the one #190
names, so it is left for its own issue.

## Summary by CodeRabbit

- **Tests**
  - Improved snapshot and fixture test isolation by using dedicated generated-snapshot directories.
  - Added cleanup between test runs to prevent stale fixture data from affecting results.
  - Expanded coverage for frozen cuts, nested and missing records, library generation, directory walking, and ordering behavior.
  - Added validation that stale frozen cuts are removed before freeze checks.


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

**This branch.** merge-base `baa1a9ca5821c6b41bd148fecc4fd87bd0f3666c`. This PR does touch `LibRainDeploySnapshot.t.sol` (+166/-56), but both functions are byte-identical to the merge-base; only their line numbers move, 599/611 to 660/672. Both functions came in with

PR #234, opened 2026-09-15
test: assert from the AST that run() calls every hook BuildScript declares

thedavidmeister:
Direct ruling, not an issue: assert that `BuildScript.run()` calls every
`internal virtual` hook `BuildScript` declares.

## What was wrong

`src/abstract/BuildScript.sol` declares four `internal virtual` hooks and calls
them from its two entry points. Nothing asserted the wiring itself. Every
existing check was output-anchored — `BuildScriptTest` drives the harness and
reads the markers the hooks wrote — so it sees a hook that is CALLED and says
nothing about one that is not. A hook added beside the two that exist today,
wired by nobody, leaves the whole suite green: a generator a deriving repo is
asked to implement and that no push ever runs, whose output drifts from its
inputs until a release cuts the drift into the append-only record.

Verified rather than assumed, on current `main`: a third hook
(`function regenerateDocs() internal virtual {}`) added to the base and called
by nothing left the suite at 534 passed / 0 failed. A hook that only reads
(`function docRoot() internal view virtual returns (string memory)`) left it at
534 passed / 0 failed. A `recordRoot();` added to `run()` — a statement of the
entry point that regenerates nothing — left it at 534 passed / 0 failed. The
unmutated baseline is the same 534 passed / 0 failed.

## What changed

`test/src/abstract/BuildScript.t.sol` gains three assertions, read out of
solc's AST of the base rather than from the source text, for the reason
`GeneratedSnapshotShapeTest` gives: this is about STRUCTURE, not formatting.
Nothing under `src/` or `script/` changes.

- `testRunCallsEveryHookThatRegenerates` — the hooks are enumerated from the
  base's own declarations (`internal` + `virtual`), not named, and every one of
  them that can write (`stateMutability` is `nonpayable`) must be called by
  `run()`, which must hold nothing else. Naming today's two is the failure
  being fixed, so the walk finds a third hook the moment it is declared.
- `testEveryHookIsReachedFromAnEntryPoint` — the other half, for the hooks that
  only read. `run()` has no use for `recordRoot()` or
  `snapshotContractNames()`, so requiring it to call them would be wrong; a
  hook reached from NEITHER entry point is the same defect as an unwired
  generator. Reachability rather than a direct call, because
  `regenerateSnapshots` reaches `freeze` as an internal function pointer and
  not as a call.
- `testTheAstIsTheBaseTheHarnessInherits` — the no-subject guard. Both walks
  are claims about a file named by a hard-coded path, and an artifact left
  behind by a moved source parses as well as a live one. The base is abstract
  so its creation code cannot anchor it the way `CreditHyperCoreTest` anchors
  the script; the chain runs through the concrete harness instead: the harness
  artifact is the harness this suite compiles, the base it inherits was
  imported from `src/abstract/BuildScript.sol`, and the base artifact describes
  that file.

## Relation to #228

Disjoint, and complementary. #207 — the issue #228 closes — asks for the
arguments, the loop bound and the call ordering INSIDE the generator hooks;
this is the wiring TO them, which #207 does not ask for and #228 does not add.

#228 (`fix-207`) opens a fixture-directory seam so `regenerateLibs` can be RUN,
and pins the bytes it emits; it touches
`script/Build.sol`, `test/concrete/BuildHarness.sol` and
`test/script/Build.t.sol`. This PR touches only
`test/src/abstract/BuildScript.t.sol` and asserts no bytes: #228 says one hook's
body does the right thing when it is run, this says every hook the base declares
is run at all. Neither test would catch the other's mutation, and the two
branches share no file.

## Migration

None. Test-only.

## QA

- Discriminating tests: `testRunCallsEveryHookThatRegenerates`,
  `testEveryHookIsReachedFromAnEntryPoint`,
  `testTheAstIsTheBaseTheHarnessInherits` - each fails on base in the only sense
  available to a test that does not exist there: the three mutations named above
  were applied to `main` itself and the whole suite stayed green, and the same
  mutations on this branch are killed with the messages below.
- Mutations applied (each reverted before the next):
  - `src/abstract/BuildScript.sol` gains `function regenerateDocs() internal
    virtual {}`, called by nothing (M1) ->
    `testRunCallsEveryHookThatRegenerates`
    (`run() does not call the hook regenerateDocs`) AND
    `testEveryHookIsReachedFromAnEntryPoint` (`nothing reaches the hook
    regenerateDocs`)
  - `src/abstract/BuildScript.sol` gains `function docRoot() internal view
    virtual returns (string memory)`, called by nothing (M3) ->
    `testEveryHookIsReachedFromAnEntryPoint` (`nothing reaches the hook
    docRoot`)
  - `run()` drops `regenerateLibs();` (M2) ->
    `testRunCallsEveryHookThatRegenerates` (`run() does not call the hook
    regenerateLibs`). Also killed on base, by the existing
    `testRunRegeneratesAndFreezesNothing` reading the marker that was never
    written - this mutation is the half that was already covered.
  - `run()` gains `recordRoot();` (M4a) ->
    `testRunCallsEveryHookThatRegenerates` (`run() holds a call that is not one
    of those hooks: 3 != 2`)
  - `run()` gains `vm.writeFile("mutation", "");` - a regeneration inlined into
    the entry point instead of reaching a hook (M4b) ->
    `testRunCallsEveryHookThatRegenerates` (`the entry point calls something
    other than a function of its own: MemberAccess != Identifier`)
  - `BASE_ARTIFACT` points at the harness's artifact instead of the base's (M5)
    -> `testTheAstIsTheBaseTheHarnessInherits` (`the base artifact describes
    another file: test/concrete/BuildScriptHarness.sol !=
    src/abstract/BuildScript.sol`), and both walks with `the contract is not the
    base: BuildScriptHarness != BuildScript`
- Oracle: the base's own declarations, read from the compiler's AST, against the
  statements of its entry points. Independent of the implementation in the sense
  that matters here: no name of any hook appears in the test, so the expected
  set
  cannot be edited to agree with a wrong `run()` — it IS the declarations, and
  the only way to satisfy it is to call them.
- Category check: the ruling asks for the property `run()` calls every
  `internal virtual` hook `BuildScript` declares, enumerated from the AST rather
  than by naming today's two, red when a hook is added unwired. Covered:
  enumerated by `visibility`/`virtual` from the contract node; every hook that
  writes is required in `run()`; the ones that only read are required to be
  reached from an entry point, which is where the literal reading of the ruling
  does not hold and cannot — `run()` calls neither `recordRoot()` nor
  `snapshotContractNames()` today, and requiring it to would be a false red on
  correct code. Both halves are red under an added hook (M1, M3).

`nix develop -c forge test -j 2` on `feefc95` (current `main`): 534 passed / 0
failed. On this branch: 537 passed / 0 failed, 534 + the three new tests. Same
`.env`, same run-to-run endpoints, no fork failures either way.


thedavidmeister:

```
Error: Compiler run failed:
```

reaching a single lint rule.

